PR #3, opened 2018-10-14
define interfaces for milestone 1

zixuanzh:
- [ ] host interface and basic host
- [ ] network interface which sets and handles stream
- [ ] stream interface


robzajac:
LGTM

PR #23, opened 2018-10-27
WIP TCP.py

zixuanzh:
@alexh 
TODOs
- [ ] exception handling in listen
- [ ] check get_addrs in other implementation, interface documentation ambiguous
e.g. https://github.com/libp2p/js-libp2p-utp/blob/master/src/create-listener.js
- [ ] dial behavior not fully understood
ref: https://github.com/libp2p/interface-transport
- [ ] double check on run_until_complete


robzajac:
Legacy; removing.

PR #44, opened 2018-11-11
Swarm new stream

stuckinaboot:
- [ ] Re-implement new stream 
- [ ] Stub out logic for listen
- [ ] Organize folders to reflect to project architecture
- [ ] Implement net stream (accidentally described as "Add mux stream")


alexh:
lgtm

PR #45, opened 2018-11-11
Muxed stream

zixuanzh:


muxer/mplex/muxed_stream.py
```diff
@@ -1,34 +1,113 @@
...
+        """
+        if self.initiator:
+            return HEADER_TAGS[action]
+        else:
+            return HEADER_TAGS[action] - 1
 
     def read(self):
-        pass
+        """
+        read from stream til end of file
```
alexh: Can you update the comment to say "read any messages from buffer associate with stream"

PR #64, opened 2018-11-19
Chat example

### LINKED ISSUE #16: Implement chat poc
Example:

https://github.com/libp2p/go-libp2p-examples/tree/master/chat

zaibon:
fixes #16 

First draft of a chat example.
Different things to notice:
- current the example is in a `chat.py` module at the root of the repository. I would like to move that into a example folder. But then import won't work anymore. What is the plan regarding the libp2p modules ? are we going to install them into a single package or each package separately ? Once I know that, I can move install the lib properly and then move `chat.py` in a proper location
- currently the code use a really simplistic version of the a peer ID. From what I've seen this has not been implemented in the code yet. 
- I used the [click library](https://click.palletsprojects.com) for the CLI part of the code. I found it nicer the argparse, but I can change it to match the rest of the code

### DIFF
--- a/chat.py
+++ b/chat.py
@@ -1,0 +1,101 @@
+#!/bin/env python3
+import asyncio
+
+import click
+from libp2p.libp2p import Libp2p
+from network.multiaddr import MultiAddr
+
+# TODO: change once muxed_connection supports extracting protocol id from messages
+PROTOCOL_ID = '/echo/1.0.0'
+
+async def read_data(stream):
+    while True:
+        read_string = (await stream.read()).decode()
+
+        if not read_string:
+            return
+
+        if read_string != "\n":
+            # Green console colour: 	\x1b[32m
+            # Reset console colour: 	\x1b[0m
+            print("\x1b[32m%s\x1b[0m> " % read_string)
+
+async def write_data(stream):
+    while True:
+        s = input('> ')
+        await stream.write(s.encode())
+
+
+
+async def run(port, destination):
+
+    if not destination:
+        lib = Libp2p(transport_opt=["/ip4/127.0.0.1/tcp/%s/p2p/hostA" % port])
+        host = await lib.new_node()
+
+        async def stream_handler(stream):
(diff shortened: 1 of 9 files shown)


zaibon:
> Users are responsible for supplying a peer_id. That's why we are just taking in what users provide in add_addr. An RSA key is generated for its peer_id when a new user joins the network. Do you have a reference link to the Go repo where something more sophisticated is implemented?

https://github.com/libp2p/go-libp2p-peer/blob/master/peer.go
Seems in the go implementation the peer_id is closely linked to the RSA key and they have a bunch of function to deal with it.

robzajac:
Thanks for this! Re: peer IDs, I believe we generate them [here](https://github.com/zixuanzh/py-libp2p/blob/master/libp2p/libp2p.py#L17) - are you referring to the test_libp2p under tests/ which does not use this peer ID?

upperwal:
Good work guys. I was looking for a python implementation for a long time. 

go-libp2p uses a SHA-256 hash on the public key to get the peer id using ```IDFromPublicKey```

```go
// IDFromPublicKey returns the Peer ID corresponding to pk
func IDFromPublicKey(pk ic.PubKey) (ID, error) {
b, err := pk.Bytes()
if err != nil {
  return "", err
}
var alg uint64 = mh.SHA2_256
if len(b) <= MaxInlineKeyLength {
  alg = mh.ID
}
hash, _ := mh.Sum(b, alg, -1)
return ID(hash), nil
}
```

[here](https://github.com/libp2p/go-libp2p-peer/blob/d3df4bca884d7a9c2d350c8120240db3c2b0f2ee/peer.go#L147-L159) is the actual code.

I see that you are directly using the public key as the id.

https://github.com/zixuanzh/py-libp2p/blob/4e2749c9151bc312875a5e07fdfef588f1872e75/libp2p/libp2p.py#L17-L19

You can take a SHA256 hash right here and then use it. In the future you need to store the ```_RSAobj``` for secio.

PR #66, opened 2018-11-21
Refactor Muxed_Connection to Mplex

zixuanzh:
- [x] muxed_connection already implemented what previously smux-multiplex defines, renamed file, moved interfaces to parent folder stream_muxer for yamux in the future. Changes to main logic will come in later.
- [x] completed feature mplex_stream close and reset based on the go repo
- [x] wrote simple echo test for connection, tests for mplex can be built from this


stream_muxer/mplex/mplex_stream.py
```diff
@@ -40,50 +40,57 @@ def get_flag(self, action):
...
+        but allows writing in the other direction.
         :return: true if successful
         """
+        #TODO error handling with timeout
+        await self.mplex_conn.send_message(self.get_flag("CLOSE"), None, self.stream_id)
 
-        if self.local_closed and self.remote_closed:
-            return True
+        remote_lock = ""
+        async with self.stream_lock:
```
robzajac: Can you explain why (in code or PR or both) we have both a stream lock and raw connection lock? I have a bit of trouble following.

stream_muxer/mplex/mplex.py
```diff
@@ -1,10 +1,10 @@
 import asyncio
```
robzajac: Do we need the old muxer folder at all? If not, let's delete.

zixuanzh: the old muxer folder is renamed to stream_muxer. we should also have a protocol_muxer folder in the future

stream_muxer/mplex/mplex_stream.py
robzajac: Got it - in other words, a stream can be modified both by the stream initiator and the stream receiver (e.g. if the receiver closes their corresponding stream and sends a CLOSE), necessitating a lock. Since we have TODOs for this, LGTM.

PR #67, opened 2018-11-21
enable CI

zaibon:
I used travis-ci and codecov.io

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+language: python
+
+python:
+  - "3.6"
+
+install:
+  - pip install -r requirements.txt
+  - pip install codecov pytest pytest-cov
+
+script:
+  - pytest --cov=./ -v
+  - pylint --rcfile=.pylintrc encryption host kademlia libp2p network peer stream_muxer transport tests
+
+after_success:
+  - codecov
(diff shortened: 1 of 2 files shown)


zixuanzh:
We are also considering using CircleCI as per py-evm.

zaibon:
All good I'll make the required change. I can move to circleci too. I picked Travis to match what I saw in the other libp2p but doest matter really.

zaibon:
Ok I changed from travis-ci to circleci. It's my first time using circleci, I think is till have a problem in the config since pytest doesn't seems to be able to find the library when running the test.

robzajac:
@zaibon I found [this](https://stackoverflow.com/a/51306900) helpful for the problem you just described. After removing __init__.py and adding conftest.py I could run `pytest` successfully. But without __init__.py, I couldn't run `pylint` from the root directory. I also couldn't run `codecov`, but I think that's because the config doesn't include [codecov install steps](https://docs.codecov.io/docs/deploying-with-docker).

Is there a way to run all of `pytest`, `pylint`, and `codecov` in harmony with Circle? If not, we may go back to evaluating Travis. Let us know your thoughts, thanks!

zaibon:
I found a configuration that seems to kinda works. But seems it only works with python3.7, python3.6 gives strange errors. 

With python3.7, pytest and pylint works. I couldn't really test codecov cause I need to fix all the linting error first, but I don't see why it wouldn't work.

alexh:
Hey @zaibon! We had a meeting and decided to stick with 3.6 for now. Therefore, lets move back to Travis. This will let us stay conformant with the other libp2p repos as well as you mentioned. Apologies for the back and forth on this!

We would like to run pytest, pylint, and codecov if possible with Travis

zaibon:
@alexh, ok sounds good.

zaibon:
@alexh ok, should be ready to merge now.

PR #73, opened 2018-11-25
Requiring Python >= 3.7

robzajac:
As per [this post](https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-asyncio), 3.7 brought big improvements to `asyncio`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,10 +9,10 @@ py-libp2p is an experimental and work-in-progress repo under heavy development. 
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


zaibon:
Seems they circle ci behave also better with python 3.7 https://github.com/zixuanzh/py-libp2p/pull/67#issuecomment-441175880

robzajac:
Leaving PR open until we know more regarding compatibility with [py-evm, which is on Python 3.6](https://github.com/ethereum/py-evm/blob/master/.circleci/config.yml).

robzajac:
Closing for now - opening issue to start discussion about this instead.

Issue #53, opened 2018-11-12
Post PoC design enhancements


robzajac:
Our PoC makes simplifying design choices that will be improved in the next py-libp2p iteration. These tasks are sorted roughly in decreasing order of priority.

- [x] handle_incoming() should write to a blocking queue buffer (instead of bytearray) and read_buffer() should block on read from the corresponding buffer. Consequently, handle_incoming() can be repeatedly scheduled to run in the background as an asyncio task (instead of being invoked on read_buffer calls) @alexh 
- [x] Protocol muxing: protocol IDs should be sent over the wire and picked up by the receiving side - this way we can multiplex protocols over the same raw connection, `multistream-select`. @stuckinaboot 
- [x] refactor `muxed_connection` to `mplex` @zixuanzh 
- [x] stream IDs should be generated more conventionally - the current design tramples a uuid into a 64 bit representation. @robzajac 
- [ ] more robust error handling, propagating exceptions properly @robzajac 
- [ ] better style around constants used in the code, such as the handle_incoming() timeout and the assumed protocol ID. (Perhaps a constants file?)
- [ ] multiple transports should be supported in Swarm

More inspiration can be found in the TODOs of the source code.

zaibon:
based on my work on #16 I also notice some component that doesn't exist yet in the code and that are not listed in the first comment on this issue.

- library around peer ID. 
- hard-coded timeout on https://github.com/zixuanzh/py-libp2p/blob/01daf8ae8a15ba6c8631bdbac73e89110eb099eb/muxer/mplex/muxed_connection.py#L99 make it hard to use stream on top of user driven code like a chat.

PR #82, opened 2018-11-27
Protocol muxing

stuckinaboot:
As per #53, I implemented protocol muxing. There are still some TODOs that are necessary in order for interoperability with the go repo to occur but basic protocol muxing is up and running. Also, note that I implemented a class multiselect_communicator that is designed to ensure the multiselect module and its matching client will communicate with one another using the same protocol

- [ ] implement multiselect client (communicates with multiselect)
- [ ] implement multiselect (responsible for selecting a protocol if possible and responding to multiselect client)
- [ ] modify swarm to use multiselect
- [ ] modify new stream (in swarm) to take in a list of protocol ids
- [ ] modify basic host to take in a list of protocol ids
- [ ] add end to end tests for protocol muxer


tests/protocol_muxer/test_protocol_muxer.py
```diff
@@ -0,0 +1,88 @@
+import pytest
```
robzajac: Really nice test setup. Only thing to add is that we should have a better way of assigning the ports used in tests, maybe by a _next_port function that gives back a monotonically increasing port number (this seems like a bit of a hack, eventually we probably want to mock the ports). Will open an issue for that later.

PR #94, opened 2018-12-07
don't manually assign port during test

### LINKED ISSUE #89: Assigning port numbers in tests to avoid sharing ports
As mentioned in #82, until we move to mocking network communication, we should assign ports to hosts in our test cases in a more canonical way which avoids: 

1) different tests trying to communicate on the same port
2) assigning port numbers [ad-hoc](https://github.com/zixuanzh/py-libp2p/blob/master/tests/protocol_muxer/test_protocol_muxer.py#L58)

One way to do this is to have tests share a _next_port function which gives a monotonically increasing port number.

zaibon:
Following my idea in https://github.com/zixuanzh/py-libp2p/issues/89#issuecomment-442881962, instead of saving the multiaddr received in the listen method of the tcp transport class, read what the service is actually listening on this allow to use 0 as port and let the OS choose a free port for us.

2 remark on this though:
- is it ok to make the tcp transport depend on the multiaddr lib ?
- the function I added to read the actual port listening has `ip4` hard-coded in it. It seems the rest of the file too, but maybe we want to already plan to support ip6 or other protocols

fixes #89

### DIFF
--- a/transport/tcp/tcp.py
+++ b/transport/tcp/tcp.py
@@ -1,7 +1,11 @@
 import asyncio
-from transport.transport_interface import ITransport
-from transport.listener_interface import IListener
+
+import multiaddr
+
 from network.connection.raw_connection import RawConnection
+from transport.listener_interface import IListener
+from transport.transport_interface import ITransport
+
 
 class TCP(ITransport):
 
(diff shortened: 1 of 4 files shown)


transport/tcp/tcp.py
```diff
@@ -22,14 +26,15 @@ def __init__(self, handler_function=None):
...
-            self.multiaddrs.append(_multiaddr)
 
             coroutine = asyncio.start_server(self.handler,
-                                            _multiaddr.value_for_protocol('ip4'),
-                                            _multiaddr.value_for_protocol('tcp'))
+                                             _multiaddr.value_for_protocol('ip4'),
+                                             _multiaddr.value_for_protocol('tcp'))
             self.server = await coroutine
+            socket = self.server.sockets[0]
+            self.multiaddrs.append(_multiaddr_from_socket(socket))
```
robzajac: Good catch, pulling the port after server creation.

robzajac:
LGTM, thanks for this. Regarding the remarks:

- This should be okay, following along from [this spec](https://github.com/libp2p/interface-transport#nodejs).

- That's right, eventually we should be generalizable to ip6 or other protocols. We will look at other protocols for any particular layer after working out more high-level features (we are targeting peer discovery soon).

Issue #92, opened 2018-11-29
Closing network resources in tests, adding close functionality


robzajac:
Currently no test does any kind of cleanup (e.g. by calling `close` methods of the underlying connections opened by a host). This balloons into errors (but not test case failures) of the "Task was destroyed but it is pending!" type in [CI](https://travis-ci.com/zixuanzh/py-libp2p).

- [ ] py-libp2p should do cleanup in tests.
- [ ] py-libp2p should add `close` methods to `host` and `network` instances in the style of [go-libp2p](https://github.com/libp2p/go-libp2p-host/blob/master/host.go#L64).

Lynndabel:
Hello, can i work on this?

Smartdevs17:
@seetadev I’d like to work on the issue.

From my understanding, current tests create hosts and network connections but never explicitly clean them up. This leaves pending async tasks and open resources, which causes CI warnings like “Task was destroyed but it is pending!” even though tests pass.

The root issue seems to be the lack of explicit lifecycle management in py-libp2p. There are no close() / shutdown methods on Host and Network, so tests have no way to properly release resources.

A clean solution would be to:

Add explicit close() (async) methods to host and network instances, similar to go-libp2p

Ensure these methods cancel background tasks and close underlying connections

Update tests to call close() during teardown so no async tasks remain pending

Happy to align with existing patterns in the codebase and discuss the expected shutdown semantics before implementation.

seetadev:
@Smartdevs17 : great, please go ahead. Keep @acul71 , @yashksaini-coder and @sumanjeet0012 in the loop.

Smartdevs17:
@seetadev thank you will do just that.

PR #95, opened 2018-12-20
Close listeners after completing connection

ngetahun:
#92 Closes open listeners
@robzajac Should testing this go into ```test_libp2p.py``` or create a new folder for testing transport?


robzajac:
Since this is a `network` level function we would create a new folder for testing `network`. Thanks!

ralexstokes:
@ngetahun going to close this due to inactivity -- please feel free to re-open or move to an issue if there is stuff in here you want addressed :)

PR #96, opened 2018-12-20
make the chat example a runnable test

### LINKED ISSUE #71: Converting examples/chat/chat.py into a pytest test case
The example at `examples/chat/chat.py` is a useful benchmark of correctness. It would be nice to have this as a runnable `pytest` test case (maybe chat over a few messages in varying orders?) to avoid running it manually upon code changes.

zaibon:
fixes #71

### DIFF
--- a/tests/examples/test_chat.py
+++ b/tests/examples/test_chat.py
@@ -1,0 +1,110 @@
+#!/bin/env python3
+import asyncio
+import sys
+
+import click
+import multiaddr
+import pytest
+
+from libp2p.libp2p import *
+from peer.peerinfo import info_from_p2p_addr
+from protocol_muxer.multiselect_client import MultiselectClientError
+
+PROTOCOL_ID = '/chat/1.0.0'
+
+
+async def hello_world(host_a, host_b):
+    async def stream_handler(stream):
+        read = await stream.read()
+        assert read == b'hello world from host b'
+        await stream.write(b'hello world from host a')
+        await stream.close()
+
+    host_a.set_stream_handler(PROTOCOL_ID, stream_handler)
+
+    # Start a stream with the destination.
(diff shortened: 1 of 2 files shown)


tests/examples/test_chat.py
```diff
@@ -0,0 +1,94 @@
...
+import sys
+
+import click
+import multiaddr
+import pytest
+
+from libp2p.libp2p import *
+from peer.peerinfo import info_from_p2p_addr
+
+# TODO: change once muxed_connection supports extracting protocol id from messages
```
zixuanzh: I believe this has been enabled with `protocol_muxer` which is equivalent of `multistream-select`. I will let @stuckinaboot shed more light on this.

zaibon: haaa could be indeed this has been fixed since then. I had this code laying around for a while. I'll check. Thanks

zaibon: So I tested and yes indeed protocol id is now passed during negotiation. So I updated the code and added a test for this too.

stuckinaboot: @zaibon great thanks, if you have any questions about negotiation feel free to let me know.

tests/examples/test_chat.py
```diff
@@ -0,0 +1,110 @@
...
+            await stream.write(message)
+        await stream.close()
+
+    host_a.set_stream_handler(PROTOCOL_ID, stream_handler)
+
+    # Start a stream with the destination.
+    # Multiaddress of the destination peer is fetched from the peerstore using 'peerId'.
+    stream = await host_b.new_stream(host_a.get_id(), [PROTOCOL_ID])
+    received = []
+    # while True: Seems the close stream event from the other host is not received
```
robzajac: Making a note that stream close functionality is not fully implemented yet - #92

Issue #60, opened 2018-11-16
Plan to move this repo into the libp2p org


raulk:
Hey all! It was great to meet @zixuanzh and @stuckinaboot at devcon4. I enjoyed our brainstorming and hacking sessions on py-libp2p.

We agreed with @pipermerriam that moving this repo to the `libp2p` organisation is appropriate, once the PoC was delivered and the EF grant is approved. This will help discoverability and will encourage people to contribute, thus having better chances at sparking a community. In fact, there's [a thread](https://github.com/ipfs/py-ipfs/issues/1) of Python contributors willing to help.

I'm capturing some thoughts on the path to migrating this repo to the libp2p org.

- [x] A **NOTICE** on the frontpage warning that this is an experimental and work-in-progress repo under heavy development.
- [x] An sponsorship acknowledgement to the Ethereum Foundation on the frontpage once the grant becomes official.
- [X] ~~README formatted according to https://github.com/RichardLitt/standard-readme, which is what we lean towards in the libp2p/IPFS universe.~~
- [x] A breakdown of features and development status/estimation, in a formatted table. This helps people assess the maturity and lend a hand in those features that are WIP. Maybe a bunch of issues tagged `help-wanted` to lessen the decision burden for willing contributors.
- [x] Evaluate existing py-multiaddr implementations, continue maintaining those in the multiformats org: https://github.com/multiformats/multiaddr/issues/7#issuecomment-439380831
- [x] Continuous Integration.
- [x] Badges pointing to the libp2p IRC channel, CI, and else.

raulk:
I stroke out the standard README point. Apparently we tried that for a while, but we didn't get a critical mass, so it's OK to drop this ask.

mgoelzer:
LGTM.  My only suggestion is to add an Apache 2 license.  I have opened PR https://github.com/zixuanzh/py-libp2p/pull/61 to do this.

raulk:
@sbuss' py-multiaddr now lives in the multiformats org! https://github.com/multiformats/py-multiaddr

One more down in the checklist.

PR #98, opened 2018-12-22
Adding feature breakdown to readme

robzajac:
As per #60, we want a breakdown of features in development in py-libp2p. This one is inspired by the [libp2p module breakdown](https://github.com/libp2p/libp2p/blob/master/REQUIREMENTS.md#libp2p-modules-implementations) for other languages, with focus on the modules that we want to develop in the near future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,16 +45,31 @@ py-libp2p aims for conformity with [the standard libp2p modules](https://github.
 
 | Transport Protocols                          | Status        |
 | -------------------------------------------- | :-----------: |
-| **`TCP`**                                    | :green_apple: |
+| **`TCP`**                                    | :lemon: tests |
 | **`UDP`**                                    | :tomato:      |
 | **`WebSockets`**                             | :tomato:      |
 | **`UTP`**                                    | :tomato:      |
+| **`WebRTC`**                                 | :chestnut:    |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,9 +28,67 @@ pytest
 ```
 Note that tests/libp2p/test_libp2p.py contains an end-to-end messaging test between two libp2p hosts, which is the bulk of our proof of concept.
 
-# Explanation of Basic Two Node Communication
+## Feature Breakdown
+py-libp2p aims for conformity with [the standard libp2p modules](https://github.com/libp2p/libp2p/blob/master/REQUIREMENTS.md#libp2p-modules-implementations). Below is a breakdown of the modules we have developed, are developing, and may develop in the future.
 
-## Core Concepts
+> Legend: :green_apple: Done &nbsp; :lemon: In Progress &nbsp; :tomato: Missing &nbsp; :chestnut: Not planned
```
zixuanzh: For completeness, I think we should also include features that are not planned at the moment. We should also put :lemon: for features that have not gone through more thorough testing.

robzajac: Thanks - just added the other features and 🍋 for modules we want to test more. Please take another look.

PR #101, opened 2018-12-29
WIP: Refactor to use multistream_select

dheatovwil:
I've forked the multistream_select repo, updated it with latest code in libp2p, and re-referenced the package.
I'm quite new so I'm not entirely sure is everything done, comments are appreciated.


robzajac:
Hi @dheatovwil, thanks for your contribution. Since the py-multistream-select repo simply uses the multistream-select code from py-libp2p, I'm not sure it needs to be included as a separate package and dependency. This also helps us stay consistent with the style of the rest of the repo (modules contained within the py-libp2p repo itself instead of separate repos). Let us know your thoughts.

dheatovwil:
From  [py-ipfs/issues/49](https://github.com/ipfs/py-ipfs/issues/49) 
> Factor out https://github.com/zixuanzh/py-libp2p/tree/master/protocol_muxer into a separate `py-multistream-select` library and update `py-libp2p` to use it (**Easy!**)
> 
>     * Current `py-multistream-select` library: https://github.com/ngetahun/py-multistream-select
I am not sure about @alexander255's intention, from my observation it seems adhering to go-libp2p structure

robzajac:
Thanks - we'll wait for followup from @alexander255.

dheatovwil:
No worries, so now I gotta package it to PyPI, is this how it works?

alexander255:
@dheatovwil: Yes, that's how it works :slightly_smiling_face: 

You likely also want to integrate https://github.com/zixuanzh/py-libp2p/blob/master/tests/protocol_muxer/test_protocol_muxer.py into your codebase.

I will look into getting your repo moved to the `multiformats` org. Please also add me and @robzajac as project owners on PyPI to increase the [bus factor](https://en.wikipedia.org/wiki/Bus_factor). (My handle is `ntninja` there & everywhere else.)

@robzajac: I hope you don't me/us moving ahead with this so quickly – I'm just excited to see some progress here. :grin:

alexander255:
BTW: I also recommend using `flit` rather than `setuptools` for packaging pure-python modules; the experience of packaging is much nicer with it.

Either way you'll want to set `python_requires = ">=3.5"` and drop the 2.7 trove classifiers. If you decide to continue using setuptools, please also add `long_description_content_type="text/markdown"` to `setup.py` to ensure that the description will be properly parsed by PyPI (not required with `flit`).

dheatovwil:
Thanks for the tips, updated the package and repo as you guys mentioned, working on the testcase now.

alexander255:
@dheatovwil: Were you able to move forward on this since your last comment?

dheatovwil:
@alexander255 Sorry about the late reply, the progress will be a little slow due to other events going on at the same time,  I think I can finalize everything in a week or two.

dheatovwil:
@alexander255 @robzajac I've requested to transfer the ownership to @stuckinaboot.  I've also gone through the testcases and travis pipeline, everything seems fine.  The testcase in this repo is updated to use the `multistream-select` package on pypi, and the coverage of that repo is at 100%.  I've also updated with all previous changes and fixed the conflicts, is there anything else I am missing?

alexander255:
@dheatovwil: LGTM. I looked over the source code and it appears to be well designed and the test are decent. However, as you are probably aware, messages are not actually length-encoded and there are no tests checking whether the correct wire format is actually produced. I also filed an [issue on the multistream spec](https://github.com/multiformats/multistream-select/issues/16) since I noticed that your source, just like everybody else's, assumes that messages are UTF-8 encoded, but that isn't actually written anywhere.

stuckinaboot:
@dheatovwil Thanks for working on this. I looked through your code and the update you made to implement the 'ls' command and have one piece of feedback. As of now, when 'ls' is seen in `negotiate` of the multiselect.py file, you send back each supported protocol as it's own message. Was there a particular reason for that (maybe I missed something in the go code)? I believe that is less efficient because each message adds overhead and the other party would have to know how many messages to expect. Can you please update this so that a line-delimited list of protocols is sent in a single message when 'ls' is received (e.g. `"protocol1\nprotocol2\nprotocol3"`)? 

Also, in order to merge your PR, we'll need your `py-multistream-select` repo moved to the multiformats org so that it becomes canonical.

dheatovwil:
@stuckinaboot Yes you are right, I've checked against go and js code, it's an honest mistake.  I've fixed it.  

Regarding that, I tried to transfer the ownership to multiformats but it says I don't have the permission to do so.  Should I talk to the guys at multiformats regarding this issue?

alexander255:
Bump. Has there been any update on the procedural issues around this?

Issue #97, opened 2018-12-21
Why not use kademlia as a submodule?


csunny:
Recently,  I want to use py-libp2p in my  [py-bitcorin](https://github.com/csunny/py-bitcoin) project,  and i notice this lib replace origin dht with kademlia project.  it is a better solution with git submodule, is it?

robzajac:
That's a good point. We originally copied in the module so we could make changes to it. A better solution might be for us to make a fork of the Kad project and include it in py-libp2p as a Git submodule.

Truthfully, we haven't found a use for Kademlia yet in the existing py-libp2p feature set, so we'll investigate this further once we start to expand our implementation (most likely when doing peer discovery).

magniff:
Why not use https://github.com/bmuller/kademlia implementation as a site-package? Just add it to the `install_requires` list, setuptools will autoinstall kademlia stuff with a proper versions as a dependency.

alexh:
We will not be importing https://github.com/bmuller/kademlia as a side package because we need to make deep changes to his implementation to suite our needs

PR #105, opened 2019-01-09
Feature/issue 104

magniff:
- Migrate to project structure, proposed at issue-104
- Tune up .travis config a bit.


magniff:
Ok, everything seems to be fine. Apparently pylint ignore path option is still broken, according to this issue https://github.com/PyCQA/pylint/issues/2541

.travis.yml
```diff
@@ -4,15 +4,18 @@ python:
...
+  - pip install --upgrade pip
+  - pip install "pytest>=3.6"
+  - pip install codecov pytest-cov pytest-asyncio pylint
+  - python setup.py develop
 
 script:
-  - pytest --cov=./ -v
-  - pylint --rcfile=.pylintrc encryption host libp2p network peer stream_muxer transport tests
+  - pytest --cov=./libp2p tests/
+  - pylint --rcfile=.pylintrc libp2p/!(kademlia|protocol_muxer) tests
```
robzajac: Good catch; we actually only wanted to exclude `kademlia` (it's an external implementation that we haven't used yet; we will likely move to a Git submodule soon (#97)). We can include `protocol_muxer` in linting after #106 lands.

.travis.yml
```diff
@@ -4,15 +4,18 @@ python:
   - "3.6"
 
 install:
-  - pip install -r requirements.txt
-  - pip install codecov pytest pytest-cov
+  - pip install --upgrade pip
+  - pip install "pytest>=3.6"
+  - pip install codecov pytest-cov pytest-asyncio pylint
```
robzajac: Could we have these dependencies in `setup.py` as well?  I can't remember or find docs about why we put them here instead.

mhchia: IMO [the style](https://github.com/ethereum/py-evm/blob/master/setup.py#L31-L41) might be pretty suitable, or even put them in a separate `requirements_dev.txt`

.travis.yml
zixuanzh: just approved #106

zixuanzh:
Should be good to go once `travis.yml` is updated.

.travis.yml
magniff: `Requirements_dev` is a good option, thumbs up!) I will have a chance to add it only at the late evening, though. Can we merge this code as is and add missing deps in a separate PR or something?

PR #115, opened 2019-01-30
Fixing setup.py

Jorropo:
Adding this allow users to lunch using ./setup.py install, wich is simpler than python{3} setup.py install.


Jorropo:
In fact this is more version dependent that I was thinking, the user must know what it runs so this isn't a good idea.

PR #116, opened 2019-01-31
DISCUSSION: Code refactor change all crypto lib to python-cryptography

Jorropo:
[Python-Cryptography](https://github.com/pyca/cryptography) is an C binding of openssl library and this is LOT more faster than Crypto and MultiHash, see your self (generating a new node key) :

|                     | Crypto | Python Cryptography |
| :-----------------: | ------ | ------------------- |
| MultiHash           | 0.94s  | 0.2s                |
| Python Cryptography | 0.93s  | 0.15s               |

Horizontaly : RSA key gen
Verticaly : sha2_256 hashing

[Python-Cryptography](https://github.com/pyca/cryptography) also have a full python implementation for be compatible on platform where the C lib isn't available.
It also give lot more of control than MultiHash or Crypto.
Python-Cryptography have just one bad point, it is huge ! This lib include so many hashing and encryption algo that importing all the lib can make performance drop, but it can easly solved by using `from xxxxx import yyyyy` and go deeper than possible (never import any module).

I would have your feelings about that (the ranking of choice is my view):
 - Should I implement cryptography directly ? (2nd choice)
 - Should I just create some (deprecated ?) functions to convert Crypto key object to cryptography and implement cryptography. So users of py-libp2p can choose what to use ? (best choice)
 - Should I create a proxy library simplifying use of cryptography (with minor performance loss (still lot better than now)). (Worst choice)
 - Other ? (It depends)

**TO DO :**
- [X] Do an example file of cryptography. (tests/benchCrypto.py)
- [ ] **Discuss with other maintainer, contributors of py-libp2p.** (in progress)
- [ ] Replace all hashing function to cryptography.
- [ ] Replace all RSA use to cryptography.

### DIFF
--- a/tests/benchCrypto.py
+++ b/tests/benchCrypto.py
@@ -1,3 +1,6 @@
+#!/usr/bin/env python3.6
+# pylint: skip-file
+
 from Crypto.PublicKey import RSA
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.asymmetric import rsa
@@ -6,6 +9,11 @@ from cryptography.hazmat.primitives.hashes import SHA256, Hash
 from time import time
 import multihash
 
+SAMPLES = 10
+
+def formatDuration(t):
+    return str(round(t, 2)) + "s"
+
 def makeItTime(func, count):
     i = 0
     time1 = time()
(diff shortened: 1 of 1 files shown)


ralexstokes:
i'd be in favor of finding higher performance cryptography libraries relative to the one we are currently using (`pycryptodome`) but this PR seems to have gotten stale -- i'll go ahead and close it. feel free to re-open if you want or we can also move this discussion to an issue

PR #117, opened 2019-01-31
Change webrtc to planned

Jorropo:
Change WebRTC transport to planned.
WebRTC transport in py-libp2p is very important for me (this is the reason why I'm here).
So be sure, I will do it when libp2p will be more developed.
Also I changed nat-pmp and upnp because this is quite simple and I will probably do it (if nobody do it before me :smile: ).

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ py-libp2p aims for conformity with [the standard libp2p modules](https://github.
 | **`UDP`**                                    | :tomato:      |
 | **`WebSockets`**                             | :tomato:      |
 | **`UTP`**                                    | :tomato:      |
-| **`WebRTC`**                                 | :chestnut:    |
+| **`WebRTC`**                                 | :tomato:      |
 | **`SCTP`**                                   | :chestnut:    |
 | **`Tor`**                                    | :chestnut:    |
 | **`i2p`**                                    | :chestnut:    |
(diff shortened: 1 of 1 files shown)


Jorropo:
Hi, @robzajac 
I'm a student in a computer school, each week end all computers are reset and the monday everybody loss 3h of software downloading because 300 peoples downloading chrome, firefox, visual studio, film, ... on a 500Mbits/s connection is pain full. (The administrator limited each outbound connection to 250KBytes/s), Also we can't use any regular BitTorrent file sharing tool because each rooms is in a private subnet.
So I'm trying to make an distributed app store accessible through web and through an desktop app with auto installation system.
At start my choice was web torrent, but web torrent doesn't provide any good broadcast system for the app store.
So I looked IPFS, that quite good, provide broadcast, distributed, ..., and in the next release of js-ipfs we will be able route with DHT.
So for the web app choice is fast but for the desktop, hummm.
I can't make WebRTC work on electron on computer's school, work good on mine but not on these computers, I only saw that on this computers so I think that a configuration in the windows10 blocking that.
So for making desktop app I only know Electron and Python, so my next choice is python.
But for that I need ipfs in python, but for ipfs in python I need libp2p in python, compatible with js-ipfs, that will work better :smiley: .

And for nat-pmp and upnp, that simple and this is a must.

Also WebRTC for Ethereum 2.0 could be good things, because it do less centralization. Node without port opening capabilities will be able to serve as a full node.

PR #120, opened 2019-02-08
Fixed some issues with the chat example

AliabbasMerchant:
The chat wouldn't work properly, and read `None`.  
It is working fine now.  
Also, I simplified the command line arguments thing.

### DIFF
--- a/examples/chat/chat.py
+++ b/examples/chat/chat.py
@@ -19,7 +19,7 @@ async def read_data(stream):
             if read_string != "\n":
                 # Green console colour: 	\x1b[32m
                 # Reset console colour: 	\x1b[0m
-                print("\x1b[32m %s\x1b[0m > " % read_string, end="")
+                print("\x1b[32m %s\x1b[0m " % read_string, end="")
 
 
 async def write_data(stream):
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: python
 
-python:
-  - "3.6"
+matrix:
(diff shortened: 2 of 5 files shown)


examples/chat/chat.py
```diff
@@ -12,34 +13,28 @@
...
-            # Green console colour: 	\x1b[32m
-            # Reset console colour: 	\x1b[0m
-            print("\x1b[32m%s\x1b[0m> " % read_string, end="")
+        read_string = await stream.read()
+        if read_string is not None:
+            read_string = read_string.decode()
+            if read_string != "\n":
+                # Green console colour: 	\x1b[32m
+                # Reset console colour: 	\x1b[0m
+                print("\x1b[32m %s\x1b[0m > " % read_string, end="")
```
zixuanzh: Can we change this line to `print("\x1b[32m%s\x1b[0m" % read_string, end="")`? The ` `  and `>` can sometimes be confusing when a user is sending two consecutive messages. Thanks

AliabbasMerchant: `print("\x1b[32m %s\x1b[0m " % read_string, end="")`  
I have removed the `>`, but have kept the spaces. This is because, the color codes do not work on Windows, and the raw `\x1b[0m` and `x1b[32m` look odd, and get mixed up with the message. The space helps to separate them.

zixuanzh: I see, we will fix the UI issues later at some point. thanks!

PR #121, opened 2019-02-09
Bump version to Python 3.7

### LINKED ISSUE #88: Moving to Python 3.7
Python 3.7 brought [some improvements to asyncio](https://docs.python.org/3/whatsnew/3.7.html#asyncio), but we have waited to make the upgrade to ensure compatibility with py-evm. But it seems py-evm [has upgraded](https://github.com/ethereum/py-evm/blob/master/.circleci/config.yml#L337) - is py-libp2p ready to upgrade as well?

zixuanzh:
close #88 and unblock #114

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 
 python:
-  - "3.6"
+  - "3.7"
 
 install:
   - pip install --upgrade pip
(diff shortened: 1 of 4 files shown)


robzajac:
LGTM, good move.

PR #123, opened 2019-02-10
Peer Discovery Interface

alexh:
I have added a peer discovery interface modeled after the interface used in the [go repo](https://github.com/libp2p/go-libp2p-discovery/blob/master/interface.go)

### DIFF
--- a/libp2p/discovery/advertiser_interface.py
+++ b/libp2p/discovery/advertiser_interface.py
@@ -1,4 +1,5 @@
 from abc import ABC, abstractmethod
+# pylint: disable=too-few-public-methods
 
 class IDiscoverer(ABC):
 
@@ -12,3 +13,4 @@ class IDiscoverer(ABC):
         :param service: service that you provide
         :raise Exception: network error
         """
+        
--- a/libp2p/discovery/advertiser_interface.py
+++ b/libp2p/discovery/advertiser_interface.py
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 # pylint: disable=too-few-public-methods
 
-class IDiscoverer(ABC):
+class IAdvertiser(ABC):
 
     def __init__(self):
         pass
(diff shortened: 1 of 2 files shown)


zaibon:
I actually have some code layout around that implement these interface as well as some basic testing around kadmelia package. I'll clean this up and create a PR

libp2p/discovery/discoverer_interface.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(self):
+        pass
+
+    @abstractmethod
+    def find_peers(self, service):
+        """
+        Find peers on the networking providing a particular service
+        :param service: service that peers must provide
+        :return: peerstore containing found peers on the network
```
zaibon: If we look at the return value of the go implementation: https://github.com/libp2p/go-libp2p-discovery/blob/master/interface.go#L19
`find_peers` could return a generator or `PeerInfo` instead of a `PeerStore`
That would prevent to load too much in memory in case there are a lot of peers to be discovered

libp2p/discovery/advertiser_interface.py
```diff
@@ -0,0 +1,16 @@
+from abc import ABC, abstractmethod
+# pylint: disable=too-few-public-methods
+
+class IDiscoverer(ABC):
```
zaibon: shouldn't it be : `IAdvertiser` ?

alexh: Yes

libp2p/discovery/discoverer_interface.py
alexh: Resolved in new commit

alexh:
> I actually have some code layout around that implement these interface as well as some basic testing around kadmelia package. I'll clean this up and create a PR

@zaibon We would welcome a PR on this, but in the meantime our team will be focusing on pubsub until the discv5 spec is more solidified, then we will implement discv5.

PR #125, opened 2019-02-22
Adding sponsorship acknowledgement to EF

robzajac:
Now that the sponsorship has been announced (see link in blog), we can add acknowledgement and close out #60.


mhchia:
Congrats guys! Awesome work!

PR #140, opened 2019-03-20
WIP: fix 136

### LINKED ISSUE #136: Wire up and test close()
Currently, we have close() in `tcp.py`, `mplex.py`, `mplex_stream.py`, `net_stream.py` but they are not properly wired up. We want to be able to do the following in `swarm.py`.

- [ ] close_peer(peer_id): close all connections to a given peer
- [ ] tearDown(): close all listeners by calling something like `transport.listener.close()` and close all connections. It should also prevent new connections and/or listeners from being added to the swarm.

We would also want to have test cases on close() in `muxed_stream()`.

Jorropo:
This pr is to fix #136 
I take the liberty to add shutdown and wait_closed
so if we have a list of transport we can do :
```python
a = [example_transport, example_transport2]
for i in a:
    i.shutdown()
for i in a:
    await i.wait_closed()
```
This code can close a lot of transport concurrently (so speed up the whole thing).
Goal :
- [ ] Allow close a lot of things :
  - [X] tcp.py
  - [X] raw_connection.py


Jorropo:
@zixuanzh I saw that the close() start a new event loop and close it after.
Can I replace that by settting close() to async and asking user to lunch libp2p in an event loop, that simplify dev and allow us to use `do()` via `asyncio.create_task()` this is the same as `setTimeout()` or `go` in go, except it will run in a single thread so create_task will run at next await (new in python3.7).
Here a small example to see how it works : (require python >= 3.7)
```python
import asyncio
from asyncio import create_task as do

async def test1(n):
    print("Test 1 is running and will fetch datas for node " + str(n))
    print("Test 1." + str(n) +" is requesting somethings in the network, here a sleep.")
    await asyncio.sleep(n)
    print("Test 1." + str(n) +" have finished and is adding some datas to peerstores")

async def test2():
    print("Test 2 is running")
    print("Test 2 is starting test 1 to query node 2")
    t1 = do(test1(2))
    print("Test 2 is starting test 1 to query node 1")
    t2 = do(test1(1))
    print("Test 2 is requesting somethings on the network, here simulated by a sleep.")
    await asyncio.sleep(1)
    print("Test 2 have finished and will now wait for test 1 to finish")
    await t1
    await t2
    print("Everythings is finish right now")

asyncio.run(test2())
```
This code produce this output (sleep time can be random and python will continue the first the retrieve his data):
```
Test 2 is running
Test 2 is starting test 1 to query node 2
Test 2 is starting test 1 to query node 1
Test 2 is requesting somethings on the network, here simulated by a sleep.
Test 1 is running and will fetch datas for node 2
Test 1.2 is requesting somethings in the network, here a sleep.
Test 1 is running and will fetch datas for node 1
Test 1.1 is requesting somethings in the network, here a sleep.
Test 2 have finished and will now wait for test 1 to finish
Test 1.1 have finished and is adding some datas to peerstores
Test 1.2 have finished and is adding some datas to peerstores
Everythings is finish right now
```
But for that to work we have a single big event loop and don't start a new one when we need like that :
https://github.com/libp2p/py-libp2p/commit/78d0a658bb71bd6081db2d19e2704181e5287d6b#diff-bdded13ba663e4ee26402b7481d8e855R61

That also very good for users because they can have their app running as side of kademlia's crawler (just lunch kademlia with asyncio.create_task in the init of the main node).

robzajac:
@Jorropo making this change would break the way we expect users to interact with libp2p. If I understand correctly, this change would force users to lock into `asyncio`. We want the concurrency layer to be abstracted away from users, and so all `asyncio` logic should occur within libp2p. This way a user could even use a different scheduling library at a higher level.

For everything else, we can move forward with adding more `close` logic to existing parts of the library. The changes you have made in this PR thus far look good!

Jorropo:
@robzajac, I have an other question.
Due to how asyncio works I can produce 2 close() fonction per object.
One that wait until the connection is really closed and one that only start the close in background (the function still not async).
I personally made close() blockant and shutdown() start closing then return, Is that good ?

PR #141, opened 2019-04-01
Floodsub RPC

zixuanzh:
Phased out custom message.py and using standard RPC format

Note that there are at least two places that are still not fully compliant as per spec [here](https://github.com/libp2p/specs/tree/master/pubsub#the-message):
- [ ] `from` is a key word in Python so we changed `from` to `from_id` in proto.
- [ ]  `seqno` should be a linearly increasing number but now it's just a uuid


stuckinaboot:
generate_message_id in pubsub is outdated

tests/pubsub/test_floodsub.py
```diff
@@ -45,68 +44,15 @@
...
 
     # Check that the msg received by node_b is the same
     # as the message sent by node_a
-    assert res_b == msg.to_str()
-
-    # Success, terminate pending tasks.
-    await cleanup()
-
-@pytest.mark.asyncio
-async def test_simple_three_nodes():
```
robzajac: Just documenting, why did we remove this test? Too repetitive?

libp2p/pubsub/pb/rpc.proto
```diff
@@ -0,0 +1,76 @@
...
+	message SubOpts {
+		optional bool subscribe = 1; // subscribe or unsubcribe
+		optional string topicid = 2;
+	}
+
+	optional ControlMessage control = 3;
+}
+
+message Message {
+	optional bytes from_id = 1;
```
robzajac: I would just point out that `from` vs. `from_id` will be a critical thing to figure out for interop. This API won't work with other libp2p clients because of `from_id`. I think it's fine for now but worth keeping in mind for the future.

libp2p/pubsub/pubsub.py
```diff
@@ -89,60 +56,54 @@ def __init__(self, host, router, my_id):
...
+            incoming = (await stream.read())
+            rpc_incoming = rpc_pb2.RPC()
+            rpc_incoming.ParseFromString(incoming)
+
+            should_publish = False
+
+            if rpc_incoming.publish:
+                # deal with RPC.publish
+                for message in rpc_incoming.publish:
+                    if message.seqno not in self.seen_messages:
```
robzajac: Isn't it true that the [`seqno` could be duplicated across clients, but the corresponding messages could be different](https://github.com/libp2p/specs/tree/master/pubsub#the-message)? E.g. if A sends message "foo" with `seqno=1` and B sends "bar" with `seqno=1". In that case, would we fail to forward a message that we should have forwarded? I might be misunderstanding this code, if so feel free to clarify.

tests/pubsub/test_floodsub.py
zixuanzh: @robzajac  yes, this test is covered by `test_three_nodes_two_topics_test_obj`

libp2p/pubsub/pubsub.py
zixuanzh: this is correct, @stuckinaboot let's fix this one in #144 ?

rpc.proto
```diff
@@ -0,0 +1,78 @@
+// Borrowed from https://github.com/libp2p/go-libp2p-pubsub/blob/master/pb/rpc.proto
```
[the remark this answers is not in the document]
zixuanzh: I would say we still keep this for two reasons, 1) rpc.proto is a lot easier to read than generated stub 2) we might still need to modify rpc.proto to generate new stub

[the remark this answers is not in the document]
robzajac: I agree with both 1) and 2), though I mean to say it looks like the rpc.proto file is duplicated. There's this one under the root dir, and there's the one under the pubsub/pb dor. Do we need both?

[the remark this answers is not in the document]
zixuanzh: oh ok will remove one of them

Issue #150, opened 2019-04-18
Kademlia Implementation Roadmap


zixuanzh:
We just landed #129 to work on Kademlia feature by feature. Thanks @zaibon! Here's a summary of what needs to be done in descending order.

- [x] replace (ip, port, node_id) tuple with multiaddr of ip, port, node_id
- [x] implement RoutedHost with Kad DHT
- [x] refactor dial by injecting Router as a dependency
- [x] implement ADD_PROVIDER and REMOVE_PROVIDER
- [ ] implement content routing with Kademlia
- [ ] implement peer discovery by extending content routing as discussed [here](https://github.com/libp2p/py-libp2p/pull/129#issuecomment-468999482) and relevant to #128 
- [ ] phase out rpcudp with gRPC

pacrob:
Closing in favor of #540

PR #153, opened 2019-04-20
Added RoutedHost

alexh:
As per #150, implement RoutedHost

### DIFF
--- a/libp2p/__init__.py
+++ b/libp2p/__init__.py
@@ -6,8 +6,10 @@ from .peer.peerstore import PeerStore
 from .peer.id import id_from_public_key
 from .network.swarm import Swarm
 from .host.basic_host import BasicHost
+from .kademlia.routed_host import RoutedHost
 from .transport.upgrader import TransportUpgrader
 from .transport.tcp.tcp import TCP
+from .kademlia.network import KademliaServer
 
 
 async def cleanup_done_tasks():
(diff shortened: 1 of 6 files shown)


libp2p/__init__.py
```diff
@@ -69,15 +95,22 @@ def initialize_default_swarm(
...
             muxer_opt=muxer_opt, sec_opt=sec_opt,
             peerstore_opt=peerstore_opt)
+    if not disc_opt:
+        disc_opt = initialize_default_kademlia(ksize=20, alpha=3,
+                                               id_opt=id_opt, storage=None)
 
     # TODO enable support for other host type
     # TODO routing unimplemented
-    host = BasicHost(swarm_opt)
+    host = RoutedHost(swarm_opt, disc_opt)
```
zixuanzh: I think we should still provide the option for only starting a BasicHost. Can we only return a BasicHost if no disc_opt is passed in and a RoutedHost only when a disc_opt is introduced?

libp2p/kademlia/routed_host.py
```diff
@@ -0,0 +1,21 @@
...
+
+class RoutedHost(BasicHost):
+    def __init__(self, _network, _kad_network):
+        super(RoutedHost, self).__init__(_network)
+        self.kad_network = _kad_network
+
+    def get_kad_network(self):
+        return self.kad_network
+
+    def kad_listen(self, port, interface='0.0.0.0'):
```
zixuanzh: Should we call these functions routed_listen instead of kad_listen since kademlia is not the only way to do routing.

raulk:
Hey folks, we're moving away from the _specialised host_ construction in go-libp2p towards a service-based host construction wired with dependency injection and IoC (inversion of control). See: https://github.com/libp2p/go-libp2p/issues/467.

With the modular dialer, instead of wrapping a host within a RoutedHost, attaching a DHT to the host with a peer discovery capability would inject a component in the dialer to resolve addrs from the DHT when a dial is carried out.

alexh:
Thanks for the heads up @raulk! We will add refactoring this PR using dependency injection and IoC to the roadmap. We should have this done in the next few days!

libp2p/kademlia/routed_host.py
alexh: Fixed!

libp2p/__init__.py
alexh: Fixed!

PR #157, opened 2019-04-25
Refactored RoutedHost into Injected Router

alexh:
As per https://github.com/libp2p/py-libp2p/pull/153#issuecomment-485403029, RoutedHost has been refactored into a dependency injection IoC model where we inject a router implementing using kademlia DHT into swarm.


libp2p/__init__.py
```diff
@@ -105,14 +107,13 @@ def initialize_default_swarm(
         swarm_opt = initialize_default_swarm(
             id_opt=id_opt, transport_opt=transport_opt,
             muxer_opt=muxer_opt, sec_opt=sec_opt,
-            peerstore_opt=peerstore_opt)
+            peerstore_opt=peerstore_opt, disc_opt=disc_opt)
+
+    swarm_opt.add_router(disc_opt)
```
zixuanzh: I think we can take out this line as disc_opt should have been passed into Swarm in Swarm construction at line 83.

libp2p/host/basic_host.py
```diff
@@ -11,21 +11,22 @@
 class BasicHost(IHost):
 
     # default options constructor
-    def __init__(self, _network):
-        self.network = _network
-        self.peerstore = self.network.peerstore
+    def __init__(self, network, router=None):
+        self._network = network
+        self.peerstore = self._network.peerstore
```
zixuanzh: let's place this line after line 17 to look cleaner

libp2p/routing/kademlia/kademlia_peer_router.py
```diff
@@ -0,0 +1,37 @@
...
+class KadmeliaPeerRouter(IPeerRouting):
+    # pylint: disable=too-few-public-methods
+
+    def __init__(self, dht_server):
+        self.server = dht_server
+
+    async def find_peer(self, peer_id):
+        """
+        Find specific Peer
+        FindPeer searches for a peer with given peer_id, returns a peer.PeerInfo
```
zixuanzh: it is returning a KadPeerInfo here from decode_peerinfo

PR #159, opened 2019-04-27
Implement security

stuckinaboot:
Implement security interfaces, security module selection, and basic secure transport (which are used to take a normal connection and turn that into a secure connection). Integrate security selection and securing a connection into the "establish a connection between two nodes" pipeline. For integrating into this pipeline, per [forums](https://discuss.libp2p.io/t/multistream-security/130), we first want to select a security module and _then_ agree on a multiplexer.

- [x] Implement security_multistream (agrees on a secure transport with the party at the other end of conn)
- [x] Test security_multistream using a basic secure transport implementation
- [x] Integrate security_multistream into "establish a connection" pipeline


libp2p/security/security_multistream.py
```diff
@@ -0,0 +1,88 @@
...
+
+    async def select_transport(self, conn, initiator):
+        """
+        Select a transport that both us and the node on the
+        other end of conn support and agree on
+        :param conn: conn to choose a transport over
+        :param initiator: true if we are the initiator, false otherwise
+        :return: selected secure transport
+        """
+        # TODO: Is conn acceptable to multiselect/multiselect_client
```
zixuanzh: multiselect takes in a stream in our repo, i can buy the argument of passing in a raw conn since security upgrade should happen before transport upgrade but we may have to refactor some of the multiselect code. To be exact, MultistreamCommunicator expects a stream as input but we are not treating it any differently than a raw connection in our code.

stuckinaboot: We don’t need to treat it different because raw conn has the same read and write functions as stream now and in python if it walks like a duck and talks like a duck it’s a duck

zixuanzh: I know, but the naming in our code can be confusing to future contributors.

stuckinaboot: So what are you saying we should change?

zixuanzh: this could work as a short-term fix but i feel like if it doesnt need to be a stream, we should probably name it something else. I dont have any better idea off the top of my head.

stuckinaboot: True, we could have an interface with a `read` and `write` function.

libp2p/security/secure_transport_interface.py
```diff
@@ -0,0 +1,27 @@
...
+    @abstractmethod
+    async def secure_inbound(self, conn):
+        """
+        Secure the connection, either locally or by communicating with opposing node via conn,
+        for an inbound connection (i.e. we are not the initiator)
+        :return: secure connection object (that implements secure_conn_interface)
+        """
+
+    @abstractmethod
+    async def secure_outbound(self, conn, peer_id):
```
robzajac: Do we actually need the peer_id for this function? I didn't see either implementation use it.

libp2p/security/insecure_security.py
```diff
@@ -0,0 +1,44 @@
+from libp2p.security.secure_transport_interface import ISecureTransport
+from libp2p.security.secure_conn_interface import ISecureConn
+
+class InsecureTransport(ISecureTransport):
```
robzajac: Is the InsecureTransport/InsecureConn needed or is it just for testing? I felt as though if `sec_opt` is `None` then we should just avoid anything to do with security instead of using something like an InsecureTransport. At least, it might be confusing for this to be a subtype of ISecureTransport.

stuckinaboot: This follows the precedent set forth by go. The idea is that you _must_ choose a security module, but you are allowed to choose one that doesn't do anything.

libp2p/security/secure_transport_interface.py
stuckinaboot: This was a go interface so I just followed the precedent.

libp2p/security/security_multistream.py
stuckinaboot: Added the comment in this PR. If we want to refactor that could be done in another PR since there's other classes that we may want to implement that sort of reader+writer interface

PR #165, opened 2019-05-07
update green apples

zixuanzh:
In Progress:
- [ ] Identify
- [ ] Content Routing with kad
- [ ] Peer Discovery with kad

Completed:
- [ ] Floodsub
- [ ] Gossipsub


stuckinaboot:
We can move tcp and swarm into green status right?

zixuanzh:
> I agree with these, lgtm. Though I'm surprised we don't have an entry for security? Or is that `Identify`?

Looks like that's not part of the green apples list and it's different from Identify.
cc @mgoelzer 

> We can move tcp and swarm into green status right?

Yes, we already did.

PR #167, opened 2019-05-16
Pin the version of multiaddr we require.

ralexstokes:
Temporary workaround for https://github.com/multiformats/py-multiaddr/issues/47

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setuptools.setup(
         "click",
         "base58",
         "pymultihash",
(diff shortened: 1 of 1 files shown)


ralexstokes:
Users should note that until https://github.com/multiformats/py-multiaddr/issues/47 is resolved, you will need the patch on this branch to run any of the code

stuckinaboot:
Good fix. I'm going to merge this and then #166 should pass too and will be ready to merge.

PR #168, opened 2019-05-19
add dependencies to correctly build

zimio:
Some python dependencies needed to be added to do the build correctly.

Also temporarily adds work around for py-multiaddr,


mhchia:
Thanks for the work. However, we changed a lot in the codebase and this PR is obsolete now.

Issue #169, opened 2019-05-22
libp2p node interoperability


stuckinaboot:
In order to test interoperability with other non-Python libp2p nodes, the first spec that should be ensured is that the libp2p open connection and open stream mechanisms interoperate.

In py-libp2p, the following messaging occurs when Node A and Node B open a new connection:

1) A dials B
2) A sends its peer ID to B and B sends its peer ID to A (this is where the Identify protocol would come into place, which still needs to be implemented to support different Identify services)
3) A and B agree on a security transport module
4) A and B agree on a multiplexer

Then, when Node A opens a new stream to Node B, the following messaging occurs in py-libp2p:

1) A sends open stream message to B
2) A and B agree on a particular protocol to use for communication across that stream

All PubSub mechanisms in libp2p rely on the underlying stream communication so imo it would be best to ensure opening connections and streams to non-python nodes works before moving on to testing PubSub’s interoperability

mhchia:
Closed by several issues before, and py-libp2p now can talk with other clients now!

PR #175, opened 2019-06-21
[WIP] Connection and stream interop between Python and Go

mhchia:
**[Sill work in progress]**

Fixes https://github.com/libp2p/py-libp2p/issues/169
Try to make py-libp2p talk with go-libp2p. Currently, in this PR I removed "sending peer id" from python, and added delimited read/write, to make python and go pass agree on the same multiselect protocol and security. Still need to confirm whether it really works.

TODOs
- [ ] Confirm if the connection is done both sides.
- [ ] Find out when we should send `peer_id`.
- [ ] Make `new_stream` work.

### DIFF
--- a/libp2p/__init__.py
+++ b/libp2p/__init__.py
@@ -76,14 +76,14 @@ def initialize_default_swarm(
     # TODO TransportUpgrader is not doing anything really
     # TODO parse muxer and sec to pass into TransportUpgrader
     muxer = muxer_opt or ["mplex/6.7.0"]
-    sec = sec_opt or {"insecure/1.0.0": InsecureTransport("insecure")}
+    sec = sec_opt or {"/plaintext/1.0.0": InsecureTransport("insecure")}
     upgrader = TransportUpgrader(sec, muxer)
 
     peerstore = peerstore_opt or PeerStore()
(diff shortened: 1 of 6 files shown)


stuckinaboot:
Really great to see someone working on this! As this is a crucial functionality that depends on the whole py-libp2p networking stack being able to work with that of go-libp2p, I'm tagging @raulk in case he has any input here. With regards to your issue (Raul please correct me if I have this wrong or if there is some other place where sending peer IDs over the wire should take place), it could be useful to implement the [identify protocol](https://docs.rs/libp2p-identify/0.2.0/libp2p_identify/) in order to achieve the sharing of peer IDs when a connection is established.

raulk:
Hey! A few notes: peer IDs are exchanged along with public keys during the crypto handshake (SecIO, TLS 1.3, Noise in the future). Identify is a vital protocol, but by the time it kicks in, you already have a secure and multiplexed connection (and identify operates on a stream).

mhchia:
Rebased onto master but tests fail. It seems like incoming connections are not handled well.

mhchia:
In this PR, only `Identify` and `mplex` fixes are not merged in master. I will soon make the PR for `mplex` fix.

PR #180, opened 2019-07-09
Change `packages` in `setup`

mhchia:
### What is wrong
```bash
$ cd py-libp2p && rm -rf build/ && python3 setup.py sdist bdist_wheel
...
$ ls build/lib/libp2p/
__init__.py
```

Previously, only the top-level package `libp2p` is included. This means only `libp2p/__init__.py` is installed while `libp2p/host`, `libp2p/protocol_muxer`, ...etc are not when we try to install the whole package. This issue makes people unable to install this package correctly.

```bash
$ cd py-libp2p && pip install .
$ cd build/  # whatever directory, just not the package root, to test our installation.
$ python
>>> import libp2p
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mhchia/.pyenv/versions/py-libp2p/lib/python3.7/site-packages/libp2p/__init__.py", line 5, in <module>
    from libp2p.security.insecure_security import InsecureTransport
ModuleNotFoundError: No module named 'libp2p.security'
```

### How to fix it
Use `setuptools.find_packages` with `exclude` instead. `find_packages` includes all subpackages. All we need to do is to `exclude` the unwanted packages e.g. `tests`. This avoids listing all of the subpackages on our own.


mhchia:
cc @ralexstokes @stuckinaboot @zixuanzh

PR #182, opened 2019-07-15
Fix and refactor: `gossipsub.join`

NIC619:
### What was wrong?
Incorrect condition check in https://github.com/libp2p/py-libp2p/blob/b001256f5f184f6685c88a16a78ab199dff954c5/libp2p/pubsub/gossipsub.py#L206

### How was it fixed?
- Remove the line
- Also add an if-topic-already-in-mesh check to prevent re-joining a topic

#### Refactor `gossipsub.join`

This was intended for the fix but then I wonder if a refactor would improve the readability of this function. If not, I will revert the refactor and just apply the fix.

#### Also fixed
- [x] select gossipsub peers which did subscribe to the topic in mesh heartbeat

### DIFF
--- a/libp2p/pubsub/gossipsub.py
+++ b/libp2p/pubsub/gossipsub.py
@@ -265,26 +265,28 @@ class GossipSub(IPubsubRouter):
     async def mesh_heartbeat(self):
         # Note: the comments here are the exact pseudocode from the spec
         for topic in self.mesh:
+            # Skip if no peers have subscribed to the topic
+            if topic not in self.pubsub.peer_topics:
+                continue
 
             num_mesh_peers_in_topic = len(self.mesh[topic])
             if num_mesh_peers_in_topic < self.degree_low:
-                if topic in self.pubsub.peer_topics:
-                    gossipsub_peers_in_topic = [peer for peer in self.pubsub.peer_topics[topic]
-                                                if peer in self.peers_gossipsub]
+                gossipsub_peers_in_topic = [peer for peer in self.pubsub.peer_topics[topic]
+                                            if peer in self.peers_gossipsub]
 
-                    # Select D - |mesh[topic]| peers from peers.gossipsub[topic] - mesh[topic]
-                    selected_peers = GossipSub.select_from_minus(
-                        self.degree - num_mesh_peers_in_topic,
-                        gossipsub_peers_in_topic,
-                        self.mesh[topic]
-                    )
+                # Select D - |mesh[topic]| peers from peers.gossipsub[topic] - mesh[topic]
+                selected_peers = GossipSub.select_from_minus(
+                    self.degree - num_mesh_peers_in_topic,
+                    gossipsub_peers_in_topic,
+                    self.mesh[topic]
+                )
 
-                    for peer in selected_peers:
-                        # Add peer to mesh[topic]
-                        self.mesh[topic].append(peer)
+                for peer in selected_peers:
+                    # Add peer to mesh[topic]
+                    self.mesh[topic].append(peer)
 
-                        # Emit GRAFT(topic) control message to peer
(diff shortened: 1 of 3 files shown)


NIC619:
Test fix depends on #181 36575e8

libp2p/pubsub/gossipsub.py
```diff
@@ -184,43 +184,30 @@ def remove_peer(self, peer_id):
         # Create mesh[topic] if it does not yet exist
```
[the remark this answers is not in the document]
mhchia: Seems no need to remove it. Please ignore it.

tests/pubsub/test_gossipsub.py
```diff
@@ -2,37 +2,69 @@
...
+    msg_content = ""
+    host_id = str(libp2p_hosts[central_node_index].get_id())
+    # Generate message packet
+    packet = generate_RPC_packet(host_id, [topic], msg_content, next_msg_id_func())
+    # publish from the randomly chosen host
+    await gossipsubs[central_node_index].publish(host_id, packet.SerializeToString())
+
+    # Check that the gossipsub of central node has fanout for the topic
+    assert topic in gossipsubs[central_node_index].fanout
+    # Check that the gossipsub of central node does not has mesh for the topic
```
mhchia:
```suggestion
    # Check that the gossipsub of central node does not have a mesh for the topic
```

tests/pubsub/test_gossipsub.py
```diff
@@ -2,37 +2,69 @@
...
+    assert topic not in gossipsubs[central_node_index].mesh
+
+    # Central node subscribe message origin
```
mhchia:
```suggestion
    # Central node subscribes the topic
```

tests/pubsub/test_gossipsub.py
```diff
@@ -2,37 +2,69 @@
...
+    await pubsubs[central_node_index].subscribe(topic)
+
+    # Check that the gossipsub of central node no longer has fanout for the topic
+    assert topic not in gossipsubs[central_node_index].fanout
+
+    for i in hosts_indices:
+        if i in subscribed_peer_indices:
+            assert str(libp2p_hosts[i].get_id()) in gossipsubs[central_node_index].mesh[topic]
+        else:
+            assert str(libp2p_hosts[i].get_id()) not in gossipsubs[central_node_index].mesh[topic]
```
mhchia: I'm thinking if we should change the `Dict[str_of_peer_id, whatever]` to `Dict[PeerID, whatever` later. What do you think about it?

tests/pubsub/test_gossipsub.py
```diff
@@ -2,37 +2,69 @@
...
+
+    # Central node subscribe message origin
+    await pubsubs[central_node_index].subscribe(topic)
+
+    # Check that the gossipsub of central node no longer has fanout for the topic
+    assert topic not in gossipsubs[central_node_index].fanout
+
+    for i in hosts_indices:
+        if i in subscribed_peer_indices:
+            assert str(libp2p_hosts[i].get_id()) in gossipsubs[central_node_index].mesh[topic]
```
mhchia: Would you mind add a check?
```python
    assert str(libp2p_hosts[central_node_index].get_id()) in gossipsubs[i].mesh[topic]
```

tests/pubsub/test_gossipsub.py
mhchia: And also a check?
```python
    assert str(libp2p_hosts[central_node_index].get_id()) not in gossipsubs[i].mesh[topic]
```

NIC619: > I'm thinking if we should change the `Dict[str_of_peer_id, whatever]` to `Dict[PeerID, whatever` later. What do you think about it?

That makes sense to me.

PR #184, opened 2019-07-18
Fix `gossipsub.handle_graft`

NIC619:
### What was wrong?
According to the [spec](https://github.com/libp2p/specs/blob/7a3ad9096d7f1192d56547237faa983f1f2898eb/pubsub/gossipsub/README.md#message-processing), when receiving a `GRAFT` message which contains the topic that we didn't subscribe to, we should respond with a `PRUNE` message.

### How was it fixed?
Respond `GRAFT` with `PRUNE` if not subscribed to the topic.

### DIFF
--- a/tests/pubsub/test_gossipsub.py
+++ b/tests/pubsub/test_gossipsub.py
@@ -164,9 +164,9 @@ async def test_handle_prune():
     libp2p_hosts = await create_libp2p_hosts(num_hosts)
 
     # Create pubsub, gossipsub instances
-    _, gossipsubs = create_pubsub_and_gossipsub_instances(libp2p_hosts, \
-                                                          SUPPORTED_PROTOCOLS, \
-                                                          10, 9, 11, 30, 3, 5, 2)
+    pubsubs, gossipsubs = create_pubsub_and_gossipsub_instances(libp2p_hosts, \
+                                                                SUPPORTED_PROTOCOLS, \
+                                                                10, 9, 11, 30, 3, 5, 3)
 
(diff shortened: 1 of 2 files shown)


NIC619:
Test fix depends on #181 36575e8

NIC619:
Test fix depends on #182 613d5cc

tests/pubsub/test_gossipsub.py
```diff
@@ -98,6 +98,110 @@
...
+    id_bob = str(libp2p_hosts[index_bob].get_id())
+
+    topic = "test_handle_prune"
+    for pubsub in pubsubs:
+        await pubsub.subscribe(topic)
+
+    await connect(libp2p_hosts[index_alice], libp2p_hosts[index_bob])
+
+    # Wait 3 seconds for heartbeat to allow mesh to connect
+    await asyncio.sleep(3)
```
mhchia: Could you elaborate more on the reason why increasing the sleep time from `2` to `3`?

NIC619: The test would fail if heartbeat is triggered after alice emit prune message but before we check that alice is not bob's mesh peer because during heartbeat alice would be added back as bob's mesh peer. So I increase the `heartbeat_interval` to 3 seconds.
This test is currently not that reliable. I only increase the wait time hoping to not causing problems too often in CI.

mhchia: Sure. I'd open an issue to track the sleep time in the tests.

PR #192, opened 2019-07-28
Add type hint

NIC619:
Add type hints to `host`, `network`, `peer`, pubsub` and `routing` module.
Leave the type hints of protobuf messages to the next PR.

PR #195, opened 2019-07-30
Add typing 2

ChihChengLiang:
- Build on top of #192
- Working on Kad


ChihChengLiang:
Kad is not a priority now.

Issue #189, opened 2019-07-24
Change usage of type of peer id from string to `peer.ID`


NIC619:
PeerInfo [stores peer ID as type `peer.ID`](https://github.com/libp2p/py-libp2p/blob/master/libp2p/peer/peerinfo.py#L32-L34) while in other place peer ID is converted to string type, for example:
- https://github.com/libp2p/py-libp2p/blob/master/libp2p/pubsub/gossipsub.py#L117
- https://github.com/libp2p/py-libp2p/blob/master/libp2p/pubsub/gossipsub.py#L131

IMO this creates confusion and makes it hard to maintain. I suggest we use type `peer.ID` as the canonical way to represent peer ID.

ralexstokes:
did #196 address this fully or is there more here?

NIC619:
Yes!

PR #196, opened 2019-07-31
Refactor `peer.ID` class and only takes in `bytes` type argument

NIC619:
Address #189 
- [x] Only takes in `bytes` type argument during instantiation
- [x] Remove `get_raw_id()`(replace with `to_bytes()`
- [x] Rename `get_xor_id()` to `xor_id`(property)
- [x]  `__eq__` compares against either type `ID` or `bytes`
- [x] Rename `peer_id` to `peer_id_bytes` if its used as bytes
- [x] Fix `peer_id` used in pubsub: change it from type `str` to `bytes`

### DIFF
--- a/libp2p/kademlia/kad_peerinfo.py
+++ b/libp2p/kademlia/kad_peerinfo.py
@@ -16,7 +16,7 @@ class KadPeerInfo(PeerInfo):
     def __init__(self, peer_id, peer_data=None):
         super(KadPeerInfo, self).__init__(peer_id, peer_data)
 
-        self.peer_id = peer_id.to_bytes()
+        self.peer_id_bytes = peer_id.to_bytes()
         self.xor_id = peer_id.xor_id
 
         self.addrs = peer_data.get_addrs() if peer_data else None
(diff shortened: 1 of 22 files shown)


libp2p/kademlia/kad_peerinfo.py
```diff
@@ -16,9 +16,8 @@ class KadPeerInfo(PeerInfo):
     def __init__(self, peer_id, peer_data=None):
         super(KadPeerInfo, self).__init__(peer_id, peer_data)
 
-        self.peer_id_obj = peer_id
-        self.peer_id = peer_id.get_raw_id()
-        self.xor_id = peer_id.get_xor_id()
+        self.peer_id_bytes = peer_id.to_bytes()
```
mhchia: Will it be simpler to just store `ID`?

libp2p/peer/id.py
```diff
@@ -18,52 +18,78 @@
...
+        return hash(self._bytes)
+
+    @classmethod
+    def from_base58(cls, b58_encoded_peer_id_str: str) -> 'ID':
+        peer_id_bytes = base58.b58decode(b58_encoded_peer_id_str)
+        pid = ID(peer_id_bytes)
+        return pid
+
+    @classmethod
+    def from_pubkey(cls, key: RsaKey) -> 'ID':
```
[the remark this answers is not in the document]
NIC619: Agree on the idea of abstracting this layer. Will open a issue.

libp2p/kademlia/kad_peerinfo.py
NIC619: Only peer ID(in bytes) is used in kad so I think we can just use `peer_id_bytes`.

PR #198, opened 2019-07-31
Use factories and fixtures in pubsub tests

mhchia:
Done
- Add factories using factory-boy
- Modify fixtures and tests to use factories
- Modify tests to use fixtures and factories
- Clean up


mhchia:
This should be ready if py37-test passes. cc @ChihChengLiang @NIC619

PR #202, opened 2019-08-02
Add isort

mhchia:
### Done
- Add `.isort.cfg` according to the [suggestion](https://github.com/psf/black/blob/e66451761fab864cce9077a3ee72999543b8ef65/README.md#how-black-wraps-lines) in black.
- Check `isort` in `tox.ini`
- Run `isort --recursive .` through the code.

### Work flow suggestion
- Before committing changes, run `isort --recurive {py-libp2p-root-dir} && black {py-libp2p-root-dir}` under `{py-libp2p-root-dir}`.
- Configure IDE to perform the sorting and formatting.

For example, to make VS Code automatically do the sorting and formatting on save, you can configure `settings.json` as the following:
```
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
        "source.organizeImports": true
    }
```


mhchia:
Closed in favor of #206 . Since I experienced some issue when trying to use isort and black all together in VS code, so [did](https://github.com/timothycrosley/isort/issues/694#issuecomment-482052696) some other people, I think it will be better to use `black` only for now.

ralexstokes:
@mhchia we should try this: https://github.com/ethereum/lahja/blob/master/tox.ini#L9-L16

ralexstokes:
@mhchia  we should try this: https://github.com/ethereum/lahja/blob/master/tox.ini#L9-L16

mhchia:
Hmm, I didn't notice lahja also use isort with black. Thanks for the pointer. I was a bit worried since I experience some strange behavior on the when sorting and formatting together in VS Code. However, I couldn't reproduce it.

I will add it again and maybe we can use it for a while and see if it works well.

ralexstokes:
@mhchia put it together in this PR: https://github.com/libp2p/py-libp2p/pull/217

mhchia:
Closed in favor of #217

PR #203, opened 2019-08-02
add typing to transport and stream_muxer

ChihChengLiang:


libp2p/stream_muxer/mplex/mplex.py
```diff
@@ -1,25 +1,49 @@
...
+
+from .constants import HeaderTags
+from .utils import encode_uvarint, decode_uvarint_from_stream
 from .mplex_stream import MplexStream
-from ..muxed_connection_interface import IMuxedConn
+
+
+if TYPE_CHECKING:
+    from multiaddr import Multiaddr
+    from libp2p.security.secure_conn_interface import ISecureConn
```
pipermerriam: I've found that this circular import problem is fixed with a `libp2p.abc` module like [this from py-evm](https://github.com/ethereum/py-evm/blob/3b0bcc21954134e0293e40b1850e6bf16e5f4c27/eth/abc.py) allow you to have good interface enforced type hints that don't cause circular imports.

libp2p/transport/transport_interface.py
```diff
@@ -1,22 +1,29 @@
 from abc import ABC, abstractmethod
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from multiaddr import Multiaddr
+    from libp2p.peer.id import ID
+    from libp2p.network.connection.raw_connection_interface import IRawConnection
+    from .listener_interface import IListener
+    from .typing import THandler
```
pipermerriam: This doesn't seem like it should need to be under the `TYPE_CHECKING` context block as that module doesn't import from anywhere else in the library.

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -1,16 +1,25 @@
...
     """
     reference: https://github.com/libp2p/go-mplex/blob/master/stream.go
     """
 
-    def __init__(self, stream_id, initiator, mplex_conn):
+    stream_id: int
+    initiator: bool
+    mplex_conn: IMuxedConn
+    read_deadline: float
+    write_deadline: float
```
NIC619: I think it would be simpler to have ttl in seconds? @mhchia

libp2p/transport/tcp/tcp.py
```diff
@@ -1,68 +1,74 @@
...
+from libp2p.transport.transport_interface import ITransport
+from libp2p.transport.typing import THandler
 
-from ..listener_interface import IListener
-from ..transport_interface import ITransport
+
+class TCPListener(IListener):
+    multiaddrs: List[Multiaddr]
+    server = None
+    handler = None
```
NIC619: Should it be
```suggestion
    handler: THandler
```
?

libp2p/transport/tcp/tcp.py
```diff
@@ -81,18 +87,17 @@ def close(self, options=None):
         if ack != expected_ack_str:
             raise Exception("Receiver did not receive peer id")
 
-        return RawConnection(host, port, reader, writer, True)
+        return RawConnection(host, str(port), reader, writer, True)
```
NIC619: Maybe convert the port needs by `asyncio.open_connection(host, port)` instead?
```
port = maddr.value_for_protocol("tcp")
reader, writer = await asyncio.open_connection(host, int(port))
```

libp2p/transport/tcp/tcp.py
```diff
@@ -1,68 +1,74 @@
...
-            self.server.close()
-            _loop = asyncio.get_event_loop()
-            _loop.run_until_complete(self.server.wait_closed())
-            _loop.close()
-            self.server = None
-            return True
-
-    async def dial(self, maddr, self_id, options=None):
+    def __init__(self) -> None:
+        self.listener = TCPListener()
```
NIC619: I think we can delete this.

libp2p/stream_muxer/mplex/mplex.py
```diff
@@ -1,24 +1,46 @@
...
 class Mplex(IMuxedConn):
     """
     reference: https://github.com/libp2p/go-mplex/blob/master/multiplex.go
     """
 
-    def __init__(self, secured_conn, generic_protocol_handler, peer_id):
+    secured_conn: ISecureConn
+    raw_conn: IRawConnection
+    initiator: bool
+    generic_protocol_handler = None
```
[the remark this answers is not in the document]
ChihChengLiang: Per offline discussion, looks like assigning a method type still not available at this moment. https://github.com/python/mypy/issues/708

libp2p/transport/tcp/tcp.py
ChihChengLiang: same as https://github.com/libp2p/py-libp2p/pull/203#discussion_r310508651

PR #206, opened 2019-08-02
Run black over all .py

mhchia:
### Done
- Put `black` before `mypy` in tox.
- Include Python protobuf files.


ralexstokes:
@mhchia does the order of `black` vs `mypy` matter? i'd think it should _not_.

and i excluded the protobuf files from the black because they are autogenerated and we should be able to effectively ignore them

ralexstokes:
also heads up on #208 as I had to move the check to a Makefile to get globbing working.

mhchia:
@ralexstokes 
1. I agree that excluding pb file makes more sense. How about we add a config file `pyproject.toml` for black which specifies we should skip pb files? I think enumerating the directories might be a little bit troublesome if you need to add new directories, and also more possible to forget adding the new ones.
2. Essentially IMO style check(flake8) is usually faster and more trivial, compared to the type check. Also trinity runs flake8 before mypy(https://github.com/ethereum/trinity/blob/master/tox.ini#L137-L143).

ralexstokes:
great! i'd look at #215 for suggestions on how to move forward (which probably mean taking the `make` process out of `tox`)

mhchia:
Closed in favor of https://github.com/libp2p/py-libp2p/pull/220.

PR #207, opened 2019-08-02
Update chat example

ralexstokes:
Two changes to update the `chat` example.

1. We want to raise a `RuntimeError` as soon as we know there is a port missing so that check is hoisted earlier in the process logic.

2. Added a `localhost` option as I was getting some "Failed to connect" messages w/ the external IP (I think bc of my firewall settings)

### DIFF
--- a/examples/chat/chat.py
+++ b/examples/chat/chat.py
@@ -31,9 +31,12 @@ async def write_data(stream):
         await stream.write(line.encode())
 
 
-async def run(port, destination):
-    external_ip = urllib.request.urlopen("https://v4.ident.me/").read().decode("utf8")
-    transport_opt = "/ip4/%s/tcp/%s" % (external_ip, port)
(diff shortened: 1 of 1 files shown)


examples/chat/chat.py
```diff
@@ -95,11 +98,21 @@ def main():
...
+        "-l",
+        "--localhost",
+        dest="localhost",
+        action="store_true",
+        help="flag indicating if localhost should be used or an external IP",
+    )
     args = parser.parse_args()
 
+    if not args.port:
+        raise RuntimeError("was not able to determine a local port")
```
mhchia: I set `default` for `args.port`. In this case, is this branch possibly be entered?

ralexstokes: it would only run this branch if `args.port` was "falsey", but it definitely won't be w/ the default non-zero value

PR #212, opened 2019-08-02
Lay ground work for new types of secure transports

### LINKED ISSUE #210: Add `secio`
The repo currently lacks a `secio` implementation; we should add one :)

ralexstokes:
Ground work for adding in `secio` transport.

Next PR on top of this one will add `secio` implementation.

### DIFF
--- a/libp2p/__init__.py
+++ b/libp2p/__init__.py
@@ -34,9 +34,14 @@ async def cleanup_done_tasks() -> None:
         await asyncio.sleep(3)
 
 
-def generate_id() -> ID:
-    new_key = RSA.generate(2048, e=65537).publickey().export_key("DER")
-    new_id = ID.from_pubkey(new_key)
+def generate_new_private_key():
+    new_key = RSA.generate(2048, e=65537)
+    return new_key
+
+
+def generate_id(private_key):
+    public_key_bytes = private_key.publickey().export_key("DER")
+    new_id = ID.from_pubkey(public_key_bytes)
     return new_id
 
 
@@ -53,7 +58,7 @@ def initialize_default_kademlia_router(
     :return: return a default kademlia instance
     """
     if not id_opt:
-        id_opt = generate_id()
+        id_opt = generate_id(generate_new_private_key())
 
     node_id = id_opt.to_bytes()
     # ignore type for Kademlia module
@@ -64,6 +69,7 @@ def initialize_default_kademlia_router(
 
 
 def initialize_default_swarm(
+    private_key: bytes,
     id_opt: ID = None,
     transport_opt: Sequence[str] = None,
     muxer_opt: Sequence[str] = None,
@@ -83,7 +89,7 @@ def initialize_default_swarm(
     """
 
     if not id_opt:
-        id_opt = generate_id()
+        id_opt = generate_id(private_key)
 
     # TODO parse transport_opt to determine transport
     transport_opt = transport_opt or ["/ip4/127.0.0.1/tcp/8001"]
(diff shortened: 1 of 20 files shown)


libp2p/security/security_multistream.py
```diff
@@ -24,64 +21,48 @@
...
-    transports: Dict[TProtocol, "ISecureTransport"]
-    multiselect: "Multiselect"
-    multiselect_client: "MultiselectClient"
+    transports: Dict[TProtocol, ISecureTransport]
+    multiselect: Multiselect
+    multiselect_client: MultiselectClient
 
-    def __init__(self) -> None:
-        # Map protocol to secure transport
+    def __init__(self, secure_transports_by_protocol) -> None:
```
pipermerriam: missing type hints?  Any chance we can get things typed and then enforced via CI?

tests/utils.py
```diff
@@ -52,3 +54,8 @@
     # Associate the peer with local ip address (see default parameters of Libp2p())
     node_a.get_peerstore().add_addrs(node_b.get_id(), node_b.get_addrs(), 10)
     return node_a, node_b
+
+
+def generate_new_private_key():
+    new_key = RSA.generate(2048, e=65537)
+    return new_key
```
pipermerriam: looks identical to the `libp2p.generate_new_private_key` function.  Any reason it's duplicated?

libp2p/security/security_multistream.py
ralexstokes: yep, there is progress on restoring use of `mypy` in this repo -- i think it has already been completed so I'll have to fix this before we can merge this in.

tests/utils.py
ralexstokes: it is :)

"a little copy is better than a little dependency", esp when the code is short and you are moving fast ;)

i'll make a note to deprecate this (we need a better way to handle key gen anyway)

libp2p/__init__.py
```diff
@@ -93,8 +98,14 @@ def initialize_default_swarm(
     # TODO TransportUpgrader is not doing anything really
     # TODO parse muxer and sec to pass into TransportUpgrader
     muxer = muxer_opt or ["mplex/6.7.0"]
-    sec = sec_opt or {TProtocol("insecure/1.0.0"): InsecureTransport("insecure")}
-    upgrader = TransportUpgrader(sec, muxer)
+    private_key_bytes = private_key.export_key("DER")
```
mhchia: Just a note: I learned we lack `KeyBook` implementation, which is useful when we need to store peers' public keys in the future. Ref: https://github.com/libp2p/go-libp2p-peerstore/blob/master/pstoremem/keybook.go

tests/security/test_security_multistream.py
```diff
@@ -1,14 +1,13 @@
...
 from libp2p import new_node
 from libp2p.peer.peerinfo import info_from_p2p_addr
 from libp2p.protocol_muxer.multiselect_client import MultiselectClientError
-from libp2p.security.insecure_security import InsecureTransport
-from libp2p.security.simple_security import SimpleSecurityTransport
-from tests.utils import cleanup, connect
+from libp2p.security.insecure.transport import InsecureSession, InsecureTransport
+from libp2p.security.simple.transport import SimpleSecurityTransport
+import multiaddr
+import pytest
```
mhchia: Is this sorting(`libp2p` above `multiaddr` and `pytest`) desired by us? If not, we probably need to adjust the configuration of isort a little bit.

ralexstokes:
just a note, this was closed after a rename operation i tried locally... will make new PR

tests/security/test_security_multistream.py
ralexstokes: yeah i noticed this -- not sure what is going on here

ralexstokes: this seems to have magically fixed itself after i rebased master in #250.

libp2p/__init__.py
ralexstokes: yeah let's make an issue

libp2p/security/security_multistream.py
ralexstokes: this has been addressed in #250.

tests/utils.py
ralexstokes: also fixed in other work, reflected in #250.

PR #214, opened 2019-08-02
Refactor ID to not use third-party type for cryptographic keys

ralexstokes:
Fixes 201.

The thing we want to do here is not use a third-party dependency as a canonical type.

I left it as `bytes` for now as that is the universal format for any private/public key we could use to derive an `ID`.

The Go impl defines their own cryptographic interfaces which we could also do. I'll leave it for further discussion.

### DIFF
--- a/libp2p/peer/id.py
+++ b/libp2p/peer/id.py
@@ -5,16 +5,6 @@ import base58
 
 import multihash
 
-from Crypto.PublicKey.RSA import RsaKey
-
-# MaxInlineKeyLength is the maximum length a key can be for it to be inlined in
-# the peer ID.
-# * When `len(pubKey.Bytes()) <= MaxInlineKeyLength`, the peer ID is the
-#   identity multihash hash of the public key.
-# * When `len(pubKey.Bytes()) > MaxInlineKeyLength`, the peer ID is the
-#   sha2-256 multihash of the public key.
-MAX_INLINE_KEY_LENGTH = 42
-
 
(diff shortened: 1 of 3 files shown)


libp2p/peer/id.py
```diff
@@ -64,21 +54,14 @@ def from_base58(cls, b58_encoded_peer_id_str: str) -> "ID":
...
-        mh_digest: multihash.Multihash = multihash.digest(key_bin, algo)
+    def from_pubkey(cls, key: bytes) -> "ID":
+        algo = multihash.Func.sha2_256
+        mh_digest = multihash.digest(key, algo)
         return cls(mh_digest.encode())
 
     @classmethod
-    def from_privkey(cls, key: RsaKey) -> "ID":
-        return cls.from_pubkey(key.publickey())
+    def from_privkey(cls, public_key: bytes) -> "ID":
```
NIC619: This should be `private_key`. But then it can't be type `bytes`...🤔

ralexstokes: yeah -- i actually had more to add here from another branch!

i ended up just dropping the `from_privkey` method, i don't see how we can keep it generic to a particular cryptosystem while taking private key bytes (bc it will be specific to the algo to get the public key)

ralexstokes:
had a few more commits to add after working on the #212, i'll push soon

ralexstokes:
ok -- this should be ready to merge in -- review away! :)

PR #224, opened 2019-08-03
Remove `make` from CI linting

### LINKED ISSUE #215: Do not require `make` during linting
Try something like this: https://stackoverflow.com/questions/25818558/use-wildcard-in-tox-command/25823102#25823102

instead of the current method proxying out to `make` to run the linting during CI in `tox`.

ralexstokes:
Fixes #215.


ralexstokes:
CI failure here is the code coverage metric which I don't think blocks this from being merged in.

PR #225, opened 2019-08-03
Update flake8 ignore rules

ralexstokes:
Removes some rules from the ignore list. We take the option to opt-out of the rule via the special case `# noqa: $RULE` comment syntax in the places where we want to obvious opt-out but otherwise have the rule in place.

The main example here is for line lengths, where we generally want to disallow lines over a certain length.

### DIFF
--- a/examples/chat/chat.py
+++ b/examples/chat/chat.py
@@ -53,9 +53,10 @@ async def run(port, destination, localhost):
         print(
             f"Run 'python ./examples/chat/chat.py"
             + localhost_opt
-            + f" -p {int(port) + 1} -d /ip4/{ip}/tcp/{port}/p2p/{host.get_id().pretty()}' on another console.\n"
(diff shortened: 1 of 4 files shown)


examples/chat/chat.py
```diff
@@ -53,9 +53,10 @@
         print(
             f"Run 'python ./examples/chat/chat.py"
             + localhost_opt
-            + f" -p {int(port) + 1} -d /ip4/{ip}/tcp/{port}/p2p/{host.get_id().pretty()}' on another console.\n"
+            + f" -p {int(port) + 1} -d /ip4/{ip}/tcp/{port}/p2p/{host.get_id().pretty()}'"
+            + " on another console."
         )
-        print("\nWaiting for incoming connection\n\n")
+        print("Waiting for incoming connection...")
```
ralexstokes: in this file, we fix the long line that was hiding, and also apply a formatting fix to remove so much whitespace

Issue #194, opened 2019-07-30
Change return `True`/`False` pattern to `try/catch` pattern


NIC619:
Change pattern like this(L59): https://github.com/libp2p/py-libp2p/blob/f0046fa3e0952b492837a698b1988e05c9821f47/libp2p/host/basic_host.py#L54-L59

to try/ catch pattern

ralexstokes:
what is the rationale here?

NIC619:
Just align with the error handling pattern of python IMO?

ralexstokes:
gotcha -- yeah pythonistas do like their exceptions :)

my own take on it is that exceptions don't compose very well which leads to several symptoms in code like non-local effects/interactions that make code harder to reason about. i also cant take library code that may throw an exception and put it into other arbitrary code w/o also considering how the exception will affect my new code -- again, higher mental load/context. if i only have to consider (and handle!) a single boolean right at the call site, it tends to keep things local, even possibly at the expense of more verbose code (altho i think in this case the verbosity is better than the non-local effects)

while we are here, we could also have a conversation about the implicit mutability of this API but that is not what this issue is about so i'll continue on :)

i'm not super against using exceptions where appropriate but i'd think we need to consider on a case-by-case basis, rather than a blanket policy

NIC619:
Agree on the mental load and verbosity/non-local effects trade-off.

These are the callees that returns true/false:
- `Network.set_stream_handler()`
- `Network.listen()`
- `Network.notify()`
- `NetStream.set_protocol()`
- `MplexStream.close()`
- `MplexStream.reset()`
- `MplexStream.set_*_deadline()`

IMO only `Network.listen()`, `MplexStream.close()` and `MplexStream.reset()` need to have a return value(or throwing exceptions).
- For `Network.listen()`
https://github.com/libp2p/py-libp2p/blob/21e013e7537d4cad66a5d73c21e05d54a40ba33e/libp2p/network/network_interface.py#L61-L65
    - Since there are possibly more than one multiaddr to listen on and only one of them we need to listen on successfully, I'm +1 on return true/false here.
- For `MplexStream.close()` and `MplexStream.reset()`
    - Not sure about this. What do we do when closing/resetting a stream failed? It seems to me more verbosity is needed in this case but probably have to figure out what could go wrong when closing/resetting a stream first.

mhchia:
Is this solved yet?

ralexstokes:
[...]

Issue #193, opened 2019-07-30
Pubsub: Tasks are spawned but no way to stop them


mhchia:
Several tasks are spawned in the background in Pubsub.

1. `handle_peer_queue` in `__init__`
https://github.com/libp2p/py-libp2p/blob/f0046fa3e0952b492837a698b1988e05c9821f47/libp2p/pubsub/pubsub.py#L110

2. `continuously_read_stream` in `stream_handler`, every time a new stream is opened from a peer.
https://github.com/libp2p/py-libp2p/blob/f0046fa3e0952b492837a698b1988e05c9821f47/libp2p/pubsub/pubsub.py#L180

3. `continuously_read_stream ` in `handle_peer_queue`, every time a new peer is connected.
https://github.com/libp2p/py-libp2p/blob/f0046fa3e0952b492837a698b1988e05c9821f47/libp2p/pubsub/pubsub.py#L208

Regardless of how the code will change after https://github.com/libp2p/py-libp2p/issues/191 is fixed, we should maintain all spawned tasks, and close them when `Pubsub` is closed or even being garbage-collected.

ralexstokes:
just to update -- ran into this while running the tests

i'd have to dig in deeper to the `asyncio` machinery to see what is going on here -- sounds like we should finish the feature set and then see what we can do here

NIC619:
Related: https://github.com/libp2p/py-libp2p/pull/226#discussion_r310675182

There are a few places where tasks are also untracked: `kademlia/`, `network/swarm.py` and `mplex/mplex.py`.

ralexstokes:
yeah, we should generally be tracking all of the tasks we make so that we can handle cancellation/shutdown properly

mhchia:
closed by #309

PR #119, opened 2019-02-02
[WIP] Logging

AliabbasMerchant:
I  have started logging using `structlog`, as discussed in #84 
Open to suggestions, advice, help and guidance 😃

PR #226, opened 2019-08-04
Add topic message validator

NIC619:
- [x] Add `pubsub.topic_validators`
- [x] Add `set_topic_validator` and `remove_topic_validator`
- [x] Add `get_msg_validators`
- [x] Add `validate_msg`
- [x] Add `signature_validator` which is just a stub
- [x] Validate signture and message in `push_msg`
- [x] `ensure_future(push_msg())` instead of `await push_msg()`?

### DIFF
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -170,6 +170,13 @@ class Pubsub:
         if topic in self.topic_validators:
             del self.topic_validators[topic]
 
+    def get_msg_validators(self, msg: rpc_pb2.Message) -> Tuple[TopicValidator, ...]:
+        return (
+            self.topic_validators[topic]
+            for topic in msg.topicIDs
+            if topic in self.topic_validators
+        )
+
     async def stream_handler(self, stream: INetStream) -> None:
         """
         Stream handler for pubsub. Gets invoked whenever a new stream is created
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -1,17 +1,7 @@
 import asyncio
 from collections import namedtuple
 import time
-from typing import (
-    Any,
-    Awaitable,
-    Callable,
-    Dict,
-    Iterable,
-    List,
-    Tuple,
-    Union,
-    TYPE_CHECKING,
-)
+from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, Iterable, List, Tuple, Union
 
 from lru import LRU
 
@@ -381,9 +371,7 @@ class Pubsub:
         async_topic_validator_futures = []
         for topic_validator in self.get_msg_validators(msg):
             if topic_validator.is_async:
-                async_topic_validator_futures.append(
-                    topic_validator.validator(msg_forwarder, msg)
-                )
+                async_topic_validator_futures.append(topic_validator.validator(msg_forwarder, msg))
             else:
                 sync_topic_validators.append(topic_validator.validator)
 
@@ -448,4 +436,4 @@ class Pubsub:
     def _is_subscribed_to_msg(self, msg: rpc_pb2.Message) -> bool:
         if not self.my_topics:
             return False
-        return all([topic in self.my_topics for topic in msg.topicIDs])
+        return any([topic in self.my_topics for topic in msg.topicIDs])
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -151,7 +151,7 @@ class Pubsub:
                         continue
                     # TODO(mhchia): This will block this read_stream loop until all data are pushed.
                     #   Should investigate further if this is an issue.
-                    await self.push_msg(msg_forwarder=peer_id, msg=msg)
+                    asyncio.ensure_future(self.push_msg(msg_forwarder=peer_id, msg=msg))
 
             if rpc_incoming.subscriptions:
                 # deal with RPC.subscriptions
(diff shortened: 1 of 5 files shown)


libp2p/pubsub/pubsub.py
```diff
@@ -20,6 +22,12 @@ def get_msg_id(msg: rpc_pb2.Message) -> Tuple[bytes, bytes]:
     return (msg.seqno, msg.from_id)
 
 
+TopicValidator = namedtuple("TopicValidator", ["validator", "is_async"])
```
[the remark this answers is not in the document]
pipermerriam: Yes, `typing.NamedTuple` is preferable.

libp2p/pubsub/pubsub.py
```diff
@@ -320,6 +361,32 @@ def handle_subscription(self, origin_id: ID, sub_message: Any) -> None:
...
+        async_topic_validator_futures = []
+        for topic_validator in self.get_msg_validators(msg):
+            if topic_validator.is_async:
+                async_topic_validator_futures.append(topic_validator.validator(msg_forwarder, msg))
+            else:
+                sync_topic_validators.append(topic_validator.validator)
+
+        for validator in sync_topic_validators:
+            if not validator(msg_forwarder, msg):
+                return False
```
pipermerriam: I'd recommend against a `True/False` based validation scheme in favor of an excepiton based on.  This diverges from the standard pattern we have in place for validation across most of the rest of our libraries **and** it requires return-value-checking.

If you want to stay with the `True/False` return scheme, I'd recommend a name change to something like `is_msg_valid(msg)`

libp2p/pubsub/pubsub.py
```diff
@@ -336,6 +403,14 @@ def handle_subscription(self, origin_id: ID, sub_message: Any) -> None:
             return
 
         # TODO: - Validate the message. If failed, reject it.
+        # Validate the signature of the message
+        # FIXME: `signature_validator` is currently a stub.
+        if not signature_validator(msg.key, msg.SerializeToString()):
+            return
```
pipermerriam: This early exit seems like a prime place for a logging message.

libp2p/pubsub/pubsub.py
```diff
@@ -336,6 +403,14 @@ def handle_subscription(self, origin_id: ID, sub_message: Any) -> None:
...
+        # Validate the message with registered topic validators
+        is_validation_passed = await self.validate_msg(msg_forwarder, msg)
+        if not is_validation_passed:
+            return
```
pipermerriam: This early return is a little confusing to me.  Reading the code (briefly) doesn't make it clear why we would return early on validation passing...  Maybe expand the code comment to try and explain this?

libp2p/pubsub/pubsub.py
```diff
@@ -361,4 +436,4 @@ def _mark_msg_seen(self, msg: rpc_pb2.Message) -> None:
     def _is_subscribed_to_msg(self, msg: rpc_pb2.Message) -> bool:
         if not self.my_topics:
             return False
-        return all([topic in self.my_topics for topic in msg.topicIDs])
+        return any([topic in self.my_topics for topic in msg.topicIDs])
```
pipermerriam: If you change the inner comprehension to use `()` then it will be a generator allowing `any` to return early on the first match.

libp2p/pubsub/pubsub.py
NIC619: There was #119 that implements logging. Besides the standard logging, he also implement some of his ideas there. Wonder if you have time to take a look. @pipermerriam

pipermerriam: I'd say I lean :-1: on the fancy stuff in #119 as it tries to automate something that I don't think automates well.  Good logging messages are "human" things, at least in this context where we're trying to understand the sequence of events with minimal noise.

Similarly there is performance overhead to logging so an approach that does something like drop a decorator on every function and log all the stuff it does is likely to have a measurable effect on speed.

I'd be much more supportive of just giving the various classes their own loggers and putting some logging messages in the places that feel important/significant and then we can expand from there.

I don't think there's a direct need for `DEBUG2` level logs or anything like that at this stage and it's very easy to add in later once the need is clear.

libp2p/pubsub/pubsub.py
NIC619: There's a similar discussion in #194 but that one is more about user facing APIs. Would also love your feedback on that.

libp2p/pubsub/pubsub.py
```diff
@@ -128,7 +141,7 @@ def get_hello_packet(self) -> bytes:
                         continue
                     # TODO(mhchia): This will block this read_stream loop until all data are pushed.
                     #   Should investigate further if this is an issue.
-                    await self.push_msg(msg_forwarder=peer_id, msg=msg)
+                    asyncio.ensure_future(self.push_msg(msg_forwarder=peer_id, msg=msg))
```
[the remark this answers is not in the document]
NIC619: An issue(#193) was opened to track this but we should expand the coverage to whole repo instead of just pubsub/. Currently there are also a few such untracked tasks in kademlia/, network/ and stream_muxer/.

tests/pubsub/test_pubsub.py
```diff
@@ -84,6 +85,148 @@ def _get_hello_packet_topic_ids():
...
+        seqno=b"\x00" * 8,
+    )
+
+    if is_topic_1_val_passed and is_topic_2_val_passed:
+        await pubsubs_fsub[0].validate_msg(pubsubs_fsub[0].my_id, msg)
+    else:
+        with pytest.raises(ValidationError):
+            await pubsubs_fsub[0].validate_msg(pubsubs_fsub[0].my_id, msg)
+
+
```
mhchia: Could you add a test/check, either a new one or to the existing one, that `push_msg` fails to push the messages if validation fails? Will really appreciate that.

PR #230, opened 2019-08-07
Modify `NetStream` to read n bytes

mhchia:
### TODO
- [ ] Add tests
- [x] Fix the rest of the code due to the API change of `MplexStream.read`. I modify the implementation according to the Go implementation, which enforces `io.Reader.Read`. I don't have a proper way to support the API `read(n = -1)` as `asyncio.StreamReader`(blocking read until EOF received), without EOF detection supported.
    - References
        - https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.read
        - https://golang.org/pkg/io/#Reader

### DIFF
--- a/libp2p/stream_muxer/mplex/mplex.py
+++ b/libp2p/stream_muxer/mplex/mplex.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Dict, Tuple, Optional
+from typing import Dict, Optional, Tuple
 
 from multiaddr import Multiaddr
 
--- a/libp2p/stream_muxer/mplex/mplex.py
+++ b/libp2p/stream_muxer/mplex/mplex.py
@@ -75,7 +75,8 @@ class Mplex(IMuxedConn):
 
     async def read_buffer(self, stream_id: int) -> bytes:
         """
-        Read a message from stream_id's buffer, check raw connection for new messages
+        Read a message from stream_id's buffer, check raw connection for new messages.
+        `StreamNotFound` is raised when stream `stream_id` is not found in `Mplex`.
         :param stream_id: stream id of stream to read from
         :return: message read
         """
@@ -86,6 +87,7 @@ class Mplex(IMuxedConn):
     async def read_buffer_nonblocking(self, stream_id: int) -> Optional[bytes]:
         """
         Read a message from `stream_id`'s buffer, non-blockingly.
+        `StreamNotFound` is raised when stream `stream_id` is not found in `Mplex`.
         """
         if stream_id not in self.buffers:
             raise StreamNotFound(f"stream {stream_id} is not found")
--- a/libp2p/stream_muxer/mplex/mplex.py
+++ b/libp2p/stream_muxer/mplex/mplex.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Dict, Tuple
+from typing import Dict, Tuple, Optional
 
 from multiaddr import Multiaddr
 
@@ -10,6 +10,7 @@ from libp2p.security.secure_conn_interface import ISecureConn
 from libp2p.stream_muxer.abc import IMuxedConn, IMuxedStream
 
 from .constants import HeaderTags
+from .exceptions import StreamNotFound
 from .mplex_stream import MplexStream
 from .utils import decode_uvarint_from_stream, encode_uvarint
 
@@ -23,6 +24,9 @@ class Mplex(IMuxedConn):
     raw_conn: IRawConnection
     initiator: bool
     peer_id: ID
+    # TODO: `dataIn` in go implementation. Should be size of 8.
+    # TODO: Also, `dataIn` is closed indicating EOF in Go. We don't have similar strategies
+    #   to let the `MplexStream`s know that EOF arrived (#235).
     buffers: Dict[int, "asyncio.Queue[bytes]"]
     stream_queue: "asyncio.Queue[int]"
 
@@ -75,17 +79,19 @@ class Mplex(IMuxedConn):
         :param stream_id: stream id of stream to read from
         :return: message read
         """
-        # TODO: propagate up timeout exception and catch
-        # TODO: pass down timeout from user and use that
-        if stream_id in self.buffers:
-            try:
-                data = await asyncio.wait_for(self.buffers[stream_id].get(), timeout=8)
-                return data
-            except asyncio.TimeoutError:
-                return None
+        if stream_id not in self.buffers:
+            raise StreamNotFound(f"stream {stream_id} is not found")
+        return await self.buffers[stream_id].get()
 
-        # Stream not created yet
-        return None
+    async def read_buffer_nonblocking(self, stream_id: int) -> Optional[bytes]:
+        """
+        Read a message from `stream_id`'s buffer, non-blockingly.
+        """
+        if stream_id not in self.buffers:
+            raise StreamNotFound(f"stream {stream_id} is not found")
+        if self.buffers[stream_id].empty():
+            return None
+        return await self.buffers[stream_id].get()
 
     async def open_stream(
         self, protocol_id: str, multi_addr: Multiaddr
@@ -170,6 +176,7 @@ class Mplex(IMuxedConn):
         :return: stream_id, flag, message contents
         """
 
+        # FIXME: No timeout is used in Go implementation.
         # Timeout is set to a relatively small value to alleviate wait time to exit
         #  loop in handle_incoming
         timeout = 0.1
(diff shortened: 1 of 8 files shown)


libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -34,13 +37,33 @@ def __init__(self, stream_id: int, initiator: bool, mplex_conn: IMuxedConn) -> N
...
+        :param n: number of bytes to read
         :return: bytes of input
         """
-        return await self.mplex_conn.read_buffer(self.stream_id)
+        if n == -1:
+            return await self.mplex_conn.read_buffer(self.stream_id)
+        return await self.read_bytes(n)
+
+    async def read_bytes(self, n: int) -> bytes:
+        if self._buf is None:
```
pipermerriam: This function could use some code comments because it isn't clear what we are doing with this other buffer and why.

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -34,13 +37,33 @@ def __init__(self, stream_id: int, initiator: bool, mplex_conn: IMuxedConn) -> N
...
 
-    async def read(self) -> bytes:
+    async def read(self, n: int = -1) -> bytes:
         """
         read messages associated with stream from buffer til end of file
+        :param n: number of bytes to read
         :return: bytes of input
         """
-        return await self.mplex_conn.read_buffer(self.stream_id)
+        if n == -1:
```
NIC619: Should we check `if self._buf is None` first before reading from `mplex_conn`?

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -34,13 +37,33 @@ def __init__(self, stream_id: int, initiator: bool, mplex_conn: IMuxedConn) -> N
...
+            self._buf = await self.mplex_conn.read_buffer(self.stream_id)
+        n_read = 0
+        bytes_buf = BytesIO()
+        while self._buf is not None and n_read < n:
+            n_to_read = min(n - n_read, len(self._buf))
+            bytes_buf.write(self._buf[:n_to_read])
+            if n_to_read == n - n_read:
+                self._buf = self._buf[n_to_read:]
+            else:
+                self._buf = None
```
NIC619: Seems no need to set it to `None` before `mplex_conn.read_buffer` as `mplex_conn.read_buffer` would return `None` if nothing is read in time.

libp2p/stream_muxer/mplex/mplex_stream.py
mhchia: Sorry that I didn't clarify why I add this function. I intend to align `Mplx.read` with the one in [Go](https://github.com/libp2p/go-mplex/blob/62fe9554facaec3f80333b61ea8d694fe615705f/stream.go#L112-L140). This is because that the current API of `Stream` make us unable to read the exact number of bytes, which is necessary when we are reading `varint`s. `read_bytes` is a temporary workaround without breaking the existing `read()` behavior. I will replace `read_bytes` with `read` and modify the necessary places before changing this PR from the draft to open.

libp2p/stream_muxer/mplex/mplex_stream.py
mhchia: Yes, you're right. Good catch. I will actually replace `read` with `read_bytes` later.

libp2p/stream_muxer/mplex/mplex_stream.py
pipermerriam: Take a look at this alternate implementation and let me know what you think about the two with respect to readability.

```python
    def read_exactly(self, num_bytes: int) -> bytes:
        while len(self._buffer) < num_bytes:

            data = self._socket.recv(4096)

            if data == b"":
                raise OSError("Connection closed")

            self._buffer.extend(data)
        payload = self._buffer[:num_bytes]
        self._buffer = self._buffer[num_bytes:]
        return bytes(payload)
```

source: https://github.com/ethereum/trinity/blob/40c2b0425a612ae7d2fea184a5f1f85fd6d9b8bc/trinity/db/manager.py#L36-L47

mhchia: From the readability perspective, I think `read_exactly` is far better. I will change to this pattern. Thanks for the direction.

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -34,13 +36,42 @@ def __init__(self, stream_id: int, initiator: bool, mplex_conn: IMuxedConn) -> N
...
+        :param n: number of bytes to read
+        :return: bytes actually read
+        """
+        if n < 0 and n != -1:
+            raise ValueError("`n` can only be -1 if it is negative")
+        # If the buffer is empty at first, blocking wait for data.
+        if len(self._buf) == 0:
+            self._buf.extend(await self.mplex_conn.read_buffer(self.stream_id))
+        # Sanity check: `self._buf` should never be empty here.
+        if self._buf is None or len(self._buf) == 0:
```
NIC619: I don't quite get these two `if` here:
```python
                # If the buffer is empty at first, blocking wait for data.
	        if len(self._buf) == 0:
	            self._buf.extend(await self.mplex_conn.read_buffer(self.stream_id))
	        # Sanity check: `self._buf` should never be empty here.
	        if self._buf is None or len(self._buf) == 0:
                    raise Exception("`self._buf` should never be empty here")
```

`len(self._buf) == 0` was checked in both `if` and `self._buf is None` comes in second `if`?

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -34,13 +36,42 @@ def __init__(self, stream_id: int, initiator: bool, mplex_conn: IMuxedConn) -> N
...
+    async def read(self, n: int = -1) -> bytes:
+        """
+        Read up to n bytes. Read possibly returns fewer than `n` bytes,
+        if there are not enough bytes in the Mplex buffer.
+        If `n == -1`, read until EOF.
+        :param n: number of bytes to read
+        :return: bytes actually read
+        """
+        if n < 0 and n != -1:
+            raise ValueError("`n` can only be -1 if it is negative")
```
NIC619:
```suggestion
            raise ValueError("n can not be negative but -1, indicating read until EOF, but got n={n}")
```

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -34,13 +36,42 @@ def __init__(self, stream_id: int, initiator: bool, mplex_conn: IMuxedConn) -> N
...
+        # If the buffer is empty at first, blocking wait for data.
+        if len(self._buf) == 0:
+            self._buf.extend(await self.mplex_conn.read_buffer(self.stream_id))
+        # Sanity check: `self._buf` should never be empty here.
+        if self._buf is None or len(self._buf) == 0:
+            raise Exception("`self._buf` should never be empty here")
```
NIC619: This error doesn't seem to be handled by the functions in the call stack and it should probably be noted in the docstring.

libp2p/stream_muxer/mplex/mplex_stream.py
ralexstokes:
```suggestion
            raise ValueError(f"the number of bytes to read ``n`` must be positive or -1 to indicate read until EOF")
```

libp2p/stream_muxer/mplex/mplex_stream.py
ralexstokes: yeah -- it seems like `self._buf` will never be `None` (ideally enforced via `mypy`) and i'm not sure we need the sanity check

mhchia: > I don't quite get these two if here:

Good point.

> yeah -- it seems like self._buf will never be None (ideally enforced via mypy) and i'm not sure we need the sanity check

You're right. Will remove the check.

libp2p/stream_muxer/mplex/mplex.py
```diff
@@ -78,17 +82,19 @@ def is_closed(self) -> bool:
...
+        return await self.buffers[stream_id].get()
 
-        # Stream not created yet
-        return None
+    async def read_buffer_nonblocking(self, stream_id: int) -> Optional[bytes]:
+        """
+        Read a message from `stream_id`'s buffer, non-blockingly.
+        """
+        if stream_id not in self.buffers:
+            raise StreamNotFound(f"stream {stream_id} is not found")
```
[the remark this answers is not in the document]
mhchia: Good catch. Added.

PR #240, opened 2019-08-14
Allow multiple peer identity types (via different cryptosystems)

### LINKED ISSUE #222: `Peer.ID` implementation only offers `RSA` right now
The peer identity spec supports several different cryptosystems for identity generation; we should at least make it straightforward to use `secp256k1` keys.

ralexstokes:
Fixes #222. 

Introduces the option for multiple cryptosystems (following the peer ID spec).

Also generates peer IDs according to the spec (the prior implementation was missing some serialization steps).

## TODO
- [x] Add conformance test to the Go implementation based on this: https://github.com/libp2p/go-libp2p-core/blob/master/peer/peer_test.go#L122


libp2p/crypto/secp256k1.py
```diff
@@ -0,0 +1,52 @@
...
+        raise NotImplementedError
+
+
+class Secp256k1PrivateKey(PrivateKey):
+    def __init__(self, impl: coincurve.PrivateKey) -> None:
+        self.impl = impl
+
+    @classmethod
+    def new(cls, secret: bytes = None) -> "Secp256k1PrivateKey":
+        private_key_impl = coincurve.PrivateKey()
```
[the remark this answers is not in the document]
ralexstokes: nice catch :)

libp2p/crypto/keys.py
```diff
@@ -0,0 +1,82 @@
...
+
+@unique
+class KeyType(Enum):
+    RSA = 0
+    Ed25519 = 1
+    Secp256k1 = 2
+    ECDSA = 3
+
+
+class Key:
```
pipermerriam: missing `ABC` as base class.

tests/peer/test_peerid.py
```diff
@@ -59,7 +53,7 @@ def test_str_more_than_10():
 def test_eq_true():
     random_id_string = ""
     for _ in range(10):
-        random_id_string += random.SystemRandom().choice(ALPHABETS)
+        random_id_string += random.choice(ALPHABETS)
```
pipermerriam: probably slightly better to use `secrets.choice` here.

libp2p/crypto/keys.py
ralexstokes: good catch

tests/peer/test_peerid.py
ralexstokes: i don't see the rationale... we just want to sample some characters from the alphabet to build a random string... i don't see the need for "stronger" randomness in this context

Issue #239, opened 2019-08-13
Ensure protocol negotiating process is correct


mhchia:
In Go, the process of negotiation for connection upgrading is:
1. Negotiate multiselect protocol id.
2. Negotiate security level.
3. Negotiate multiselect protocol id.
4. Negotiate multiplexer.

In Python, some steps are wrong or missing. I've fixed the `TransportUpgrader` and made it work with Go in https://github.com/libp2p/py-libp2p/pull/175/files#diff-6c722cdc8acbbb2dec74dc5e7e04ac6d.

### To-Do
- [ ] Add the fix in https://github.com/libp2p/py-libp2p/pull/175 for both side(Python to Go and Go to Python).
- [ ] Add tests.

ralexstokes:
Related: #242 .

ralexstokes:
Related: #243.

PR #252, opened 2019-08-16
Negotiate multiplexer protocol when upgrading to `MuxedConn`

mhchia:
- Depends on #248 .
- Fixes https://github.com/libp2p/py-libp2p/issues/239.
- Extracted from #175 .

### DIFF
--- a/libp2p/__init__.py
+++ b/libp2p/__init__.py
@@ -13,9 +13,9 @@ from libp2p.peer.peerstore import PeerStore
 from libp2p.peer.peerstore_interface import IPeerStore
 from libp2p.routing.interfaces import IPeerRouting
 from libp2p.routing.kademlia.kademlia_peer_router import KadmeliaPeerRouter
-from libp2p.security.insecure.transport import InsecureTransport
+from libp2p.security.insecure.transport import PLAINTEXT_PROTOCOL_ID, InsecureTransport
 from libp2p.security.secure_transport_interface import ISecureTransport
-from libp2p.stream_muxer.mplex.mplex import Mplex
+from libp2p.stream_muxer.mplex.mplex import MPLEX_PROTOCOL_ID, Mplex
 from libp2p.stream_muxer.muxer_multistream import MuxerClassType
 from libp2p.transport.tcp.tcp import TCP
 from libp2p.transport.upgrader import TransportUpgrader
@@ -96,10 +96,9 @@ def initialize_default_swarm(
     # TODO: Parse `transport_opt` to determine transport
     transport = TCP()
 
-    # TODO TransportUpgrader is not doing anything really
-    muxer_transports_by_protocol = muxer_opt or {TProtocol("/mplex/6.7.0"): Mplex}
+    muxer_transports_by_protocol = muxer_opt or {MPLEX_PROTOCOL_ID: Mplex}
     security_transports_by_protocol = sec_opt or {
-        TProtocol("/plaintext/1.0.0"): InsecureTransport(key_pair)
+        PLAINTEXT_PROTOCOL_ID: InsecureTransport(key_pair)
     }
     upgrader = TransportUpgrader(
         security_transports_by_protocol, muxer_transports_by_protocol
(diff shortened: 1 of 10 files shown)


libp2p/protocol_muxer/multiselect_communicator.py
```diff
@@ -1,35 +1,47 @@
...
+
+class RawConnectionCommunicator(IMultiselectCommunicator):
+    conn: IRawConnection
+
+    def __init__(self, conn: IRawConnection) -> None:
+        self.conn = conn
+
+    async def write(self, msg_str: str) -> None:
+        msg_bytes = delim_encode(msg_str)
+        self.conn.writer.write(msg_bytes)
```
NIC619: Should this be awaited?

libp2p/transport/upgrader.py
```diff
@@ -42,14 +44,15 @@ def upgrade_listener(self, transport: ITransport, listeners: IListener) -> None:
...
+    async def upgrade_connection(
+        self,
         conn: ISecureConn,
         generic_protocol_handler: GenericProtocolHandlerFn,
         peer_id: ID,
-    ) -> Mplex:
+    ) -> IMuxedConn:
         """
-        Upgrade raw connection to muxed connection
+        Upgrade secured connection to be a muxed connection
```
NIC619: nit:
```suggestion
        Upgrade secured connection to a muxed connection
```

libp2p/protocol_muxer/multiselect_communicator.py
mhchia: I don't think so. `conn.writer` is `asyncio.StreamWriter`, and `asyncio.StreamWriter.write` is a sync function.

libp2p/__init__.py
```diff
@@ -91,31 +93,29 @@ def initialize_default_swarm(
...
 
-    # TODO parse transport_opt to determine transport
-    transport_opt = transport_opt or ["/ip4/127.0.0.1/tcp/8001"]
+    # TODO: Parse `transport_opt` to determine transport
     transport = TCP()
 
     # TODO TransportUpgrader is not doing anything really
-    # TODO parse muxer and sec to pass into TransportUpgrader
-    muxer = muxer_opt or ["mplex/6.7.0"]
+    muxer_transports_by_protocol = muxer_opt or {TProtocol("/mplex/6.7.0"): Mplex}
```
pipermerriam: These strings seems like good candidates for constants.

libp2p/stream_muxer/muxer_multistream.py
```diff
@@ -0,0 +1,70 @@
...
+    ) -> None:
+        self.transports = {}
+        self.multiselect = Multiselect()
+        self.multiselect_client = MultiselectClient()
+        self.order_preference = []
+        for protocol, transport in muxer_transports_by_protocol.items():
+            self.add_transport(protocol, transport)
+
+    def add_transport(self, protocol: TProtocol, transport: MuxerClassType) -> None:
+        self.transports[protocol] = transport
```
pipermerriam: Should there be a check here to see if the `protocol/transport` is already present?  What should the behavior be if it is?  Should the `protocol` instance be removed from `order_preference` and appended to the end?  Otherwise it will be in the list twice.

mhchia: > Should there be a check here to see if the protocol/transport is already present? 

It is not checked in Go. It just removes the previous one and adds it again. Reference: https://github.com/libp2p/go-stream-muxer-multistream/blob/master/multistream.go#L35-L39

> Should the protocol instance be removed from order_preference and appended to the end?

Good catch. It should be but it is not handled in Go. I will change it.

mhchia: @pipermerriam 

I turn out to use `OrderedDict` for `MuxerMultistream.transport`. This way we can just use `transport.keys()` as the precedence when negotiation. I also remove a protocol in `add_transport` when it was already added before, to ensure that its key is reordered.

PR #253, opened 2019-08-16
Add `/plaintext/2.0.0` secure channel

### LINKED ISSUE #244: Add `/plaintext/2.0.0` protocol for insecure transport
If we want to use this protocol to test "insecurity" we should use the protocol listed in the Go implementation `/plaintext/2.0.0`, implemented here: https://github.com/libp2p/go-libp2p-core/blob/master/sec/insecure/insecure.go

mhchia:
- Depends on #252 .
- Fixes #244 .
- Based on the latest [spec](https://github.com/libp2p/specs/blob/d5ef4c9773d07613e89e508d5ddc411c2e3d3db8/plaintext/README.md).


mhchia:
Implemented it to make us able to communicate with Go before #254 is ready. I've tested this implementation with `go-libp2p-core v0.2.2`, we can now establish connections both from Python to Go and Go to Python now.  One modification to `go-libp2p` needed to be done is, need to change the config to use `/plaintext/2.0.0` over `/plaintext/1.0.0` when initializing the node, because in `go-libp2p` `/plaintext/1.0.0` is still used now.

libp2p/security/insecure/transport.py
```diff
@@ -1,12 +1,37 @@
...
+
+# Reference: https://github.com/libp2p/go-libp2p-core/blob/master/sec/insecure/insecure.go
+
+
+PLAINTEXT_PROTOCOL_ID = TProtocol("/plaintext/2.0.0")
 
 
 class InsecureSession(BaseSession):
-    pass
+    # FIXME: Update the read/write to `BaseSession`
```
NIC619: Can you expand on this comment?

libp2p/security/insecure/transport.py
```diff
@@ -21,12 +46,29 @@ class InsecureTransport(BaseSecureTransport):
...
     async def secure_outbound(self, conn: IRawConnection, peer_id: ID) -> ISecureConn:
         """
         Secure the connection, either locally or by communicating with opposing node via conn,
         for an inbound connection (i.e. we are the initiator)
         :return: secure connection object (that implements secure_conn_interface)
         """
-        return InsecureSession(self, conn, peer_id)
+        session = InsecureSession(self, conn, peer_id)
+        await session.run_handshake()
+        # TODO: Check if `remote_public_key is not None`. If so, check if `session.remote_peer`
```
NIC619: What's the `remote_public_key`? and to `check if session.remote_peer`...?

libp2p/security/insecure/transport.py
```diff
@@ -21,12 +46,29 @@ class InsecureTransport(BaseSecureTransport):
...
+        received_peer_id = session.get_remote_peer()
+        if session.get_remote_peer() != peer_id:
```
NIC619:
```suggestion
        if received_peer_id != peer_id:
```

libp2p/security/insecure/transport.py
```diff
@@ -21,12 +46,29 @@ class InsecureTransport(BaseSecureTransport):
...
+            raise UpgradeFailure(
+                "remote peer sent unexpected peer ID. "
+                f"expected={peer_id} received={received_peer_id}"
+            )
```
NIC619: Should we close the session in this case?

libp2p/security/simple/transport.py
```diff
@@ -30,6 +30,10 @@ def __init__(self, local_key_pair: KeyPair, key_phrase: str) -> None:
             )
 
         session = InsecureSession(self, conn, ID(b""))
+        # TODO: Calls handshake to make them know the peer id each other, otherwise tests fail.
```
NIC619:
```suggestion
        # TODO: Calls handshake to make them know the peer id of each other, otherwise tests fail.
```
Question:
- What do you mean by `call handshake`?
- Does test fail now because we don't `call handshake` here? or does test succeed now because we `await session.run_handshake()`?

libp2p/security/simple/transport.py
```diff
@@ -55,6 +59,10 @@ def __init__(self, local_key_pair: KeyPair, key_phrase: str) -> None:
             )
 
         session = InsecureSession(self, conn, peer_id)
+        # TODO: Calls handshake to make them know the peer id each other, otherwise tests fail.
```
NIC619: ditto

libp2p/stream_muxer/abc.py
```diff
@@ -35,6 +34,11 @@ def __init__(
         :param peer_id: peer_id of peer the connection is to
         """
 
+    @property
+    @abstractmethod
+    def initiator(self) -> bool:
```
NIC619: suggest
```suggestion
    def is_initiator(self) -> bool:
```
Maybe we can do this renaming across the codebase all together in this PR?(`initiator` -> `is_initiator`)

libp2p/stream_muxer/muxer_multistream.py
```diff
@@ -0,0 +1,71 @@
...
+        self.transports[protocol] = transport
+        self.multiselect.add_handler(protocol, None)
+        self.order_preference.append(protocol)
+
+    async def select_transport(self, conn: IRawConnection) -> MuxerClassType:
+        """
+        Select a transport that both us and the node on the
+        other end of conn support and agree on
+        :param conn: conn to choose a transport over
+        :param initiator: true if we are the initiator, false otherwise
```
NIC619:
```suggestion
```

libp2p/security/insecure/transport.py
mhchia: I haven't finished this part yet because only PeerID is used in my interop testing. Will add it later.

libp2p/security/simple/transport.py
mhchia: 1. I mean calling `InsecureSession.run_handshake`.
2. The test passes now because we `await session.run_handshake()`. I will rephrase the sentence.

libp2p/stream_muxer/abc.py
mhchia: > Maybe we can do this renaming across the codebase all together in this PR?(initiator -> is_initiator)

Sounds good to me.

libp2p/security/insecure/exceptions.py
```diff
@@ -0,0 +1,2 @@
+class UpgradeFailure(Exception):
```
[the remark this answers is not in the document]
mhchia: Sounds good to me. Maybe we should have an issue tracking the exceptions architecture.

libp2p/security/insecure/transport.py
mhchia: I use `self.writer.write` and `self.reader.read` directly, instead of `self.write` and `self.read`. Thinking we should possibly move the logic to `self.write` and `self.read`.

mhchia:
Note: Non-hacking tests against `go-libp2p` can be added once https://github.com/libp2p/go-libp2p/pull/706 is merged. Currently, I manually modify `go-libp2p` code to use the latest Insecure constructor in my local environment to enable plaintext 2.0.

libp2p/stream_muxer/muxer_multistream.py
mhchia: This should have been fixed in #252 .

libp2p/security/insecure/transport.py
```diff
@@ -1,12 +1,37 @@
...
 
 class InsecureSession(BaseSession):
-    pass
+    # FIXME: Update the read/write to `BaseSession`
+    async def run_handshake(self):
+        msg = make_exchange_message(self.local_private_key.get_public_key())
+        msg_bytes = msg.SerializeToString()
+        encoded_msg_bytes = encode_fixedint_prefixed(msg_bytes)
+        self.writer.write(encoded_msg_bytes)
+        await self.writer.drain()
```
pipermerriam: FYI, if you *think* that you might have multiple code paths concurrently calling this you might need to wrap this in a lock.  No need to do so pre-emptively but wanted to alert you to this.

see: https://github.com/ethereum/lahja/blob/93610b2eb46969ff1797e0748c7ac2595e130aef/lahja/asyncio/endpoint.py#L99-L102

libp2p/utils.py
```diff
@@ -0,0 +1,95 @@
...
+        number >>= 7
+        if number:
+            buf += bytes((towrite | 0x80,))
+        else:
+            buf += bytes((towrite,))
+            break
+    return buf
+
+
+def decode_uvarint(buff: bytes, index: int) -> Tuple[int, int]:
```
pipermerriam: are these encoding/decoding `LEB128` encoded integers?

mhchia: Yes, it is the unsigned LEB128. In go-libp2p, they use the function in standard library `binary.PutUvarint` to encode.

libp2p/security/insecure/transport.py
mhchia: Really appreciate that you bring this up! Since write is changed to an [async function in Python 3.8](https://docs.python.org/3.8/library/asyncio-stream.html#asyncio.StreamWriter.write), I will leave the pattern "write and drain" unchanged and add locks to these places.

libp2p/security/insecure/transport.py
mhchia: IMO we should leave the `Connection.close` to the caller. Will do it in `Swarm.conn_handler`.

libp2p/security/simple/transport.py
```diff
@@ -21,15 +23,19 @@ def __init__(self, local_key_pair: KeyPair, key_phrase: str) -> None:
...
         :return: secure connection object (that implements secure_conn_interface)
         """
-        await conn.write(self.key_phrase.encode())
-        incoming = (await conn.read()).decode()
+        await conn.write(encode_fixedint_prefixed(self.key_phrase.encode()))
+        incoming = (await read_fixedint_prefixed(conn)).decode()
 
         if incoming != self.key_phrase:
-            raise Exception(
+            raise SecurityUpgradeFailure(
```
NIC619: What do you think about raising exception like `HandshakeFailure` here and catch it upstream and then raise `UpgradeFailure` there(in `TransportUpgrader.upgrade_security`)?
`HandshakeFailure` here -> catch by `SecurityMultistream` and raise `MultiselectFailure` -> catch by `TransportUpgrader` and raise `UpgradeFailure`.

libp2p/security/insecure/transport.py
```diff
@@ -21,12 +44,29 @@ class InsecureTransport(BaseSecureTransport):
...
         for an inbound connection (i.e. we are the initiator)
         :return: secure connection object (that implements secure_conn_interface)
         """
-        return InsecureSession(self, conn, peer_id)
+        session = InsecureSession(self, conn, peer_id)
+        await session.run_handshake()
+        # TODO: Check if `remote_public_key is not None`. If so, check if `session.remote_peer`
+        received_peer_id = session.get_remote_peer()
+        if received_peer_id != peer_id:
+            raise SecurityUpgradeFailure(
```
NIC619: What do you think about raising exception like `HandshakeFailure` here and catch it upstream and then raise `UpgradeFailure` there(in `TransportUpgrader.upgrade_security`)?
`HandshakeFailure` here -> catch by `SecurityMultistream` and raise `MultiselectFailure` -> catch by `TransportUpgrader` and raise `UpgradeFailure`.

libp2p/stream_muxer/muxer_multistream.py
```diff
@@ -0,0 +1,71 @@
...
+        self.order_preference = []
+        for protocol, transport in muxer_transports_by_protocol.items():
+            self.add_transport(protocol, transport)
+
+    def add_transport(self, protocol: TProtocol, transport: MuxerClassType) -> None:
+        self.transports[protocol] = transport
+        self.multiselect.add_handler(protocol, None)
+        self.order_preference.append(protocol)
+
+    async def select_transport(self, conn: IRawConnection) -> MuxerClassType:
```
NIC619: What do you think about catching `MultiselectClientError` and `MultiselectError` here then raise and propagate `MultiselectFailure` exception?

Also applies to `SecurityMultistream.select_transport `.

libp2p/security/simple/transport.py
```diff
@@ -42,19 +48,23 @@ def __init__(self, local_key_pair: KeyPair, key_phrase: str) -> None:
...
+        await conn.write(encode_fixedint_prefixed(self.key_phrase.encode()))
+        incoming = (await read_fixedint_prefixed(conn)).decode()
 
         # Force context switch, as this security transport is built for testing locally
         # in a single event loop
         await asyncio.sleep(0)
 
         if incoming != self.key_phrase:
-            raise Exception(
+            raise SecurityUpgradeFailure(
```
NIC619: ditto

libp2p/security/simple/transport.py
mhchia: Sounds good to me. I will change it tomorrow.

mhchia:
Added the verification for the remote `pubkey` and `peer_id`, and store the `pubkey` in the `InsecureSession`.

libp2p/utils.py
pipermerriam: maybe this other implementation I did in py-wasm might be a useful reference.  :+1: 

https://github.com/ethereum/py-wasm/blob/master/wasm/parsers/leb128.py

libp2p/security/insecure/transport.py
```diff
@@ -1,15 +1,55 @@
...
+        if remote_peer_id_from_pubkey != remote_peer_id:
+            raise HandshakeFailure(
+                "peer id and pubkey from the remote mismatch: "
+                f"remote_peer_id={remote_peer_id}, remote_pubkey={remote_pubkey}, "
+                f"remote_peer_id_from_pubkey={remote_peer_id_from_pubkey}"
+            )
+
+        # Nothing is wrong. Store the `pubkey` and `peer_id` in the session.
+        self.remote_permanent_pubkey = remote_pubkey
+        self.remote_peer_id = ID(remote_msg.id)
```
NIC619: `self.remote_peer_id` is initialized during `__init__`, should we also compare `remote_peer_id` against `self.remote_peer_id` before setting it to new value?
https://github.com/libp2p/py-libp2p/blob/6b05d9ab8f0de5e15f8a65ffaa199e760d209cbe/libp2p/security/base_session.py#L16-L23

libp2p/security/insecure/transport.py
```diff
@@ -24,12 +64,30 @@ class InsecureTransport(BaseSecureTransport):
...
         """
         Secure the connection, either locally or by communicating with opposing node via conn,
         for an inbound connection (i.e. we are the initiator)
         :return: secure connection object (that implements secure_conn_interface)
         """
-        return InsecureSession(self, conn, peer_id)
+        session = InsecureSession(self, conn, peer_id)
+        await session.run_handshake()
+        received_peer_id = session.get_remote_peer()
+        if session.remote_permanent_pubkey is not None and received_peer_id != peer_id:
```
NIC619: IMO this check should be handled in `run_handshake`(L50).

libp2p/exceptions.py
```diff
@@ -1,6 +1,8 @@
-class ValidationError(Exception):
+class BaseLibp2pError(Exception):
+    pass
+
+
+class ValidationError(BaseLibp2pError):
     """
     Raised when something does not pass a validation check.
     """
```
NIC619:
```
"""
pass
```

libp2p/crypto/rsa.py
```diff
@@ -30,6 +35,11 @@ def new(cls, bits: int = 2048, e: int = 65537) -> "RSAPrivateKey":
     def to_bytes(self) -> bytes:
         return self.impl.export_key("DER")
 
+    @classmethod
+    def from_bytes(cls, key_bytes: bytes) -> "RSAPrivateKey":
```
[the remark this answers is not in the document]
mhchia: Probably yes if we want to initialize a `RSAPrivateKey` with raw bytes? I added this just from the symmetricity. I can remove it if you think it's better.

libp2p/exceptions.py
mhchia: pass is not necessary since we have the docstring.

libp2p/security/insecure/transport.py
mhchia: The reason why I do it in `secure_outbound` is, we only need to check "whether the `PeerID` of node we are dialing is the same as the one it send to us" in `secure_outbound`. If we put it in `run_handshake`, we will need to add redundant checks like "whether we are the initiator or not".

libp2p/security/insecure/transport.py
mhchia: Good point. Added the check in the latest commit.

libp2p/security/insecure/transport.py
mhchia: On second thought, I move all these logic inside `run_handshake`. Thanks for the idea.

libp2p/crypto/rsa.py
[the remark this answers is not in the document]
NIC619: Yeah I guess. I saw that it's added in `RSAPrivateKey` but not `Secp256k1PrivateKey` so wonder why.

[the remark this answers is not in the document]
mhchia: It is because `Secp256k1PrivateKey` already has `new`, which functions exactly as the `from_bytes`: https://github.com/libp2p/py-libp2p/pull/253/files/6177024d4fb9eaf9fc6244cf1acacc1ae2742ad6#diff-35f98b0fb8106eef634a5c8ae53c9494R30

PR #256, opened 2019-08-20
Internalize the protobuf serialization to the concept of a `Key`

ralexstokes:
Given its use across various components of `libp2p` (not just peer IDs),
it makes the abstraction cleaner to pull the serialization into the
key class and expose the canonical serialization to bytes.

### DIFF
--- a/libp2p/crypto/keys.py
+++ b/libp2p/crypto/keys.py
@@ -32,11 +32,28 @@ class Key(ABC):
         """
         ...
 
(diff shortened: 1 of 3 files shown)


NIC619:
IIRC when sending our `pubkey` over the wire we still need to serialize it(to protobuf)?
https://github.com/libp2p/py-libp2p/pull/253/commits/ad634f6050689293c165d99a58ea430b066a112b#diff-0fc20882a95215d59376cab91c759d0aR90-R92

Issue #211, opened 2019-08-02
Deprecate `SimpleSecurityTransport`


ralexstokes:
The security of this transport seems to depend on a symmetric key shared ahead of time and this class may only exist for testing purposes.

If it isn't used in the wider scope of the `libp2p` project I'd suggest just removing it after we have our `secio` implementation. In the case that it is useful for testing or something, we can move it under our `tests`... the goal here is to not provide a security transport unless it is actually providing secure communication under the security model of a user of this library.

ralexstokes:
If we move ahead with this, depends on #210 .

ralexstokes:
This transport also doesn't provide transport-level security (comms are unencrypted) so I'm definitely +1 on deprecating it

mhchia:
+1 on deprecating `SimpleSecurity`. I think it makes sense that we have `Insecure` and `Secure`(secio) in this stage.

PR #257, opened 2019-08-22
Fix the issue when importing from other package

mhchia:


mhchia:
CI failed in dbd4ed4. Looks like mypy never check these sub-packages before. 🤔 
It should be fixed by adding `# type: ignore` in simple transport, not actually fixing the issue because simple security transport should be deprecated soon. https://github.com/libp2p/py-libp2p/issues/211

PR #258, opened 2019-08-22
Adjust coincurve version to match Trinity's

NIC619:


mhchia:
Merged now for testing trinity side.

Issue #236, opened 2019-08-13
Add Identify protocol


mhchia:
Identify protocol is called right after a connection is established. A working stream handler is wirtten in https://github.com/libp2p/py-libp2p/pull/175/files#diff-d849b88ca6d47c303e72e27bb20048d6, but the logic of logic of opening the stream and the tests are not finished yet.

### Reference
- Spec: https://github.com/libp2p/specs/tree/master/identify

mhchia:
Note: Removed `Interoperability` tag. Though `IdentifyConn` is called right after a connection is established in [basic_host](https://github.com/libp2p/go-libp2p/blob/master/p2p/host/basic/basic_host.go#L244), it seems not a must to finish [`Identify` request](https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/identify/id.go#L133-L170).

cc @ralexstokes

ShadowJonathan:
Possible duplicate of #358, or is this issue closed with that merge?

ralexstokes:
We should keep #358 but i think we can go ahead and close this... feel free to re-open @mhchia if you want to track something else here

PR #268, opened 2019-08-26
Fix: `MplexStream` is not parsed

### LINKED ISSUE #259: Change LEB128 implementation
Mentioned by https://github.com/libp2p/py-libp2p/pull/253#discussion_r315775276, the implementation in py-wasm is worth referencing. At least we can add the overflow check and EOF check in our implementation.

mhchia:
After this PR, we can communicate with Go with streams with no error. I tested the `chat.py` and [`go-libp2p-examples/chat`](https://github.com/libp2p/go-libp2p-examples/tree/master/chat) with Plaintext transport.

- Close https://github.com/libp2p/py-libp2p/pull/175. There is only `Identify` protocol left, which is not necessarily required.
    - Identify protocol: https://github.com/libp2p/py-libp2p/issues/236.
- Fix #259 .

### DIFF
--- a/libp2p/network/swarm.py
+++ b/libp2p/network/swarm.py
@@ -156,15 +156,10 @@ class Swarm(INetwork):
         if not addrs:
             raise SwarmException("No known addresses to peer")
 
-        multiaddr = addrs[0]
-
         muxed_conn = await self.dial_peer(peer_id)
 
-        # Use muxed conn to open stream, which returns
-        # a muxed stream
-        # TODO: Remove protocol id from being passed into muxed_conn
-        # FIXME: Remove multiaddr from being passed into muxed_conn
-        muxed_stream = await muxed_conn.open_stream(protocol_ids[0], multiaddr)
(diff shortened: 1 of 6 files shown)


libp2p/network/swarm.py
```diff
@@ -156,15 +156,10 @@ def set_stream_handler(
...
-        multiaddr = addrs[0]
-
         muxed_conn = await self.dial_peer(peer_id)
 
-        # Use muxed conn to open stream, which returns
-        # a muxed stream
-        # TODO: Remove protocol id from being passed into muxed_conn
-        # FIXME: Remove multiaddr from being passed into muxed_conn
-        muxed_stream = await muxed_conn.open_stream(protocol_ids[0], multiaddr)
+        # Use muxed conn to open stream, which returns a muxed stream
```
ralexstokes: comment is redundant, in my opinion... this won't block merge but i may remove it w/o warning in a future PR ;)

PR #277, opened 2019-09-04
Introduces IO abstractions and classes for `msgio` IO

ralexstokes:
This PR begins the process of introducing generic IO abstractions so that we can solidify around common interfaces and not build everything on top of a "raw connection" (which provides just one implementation of the IO interfaces).

This PR also adds `msgio` classes; these don't change much over the existing functions we had, other than providing an implementation of the new IO interfaces

### DIFF
--- a/libp2p/io/abc.py
+++ b/libp2p/io/abc.py
@@ -1,0 +1,34 @@
+from abc import ABC, abstractmethod
+
+
+class Closer(ABC):
+    async def close(self) -> None:
+        ...
+
+
+class Reader(ABC):
+    @abstractmethod
+    async def read(self, n: int = -1) -> bytes:
+        ...
+
+
+class Writer(ABC):
+    @abstractmethod
+    async def write(self, data: bytes) -> int:
+        ...
+
+
+class WriteCloser(Writer, Closer):
+    pass
+
+
+class ReadCloser(Reader, Closer):
+    pass
+
+
+class ReadWriter(Reader, Writer):
+    pass
+
+
+class ReadWriteCloser(Reader, Writer, Closer):
+    pass
--- a/libp2p/io/msgio.py
+++ b/libp2p/io/msgio.py
@@ -8,14 +8,7 @@ NOTE: currently missing the capability to indicate lengths by "varint" method.
 # TODO unify w/ https://github.com/libp2p/py-libp2p/blob/1aed52856f56a4b791696bbcbac31b5f9c2e88c9/libp2p/utils.py#L85-L99  # noqa: E501
 from typing import Optional, cast
 
-from libp2p.io.abc import (
-    Closer,
-    ReadCloser,
-    Reader,
-    ReadWriteCloser,
-    WriteCloser,
-    Writer,
-)
+from libp2p.io.abc import Closer, ReadCloser, Reader, ReadWriteCloser, WriteCloser
 from libp2p.io.utils import read_exactly
 
 SIZE_LEN_BYTES = 4
@@ -42,11 +35,10 @@ async def read_next_message(reader: Reader) -> bytes:
     return await reader.read(length)
 
 
-class MsgIOWriter(Writer, Closer):
+class MsgIOWriter(WriteCloser):
     write_closer: WriteCloser
 
     def __init__(self, write_closer: WriteCloser) -> None:
-        super().__init__()
         self.write_closer = write_closer
 
     async def write(self, data: bytes) -> int:
@@ -61,7 +53,7 @@ class MsgIOWriter(Writer, Closer):
         await self.write_closer.close()
 
 
-class MsgIOReader(Reader, Closer):
+class MsgIOReader(ReadCloser):
     read_closer: ReadCloser
     next_length: Optional[int]
 
--- a/libp2p/io/msgio.py
+++ b/libp2p/io/msgio.py
@@ -32,6 +32,16 @@ def encode_msg_with_length(msg_bytes: bytes) -> bytes:
     return len_prefix + msg_bytes
 
 
+# NOTE: temporary for this PR
+encode = encode_msg_with_length
+
+
+# NOTE: temporary for this PR
+async def read_next_message(reader: Reader) -> bytes:
+    length = await read_length(reader)
+    return await reader.read(length)
+
+
 class MsgIOWriter(Writer, Closer):
     write_closer: WriteCloser
 
--- a/libp2p/io/msgio.py
+++ b/libp2p/io/msgio.py
@@ -1,24 +1,92 @@
-from libp2p.network.connection.raw_connection_interface import IRawConnection
+"""
+``msgio`` is an implementation of `https://github.com/libp2p/go-msgio`.
 
-from .exceptions import MissingLengthException, MissingMessageException
+from that repo: "a simple package to r/w length-delimited slices."
+
+NOTE: currently missing the capability to indicate lengths by "varint" method.
+"""
+# TODO unify w/ https://github.com/libp2p/py-libp2p/blob/1aed52856f56a4b791696bbcbac31b5f9c2e88c9/libp2p/utils.py#L85-L99  # noqa: E501
+from typing import Optional, cast
+
+from libp2p.io.abc import (
+    Closer,
+    ReadCloser,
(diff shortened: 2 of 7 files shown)


libp2p/io/utils.py
```diff
@@ -0,0 +1,21 @@
...
+) -> bytes:
+    """
+    NOTE: relying on exceptions to break out on erroneous conditions, like EOF
+    """
+    data = await reader.read(n)
+
+    for _ in range(retry_count):
+        if len(data) < n:
+            remaining = n - len(data)
+            data += await reader.read(remaining)
```
mhchia: Would we want to sleep for a while(e.g. 10 ms) in every loop, to yield the CPU and wait for a while for the data? Otherwise, it seems to me that this `retry_count` might be not so efficient. I checked the code in `asyncio.Reader` it seems in every call to `asyncio.Reader.read` there is a `wait_for_data`. However, in other implementations of `Reader`, `Reader.read` can possibly return right away if there is no data in buffer available.

libp2p/security/base_session.py
```diff
@@ -31,7 +31,7 @@ def __init__(
         self.remote_permanent_pubkey = None
 
         self.conn = conn
-        self.initiator = self.conn.initiator
+        self.initiator = peer_id is not None
```
mhchia: Since `initiator` is still exposed in `IRawConnection`, why is the reason we change to `peer_id is not None`?

libp2p/io/msgio.py
```diff
@@ -1,24 +1,102 @@
...
+async def read_next_message(reader: Reader) -> bytes:
+    length = await read_length(reader)
+    return await reader.read(length)
+
+
+class MsgIOWriter(Writer, Closer):
+    write_closer: WriteCloser
+
+    def __init__(self, write_closer: WriteCloser) -> None:
+        super().__init__()
```
mhchia: Do we need this?

libp2p/io/msgio.py
```diff
@@ -1,24 +1,102 @@
...
+
+    async def write_msg(self, msg: bytes) -> None:
+        data = encode_msg_with_length(msg)
+        await self.write_closer.write(data)
+
+    async def close(self) -> None:
+        await self.write_closer.close()
+
+
+class MsgIOReader(Reader, Closer):
```
mhchia: Could we just implement `ReadCloser`?

libp2p/io/msgio.py
```diff
@@ -1,24 +1,102 @@
...
+encode = encode_msg_with_length
+
+
+# NOTE: temporary for this PR
+async def read_next_message(reader: Reader) -> bytes:
+    length = await read_length(reader)
+    return await reader.read(length)
+
+
+class MsgIOWriter(Writer, Closer):
```
mhchia: Could we implement `WriteCloser`?

libp2p/io/abc.py
```diff
@@ -0,0 +1,34 @@
...
+
+class Reader(ABC):
+    @abstractmethod
+    async def read(self, n: int = -1) -> bytes:
+        ...
+
+
+class Writer(ABC):
+    @abstractmethod
+    async def write(self, data: bytes) -> int:
```
mhchia: Though in Go implementation `write` returns `int` which indicates how many bytes are written, IMO we can just return `None` in Python because so far I haven't seen any use case of this return value. And also, `asyncio.StreamWriter.write` returns `None`. What do you think about it?

libp2p/io/utils.py
ralexstokes: my thinking here is that we will yield the control flow at every `await` -- i could be wrong here but if this is in fact how python async works then looping around a bunch would imply there is not much else waiting on the event loop anyway... i figured it would become obvious if this function is a problem (e.g. it would look like blocking the event loop) at which point we can revisit this -- in the interest of time, i'd favor moving ahead and merging this for now

libp2p/security/base_session.py
ralexstokes: i ended up removing the `conn` as an argument here in a later PR (and also what happened was that I was passing in a `MsgIOReadWriter` that doesn't have the `initiator` property... something feels pretty ad-hoc to me about attaching the `initiator` flag to the connection we pass around so i'm generally working to remove it)

libp2p/io/msgio.py
ralexstokes: possibly not -- let's see what happens if i remove it

ralexstokes: i'm not super happy w/ the inheritance going on in this file at the moment

libp2p/io/utils.py
carver: Yeah, I don't have a perfect model of asyncio in my head, but `await` always yielding control flow is how I think of it. I don't really follow what alternative you're suggesting, though.

libp2p/io/msgio.py
ralexstokes: yeah that may be a better way to do this...

libp2p/io/msgio.py
ralexstokes: ditto

libp2p/io/abc.py
ralexstokes: i don't know enough about how libp2p uses the data to say.... in general a low-level write call may only write some of the data and the return information tells you that... it could be that we are working w/ abstractions that just guarantee an entire write but in general may not in the future (i.e. different async library or more performant code).... if it becomes obvious we don't need this then i'm very fine removing it :)

ralexstokes:
going to merge this in, in the interest of expediency w/ interop testing... happy to field further PR feedback

libp2p/io/utils.py
ralexstokes: the alternative would be to explicitly add something like

```python
for _ in range(retry_count):
    await asyncio.sleep(0.1)
    if len(data) < n:
        remaining = n - len(data)
        ...
```

so that we ensure to yield to the event loop on every loop iteration (... too many loops...).

something like `await asyncio.sleep(0)` is sometimes recommended to break up a chunk of pure compute so that you don't block the thread if there are other coroutines waiting to make progress. in this case i think we are fine like i said above bc we will yield on the `reader.read`.

there is the separate issue we are all touching on which is "does this function satisfy the semantics of what we want to call 'read_exactly'" and i agree its less clear that this is the case -- my intention was to just keep an eye on it :)

lithp: I'm coming in blind here but wanted to add that `asyncio.StreamReader`s already has [a `readexactly` method](https://github.com/python/cpython/blob/master/Lib/asyncio/streams.py#L1056), it might be useful to see how asyncio does it.

@mhchia I think your concern is that non-asyncio implementations of `libp2p.io.abc.Reader` might return immediately if there is no data? So this coroutine could end up spinning forever as it calls `read()` and gets nothing and calls `read()` again? I think a reasonable way to fix that situation would be to say that `read()` must always block until it can return at least one byte, another way might be to add a method like `wait_for_data` which this method can use. But adding `asyncio.sleep()` is usually a sign that there's a better design to be found somewhere, in this case it would mean the client not only does useless work every 10ms (which would hurt the battery of whatever system it's running on) but it also assumes an artificial 5ms (on average) delay before it even notices incoming messages!

mhchia: > I think your concern is that non-asyncio implementations of libp2p.io.abc.Reader might return immediately if there is no data? So this coroutine could end up spinning forever as it calls read() and gets nothing and calls read() again?

Yeah exactly. Since there is nothing mentioned in the `Reader` interface about the behavior of `read`, I was thinking people are possibly not aware that, `readexactly` doesn't help if their implementation does not blocking-wait for the first byte. Probably a note or warning in this function will be helpful.

> I think a reasonable way to fix that situation would be to say that read() must always block until it can return at least one byte, another way might be to add a method like wait_for_data which this method can use.

This way makes a lot of sense:) 

> But adding asyncio.sleep() is usually a sign that there's a better design to be found somewhere, in this case it would mean the client not only does useless work every 10ms (which would impact the battery of whatever system it's running on) but it also assumes an artificial 5ms (on average) delay before it even notices incoming messages!

Good point. I think it might be better to leave it as is.

PR #278, opened 2019-09-04
Add buffering to secio reads

ralexstokes:
Builds on top of #277.

Incorporates the new IO abstractions there and then introduces the main change: buffering in our `secio` implementation -- before, it was reading entire messages which fail if a nested protocol, e.g., works w/ varint framing


ralexstokes:
going to go ahead and merge this; expediency to interop and all... happy to take post-merge PR feedback :)

PR #279, opened 2019-09-04
Echo demo

ralexstokes:
Python implementation of https://github.com/libp2p/go-libp2p-examples/blob/master/echo/main.go

### DIFF
--- a/examples/echo/echo.py
+++ b/examples/echo/echo.py
@@ -1,0 +1,141 @@
+import argparse
+import asyncio
+import urllib.request
+
+import multiaddr
+
+from libp2p import new_node
+from libp2p.crypto.secp256k1 import create_new_key_pair
+from libp2p.network.stream.net_stream_interface import INetStream
+from libp2p.peer.peerinfo import info_from_p2p_addr
+from libp2p.typing import TProtocol
+
+PROTOCOL_ID = TProtocol("/echo/1.0.0")
+
+
+async def _echo_stream_handler(stream: INetStream) -> None:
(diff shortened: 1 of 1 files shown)


libp2p/io/msgio.py
```diff
@@ -1,24 +1,102 @@
...
+
+
+class MsgIOReader(Reader, Closer):
+    read_closer: ReadCloser
+    next_length: Optional[int]
+
+    def __init__(self, read_closer: ReadCloser) -> None:
+        # NOTE: the following line is required to satisfy the
+        # multiple inheritance but `mypy` does not like it...
+        super().__init__(read_closer)  # type: ignore
```
pipermerriam: You could do this the gross way and do `ParentClass.__init__(self, read_closer)`

(though I can't figure out which parent class `__init__` this is intended to call since neither `Reader` or `Closer` seem to implement `__init__`

ralexstokes: yeah, this part is weird which is why i'm not certain `mypy` will deal w/ it properly...

given the multiple inheritance in `MsgIOReadWriter`, the call chain is:

```
MsgIOReadWriter.__init__ -> MsgIOReader.__init__ -> MsgIOWriter.__init__
```

and `mypy` doesn't like that if you just statically look at line 71, the super class is actually `object` (which has a niladic `__init__`)....

i'm open to changing this and want to think about it some more...

ralexstokes:
going to merge so that this doesn't get stale, happy to field additional PR feedback post-merge

Issue #235, opened 2019-08-11
Fix `reset` and `close` in `Mplex`


mhchia:
Currently, EOF(or that the the stream on the other side closes their write end) is not handled in Mplex. This is necessary when we want to implement "read until eof", which blocking waits until EOF is received.
### Reference
- `io.Reader` in Go: https://golang.org/pkg/io/#Reader
- `MuxedStream` in Go: https://github.com/libp2p/go-libp2p-core/blob/master/mux/mux.go#L18-L33
- `asyncio.StreamReader` in Python: https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.read
- `socket.recv`: https://docs.python.org/3/library/socket.html#socket.socket.recv

PR #282, opened 2019-09-05
Fix mplex stream close reset

mhchia:
Fixes https://github.com/libp2p/py-libp2p/issues/235.

### DIFF
--- a/libp2p/stream_muxer/mplex/mplex_stream.py
+++ b/libp2p/stream_muxer/mplex/mplex_stream.py
@@ -1,9 +1,13 @@
 import asyncio
+from typing import TYPE_CHECKING
 
-from libp2p.stream_muxer.abc import IMuxedConn, IMuxedStream
+from libp2p.stream_muxer.abc import IMuxedStream
 
 from .constants import HeaderTags
 from .datastructures import StreamID
+
+if TYPE_CHECKING:
+    from libp2p.stream_muxer.mplex.mplex import Mplex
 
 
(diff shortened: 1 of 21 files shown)


mhchia:
Tests fail currently. Still need debugging.

mhchia:
Working on proper tests now.

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -90,63 +122,62 @@ def is_initiator(self) -> bool:
...
                     else HeaderTags.ResetReceiver
                 )
-                await self.mplex_conn.send_message(flag, None, self.stream_id)
-
-            self.local_closed = True
-            self.remote_closed = True
+                asyncio.ensure_future(
+                    self.mplex_conn.send_message(flag, None, self.stream_id)
+                )
+                await asyncio.sleep(0)
```
ralexstokes: why do we need the `sleep` here?

ralexstokes: its not clear to me how it interacts w/ the `close_lock`

ralexstokes: yeah, i'd think the usual way to approach that would be to just `await self.mplex_conn.send_message` and then if we do need some type of synchronization we can add it, integrating the other parts of the machinery that need the `asyncio.Lock` or whatever mechanism makes sense

PR #285, opened 2019-09-08
Add the pipelined version of multistream-select

ralexstokes:
When testing interop against the rust client, we realized `multistream-select` participants can pipeline the protocol suite along w/ the header into the "handshake" announcing the intention to negotiate over `multistream-select`.

This PR adds support for this version so we can interoperate w/ both the Go and Rust clients.

### DIFF
--- a/libp2p/protocol_muxer/multiselect_client.py
+++ b/libp2p/protocol_muxer/multiselect_client.py
@@ -16,28 +16,6 @@ class MultiselectClient(IMultiselectClient):
     module in order to select a protocol id to communicate over
     """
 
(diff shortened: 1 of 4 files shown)


libp2p/protocol_muxer/multiselect_client.py
```diff
@@ -10,51 +12,48 @@
 PROTOCOL_NOT_FOUND_MSG = "na"
 
 
+logger = logging.getLogger("trinity.libp2p.multiselect")
+logger.setLevel(logging.DEBUG)
```
ralexstokes: nice!

ralexstokes:
closing for now; its not clear this is needed any more

PR #287, opened 2019-09-09
Fix close/reset behavior

mhchia:
- Add tests for `close` and `reset` in `NetStream`.
- Add tests against libp2p-daemon for `close` and `reset` in `NetStream`.
- Remove `cleanup`, which produces false CI failure because it cancels tasks and then *await* on them with `asyncio.CancelledError` suppressed.

### DIFF
--- a/libp2p/network/stream/exceptions.py
+++ b/libp2p/network/stream/exceptions.py
@@ -1,0 +1,17 @@
+from libp2p.exceptions import BaseLibp2pError
+
+
+class StreamError(BaseLibp2pError):
+    pass
+
+
+class StreamEOF(StreamError, EOFError):
+    pass
+
+
+class StreamReset(StreamError):
+    pass
+
+
+class StreamClosed(StreamError):
+    pass
--- a/libp2p/stream_muxer/mplex/mplex_stream.py
+++ b/libp2p/stream_muxer/mplex/mplex_stream.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING
 
 from libp2p.stream_muxer.abc import IMuxedStream
 
@@ -8,7 +8,6 @@ from .datastructures import StreamID
 from .exceptions import MplexStreamClosed, MplexStreamEOF, MplexStreamReset
 
 if TYPE_CHECKING:
-    from typing import Any  # noqa: F401
     from libp2p.stream_muxer.mplex.mplex import Mplex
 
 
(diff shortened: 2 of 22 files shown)


mhchia:
Tested against the daemon. Should be ready.

mhchia:
Rebased to master.

libp2p/network/stream/exceptions.py
```diff
@@ -0,0 +1,17 @@
...
+class StreamEOF(StreamError, EOFError):
+    pass
+
+
+class StreamReset(StreamError):
+    pass
+
+
+class StreamClosed(StreamError):
+    pass
```
ralexstokes: nice error tree :)

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -58,19 +58,41 @@ def is_initiator(self) -> bool:
         done, pending = await asyncio.wait(  # type: ignore
             [
                 self.event_reset.wait(),
-                self.event_remote_closed.wait(),
                 self.incoming_data.get(),
+                self.event_remote_closed.wait(),
```
ralexstokes: i thought the matter here wasn't important -- is it?

libp2p/stream_muxer/mplex/mplex_stream.py
```diff
@@ -58,19 +58,41 @@ def is_initiator(self) -> bool:
...
             raise MplexStreamReset
+
+        if len(done) != 1:
+            raise Exception(f"Should be exactly 1 job in {done}.")
+        done_task = tuple(done)[0]
+        # NOTE: Ignore type check because the typeshed for `asyncio.Task` does not
+        #   have the field `_coro`.
+        coro_qualname = done_task._coro.__qualname__  # type: ignore
+        # If `qualname == "Queue.get"` then there is incoming data. We can add it to the buffer.
+        if coro_qualname == "Queue.get":
```
ralexstokes: we probably want a more stable way to do this moving forward (e.g. what happens if they rename `Queue.get` in python 3.8, etc...)

but if it works now, i think its good enough to merge in so we can leverage this in interop -- we have been seeing some stuff around not closing streams across the clients

libp2p/network/stream/exceptions.py
mhchia: Thanks a lot! Trying to make it easier to handle.

libp2p/stream_muxer/mplex/mplex_stream.py
mhchia: Just checked the source code and you're right it is not.

libp2p/stream_muxer/mplex/mplex_stream.py
mhchia: I changed the code `_wait_for_data` to compare the tasks precisely, instead of checking the function name. This should not be an issue anymore. Thanks a lot for bringing this up!

PR #292, opened 2019-09-11
Fix `transport.dial` in swarm

mhchia:
Currently, CI fails. This fix the error in https://travis-ci.com/libp2p/py-libp2p/jobs/233843093.


mhchia:
I'm going ahead to fix the failed CI.

Issue #260, opened 2019-08-23
Add proper loggings for event


mhchia:
We barely have loggings now. In the interop testing, we will need to capture logs from py-libp2p to identify whether a certain event occurred. At least we should add logs in the important parts like "connection fully established", "stream is opened", ..etc.

ShadowJonathan:
Can this issue be closed with #298, or is more logging needed?

ralexstokes:
at this point we probably always need more logging... i'd support closing this issue though due to lack of specificity

seetadev:
Wish to share that this issue was completed by @parth-soni07 : https://github.com/libp2p/py-libp2p/pull/1119 . PR meregd.

PR #298, opened 2019-09-14
Add a little more loggings

NIC619:
Was going to take #260 but while I was at it I'm not exactly sure which parts/modules to add loggings to. I was worried adding loggings to too many modules at once will be counterproductive for team at interop when debugging. So this turns into a small PR that adds logging to
- swarm
    - upgrade error in `dial_peer`
    - `new_stream`
    - upgrade error in `conn_handler`
    - `close`
    - `close_peer`
- pubsub
    - message read from stream in `continuously_read_stream`
    - new peer in `_handle_new_peer`
    - `subscribe`/`unsubscribe`
    - `publish`
    - validation error in `push_msg`
- gossipsub
    - `attach`
    - `add_peer`/`remove_peer`
    - `publish`
    - `join`/`leave`
- `floodsub`
    - `publish`

### DIFF
--- a/libp2p/network/swarm.py
+++ b/libp2p/network/swarm.py
@@ -80,14 +80,12 @@ class Swarm(INetwork):
 
     def set_stream_handler(
         self, protocol_id: TProtocol, stream_handler: StreamHandlerFn
-    ) -> bool:
+    ) -> None:
         """
         :param protocol_id: protocol id used on stream
         :param stream_handler: a stream handler instance
-        :return: true if successful
         """
         self.multiselect.add_handler(protocol_id, stream_handler)
-        return True
 
     async def dial_peer(self, peer_id: ID) -> IMuxedConn:
         """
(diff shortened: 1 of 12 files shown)


libp2p/network/swarm.py
```diff
@@ -128,11 +126,10 @@ def set_stream_handler(
         try:
             secured_conn = await self.upgrader.upgrade_security(raw_conn, peer_id, True)
         except SecurityUpgradeFailure as error:
-            # TODO: Add logging to indicate the failure
+            error_msg = "fail to upgrade security for peer %s" % peer_id
+            logger.debug(error_msg)
```
mhchia: Would it be more efficient to use `logger.debug("fail to upgrade security for peer %s", peer_id)`, for the log aggregation purpose?

libp2p/pubsub/floodsub.py
```diff
@@ -10,6 +11,9 @@
 
 PROTOCOL_ID = TProtocol("/floodsub/1.0.0")
 
+logger = logging.getLogger("libp2p.pubsub.floodsub")
```
mhchia: Could you elaborate on the advantage to have `logger` as the global, instead of a `ClassVar` in `Floodsub`?

NIC619: Just copying the loggers added during interop (with logging level set to `DEBUG`) and I guess these are temporary. So we can update it all at once later (make it a classvar and unset logging level).

PR #299, opened 2019-09-15
Add more error handling

NIC619:
Handling errors raised in
- `decode_uvarint_from_stream` / `read_varint_prefixed_bytes` / `read_delim`
    - catch the errors in `pubsub.continuously_read_stream`
    - catch the errors when handling incoming stream in `generic_protocol_handler` in `Mplex`
    - catch the errors in `MultiselectCommunicator.read`
         - raise `MultiselectCommunicatorError` and catch the error in `Multiselect` or `MultiselectClient`

### DIFF
--- a/libp2p/protocol_muxer/multiselect.py
+++ b/libp2p/protocol_muxer/multiselect.py
@@ -2,7 +2,7 @@ from typing import Dict, Tuple
 
 from libp2p.typing import StreamHandlerFn, TProtocol
 
-from .exceptions import MultiselectError
+from .exceptions import MultiselectCommunicatorError, MultiselectError
 from .multiselect_communicator_interface import IMultiselectCommunicator
 from .multiselect_muxer_interface import IMultiselectMuxer
 
@@ -46,7 +46,10 @@ class Multiselect(IMultiselectMuxer):
         # Read and respond to commands until a valid protocol ID is sent
         while True:
             # Read message
-            command = await communicator.read()
+            try:
+                command = await communicator.read()
+            except MultiselectCommunicatorError as error:
+                raise MultiselectError(str(error))
 
             # Command is ls or a protocol
             if command == "ls":
@@ -76,7 +79,10 @@ class Multiselect(IMultiselectMuxer):
         await communicator.write(MULTISELECT_PROTOCOL_ID)
 
         # Read in the protocol ID from other party
-        handshake_contents = await communicator.read()
+        try:
+            handshake_contents = await communicator.read()
+        except MultiselectCommunicatorError as error:
+            raise MultiselectError(str(error))
 
         # Confirm that the protocols are the same
         if not validate_handshake(handshake_contents):
--- a/libp2p/protocol_muxer/multiselect.py
+++ b/libp2p/protocol_muxer/multiselect.py
@@ -37,7 +37,7 @@ class Multiselect(IMultiselectMuxer):
         Negotiate performs protocol selection
         :param stream: stream to negotiate on
         :return: selected protocol name, handler function
-        :raise Exception: negotiation failed exception
+        :raise MultiselectError: raised when negotiation failed
         """
 
         # Perform handshake to ensure multiselect protocol IDs match
@@ -49,7 +49,7 @@ class Multiselect(IMultiselectMuxer):
             try:
                 command = await communicator.read()
             except MultiselectCommunicatorError as error:
-                raise MultiselectError(str(error))
+                raise MultiselectError(error)
 
             # Command is ls or a protocol
             if command == "ls":
@@ -76,7 +76,7 @@ class Multiselect(IMultiselectMuxer):
         """
         Perform handshake to agree on multiselect protocol
         :param communicator: communicator to use
-        :raise Exception: error in handshake
+        :raise MultiselectError: raised when handshake failed
         """
 
         # TODO: Use format used by go repo for messages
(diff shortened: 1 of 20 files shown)


NIC619:
I think this part(`decode_uvarint_from_stream` / `read_varint_prefixed_bytes` / `read_delim`) is ready for review as I continue with other parts.

libp2p/protocol_muxer/multiselect.py
```diff
@@ -46,7 +46,10 @@ def add_handler(self, protocol: TProtocol, handler: StreamHandlerFn) -> None:
         # Read and respond to commands until a valid protocol ID is sent
         while True:
             # Read message
-            command = await communicator.read()
+            try:
+                command = await communicator.read()
+            except MultiselectCommunicatorError as error:
+                raise MultiselectError(str(error))
```
ralexstokes:
```suggestion
                raise MultiselectError(error)
```

i think we can do this, right? (just less code...)

libp2p/protocol_muxer/multiselect.py
```diff
@@ -76,7 +79,10 @@ def add_handler(self, protocol: TProtocol, handler: StreamHandlerFn) -> None:
         await communicator.write(MULTISELECT_PROTOCOL_ID)
 
         # Read in the protocol ID from other party
-        handshake_contents = await communicator.read()
+        try:
+            handshake_contents = await communicator.read()
+        except MultiselectCommunicatorError as error:
+            raise MultiselectError(str(error))
```
ralexstokes: ditto on the `str` here

libp2p/security/secio/transport.py
```diff
@@ -408,14 +409,21 @@ def _mk_session_from(
...
-    received_nonce = await _finish_handshake(session, remote_nonce)
+    try:
+        received_nonce = await _finish_handshake(session, remote_nonce)
+    # `IOException` includes errors raised while read from/write to raw connection
+    except IOException:
+        raise SecioException("connection closed")
     if received_nonce != local_nonce:
         await conn.close()
-        raise HandshakeFailed()
+        raise InconsistentNonce()
```
ralexstokes: i like the additional level of detail on resolving these

NIC619:
> will be nice to have greater insight into what is breaking where as we use this library

Sure! Do you mean adding more comments on each `try…except` section?

libp2p/protocol_muxer/multiselect_communicator.py
```diff
@@ -11,9 +14,26 @@ def __init__(self, read_writer: ReadWriteCloser) -> None:
...
 
     async def read(self) -> str:
-        data = await read_delim(self.read_writer)
+        """
+        :raise MultiselectCommunicatorError: raised when failed to read from underlying reader
+        """
+        try:
+            data = await read_delim(self.read_writer)
+        # `IOException` includes `IncompleteReadError` and `StreamError`
+        except (ParseError, IOException, ValueError):
```
mhchia: Note: `ValueError` seems more appropriate to be replaced by other errors like `IOException`.

libp2p/pubsub/pubsub.py
```diff
@@ -154,7 +157,13 @@ def get_hello_packet(self) -> rpc_pb2.RPC:
         peer_id = stream.mplex_conn.peer_id
 
         while True:
-            incoming: bytes = await read_varint_prefixed_bytes(stream)
+            try:
+                incoming: bytes = await read_varint_prefixed_bytes(stream)
+            except (ParseError, IncompleteReadError) as error:
```
mhchia: When encountering these errors, we should handle 1) the stream is only closed(eof), then close the stream. 2) peer sent the wrong format of message, then reset.
https://github.com/libp2p/go-libp2p-pubsub/blob/master/comm.go#L38-L46

libp2p/pubsub/pubsub.py
```diff
@@ -154,7 +157,13 @@ def get_hello_packet(self) -> rpc_pb2.RPC:
...
+                logger.debug(
+                    "read corrupted data from peer %s, error=%s", peer_id, error
+                )
+                continue
```
mhchia: With either case mentioned above, we should return from the loop.
https://github.com/libp2p/go-libp2p-pubsub/blob/master/comm.go#L38-L46

libp2p/pubsub/pubsub.py
```diff
@@ -228,10 +237,20 @@ def get_msg_validators(self, msg: rpc_pb2.Message) -> Tuple[TopicValidator, ...]
...
         :param stream: newly created stream
         """
-        await self.continuously_read_stream(stream)
+        try:
+            await self.continuously_read_stream(stream)
+        except (StreamEOF, StreamReset) as error:
+            logger.debug("fail to read from stream, error=%s", error)
+            stream.reset()
+            # TODO: what to do when the stream is terminated?
+            # disconnect the peer?
```
mhchia: https://github.com/libp2p/go-libp2p-pubsub/blob/master/comm.go#L38-L46

libp2p/security/insecure/transport.py
```diff
@@ -44,12 +45,21 @@ def __init__(
...
         encoded_msg_bytes = encode_fixedint_prefixed(msg_bytes)
-        await self.write(encoded_msg_bytes)
+        try:
+            await self.write(encoded_msg_bytes)
+        except RawConnError:
+            raise HandshakeFailure("connection closed")
 
-        remote_msg_bytes = await read_fixedint_prefixed(self.conn)
+        try:
+            remote_msg_bytes = await read_fixedint_prefixed(self.conn)
```
mhchia: Note: should be changed to use `libp2p.io.msgio`

libp2p/protocol_muxer/multiselect_communicator.py
mhchia: Ah it seems `ValueError` is removed from `read_delim`?

NIC619:
after offline discussion with @mhchia , I'm going to go ahead and merge this. @mhchia will finish the PR feedback

Issue #281, opened 2019-09-04
Detect peers' disconnections in `Swarm`


mhchia:
As mentioned in https://github.com/libp2p/py-libp2p/pull/271#discussion_r320625542, we have no idea when peers disconnect. We should handle it to keep  `Swarm.connections` up to date.

PR #293, opened 2019-09-11
Refactor mplex and start to add close detection

mhchia:
- Add `SwarmConn` to make `Conn` have reference to `Swarm`. Also, refactor `generic_stream_handler` out of `Mplex`, to make testing easier.
- Add `close` and `shutting_down` logic in mplex, and detect connection failure in `handle_incoming`.
- Refactor `generic_protocol_handler` out of `Mplex`.
- Move protocol negotiation logics to `BasicHost`.
- Refactor tests.

### TODO
- [x] Refactor `Mplex.handle_incoming`.
- [x] Add tests
    - [x] `Swarm`
        - [x] Detection of peers disconnection
        - [x] `Swarm.remove_conn`
    - [x] `SwarmConn`
    - [x] `Mplex`.

PR #303, opened 2019-09-19
Refactor: `Mplex`, `Swarm`, and `BasicHost` 

mhchia:
- Extracted from https://github.com/libp2p/py-libp2p/pull/293.
- Fix https://github.com/libp2p/py-libp2p/issues/281.
- Refactor `generic_protocol_handler` out of `Mplex`. It makes protocol negotiation abstracted out of `Mplex`, which makes testing easier. For example, we can test with a `MplexStream` without the initial protocol negotiation.
  - `generic_protocol_handler` is no longer registered by default in `Swarm`. Now, it is implemented in`BasicHost`, and is registered as the `stream_handler` in `Swarm`.
- Add `SwarmConn`, which is a wrapper of `MuxedConn`, has the reference to Swarm.

### DIFF
--- a/libp2p/network/connection/swarm_connection.py
+++ b/libp2p/network/connection/swarm_connection.py
@@ -44,7 +44,6 @@ class SwarmConn(INetConn):
     async def _handle_new_streams(self) -> None:
         # TODO: Break the loop when anything wrong in the connection.
         while True:
-            print("!@# SwarmConn._handle_new_streams")
             stream = await self.conn.accept_stream()
             # Asynchronously handle the accepted stream, to avoid blocking the next stream.
             await self.run_task(self._handle_muxed_stream(stream))
@@ -57,7 +56,6 @@ class SwarmConn(INetConn):
             await self.run_task(self.swarm.common_stream_handler(net_stream))
 
     async def _add_stream(self, muxed_stream: IMuxedStream) -> NetStream:
-        print("!@# SwarmConn._add_stream:", muxed_stream)
         net_stream = NetStream(muxed_stream)
         # Call notifiers since event occurred
         for notifee in self.swarm.notifees:
@@ -65,7 +63,6 @@ class SwarmConn(INetConn):
         return net_stream
 
     async def start(self) -> None:
-        print("!@# SwarmConn.start")
         await self.run_task(self._handle_new_streams())
 
     async def run_task(self, coro: Awaitable[Any]) -> None:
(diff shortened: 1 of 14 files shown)


libp2p/network/connection/swarm_connection.py
```diff
@@ -0,0 +1,76 @@
...
+        await self.conn.close()
+        for task in self._tasks:
+            task.cancel()
+
+        # TODO: Reset streams for local.
+        # TODO: Notify closed.
+
+    async def _handle_new_streams(self) -> None:
+        # TODO: Break the loop when anything wrong in the connection.
+        while True:
```
ralexstokes: does it make sense to put this `while` block inside a `try` block and then put `await self.close()` inside the `finally:` clause?

libp2p/stream_muxer/mplex/mplex.py
```diff
@@ -34,17 +33,13 @@ class Mplex(IMuxedConn):
     next_channel_id: int
     streams: Dict[StreamID, MplexStream]
     streams_lock: asyncio.Lock
+    new_stream_queue: "asyncio.Queue[IMuxedStream]"
     shutdown: asyncio.Event
 
     _tasks: List["asyncio.Future[Any]"]
 
     # TODO: `generic_protocol_handler` should be refactored out of mplex conn.
```
ralexstokes: is this comment still relevant?

libp2p/network/connection/swarm_connection.py
mhchia: It seems pretty convenient but is it also possible that we catch unwanted exceptions? In the latest code I catch only the `MuxedConnUnavailable` in this loop.

https://github.com/libp2p/py-libp2p/pull/304/files#diff-cc33a5151ea2f5acfde91e1dd54be5d4R52-R63

libp2p/stream_muxer/mplex/mplex.py
mhchia: Ah I removed this line in #305 , but I can also fix it here.

mhchia:
Big thanks for reviewing! I will solve the conflicts first and then get this ready again tmr.

libp2p/network/connection/swarm_connection.py
mhchia: On second thought, I'm thinking it might be possibly an easier pattern for the shutting down logic in `Mplex` in #304 , will investigate more. Thanks for bringing this up.

libp2p/stream_muxer/mplex/mplex.py
ralexstokes: if you removed it elsewhere, we can leave it here -- just want to keep `master` as clean as possible :)

Issue #219, opened 2019-08-03
Propose to rename `INetwork.notify`


NIC619:
What `notify` does: register a notifee to the `Network`(i.e., `Swarm`).
https://github.com/libp2p/py-libp2p/blob/c8005c811330f540bff246838582f886da9ab39f/libp2p/network/swarm.py#L234-L242

Propose to rename it to `new_notifee` or `register_notifee` to avoid confusion. Open to other suggestions.

mhchia:
+1, I'm good with both but prefer `register_notifee` more.

ralexstokes:
yeah given this code, i think `register_notifee` is a more precise name.

ralexstokes:
another question though: does the type checker allow something that is not an `INotifee` to be used w/ this function? if not, then we don't need the `isinstance` check and can remove the boolean indication.

NIC619:
Yeah mypy will catch this and it makes sense to remove the `isinstance` check.
But other code base that use pylibp2p as library will then be able to register a non-`INotifee` type notifee. Though I'm not sure if this type of error is something that we should guard against. Do we need to add an `isinstance` check on every function arguments?🤔

mhchia:
Closed by #305

PR #305, opened 2019-09-19
Change `Notifee`, add tests for `SwarmConn` and `Mplex`

mhchia:
Based on #304 .
* Change MuxedConn in Notifee to NetConn.
* Rename `NetStream.mplex_conn` to `muxed_conn`.
* Remove `protocol_id` in `Swarm.new_stream`.
* Rename `INetwork.notify` to `register_notifee` https://github.com/libp2p/py-libp2p/issues/219.
* Refine the tests for `Notifee`.
* Add tests for `SwarmConn`, `Mplex`, and `MplexStream`.
* Wrap "calling swarm.notifee" inside `Swarm`.


mhchia:
Reflected the suggestions from previous PRs. This should be ready for review now:)

libp2p/network/swarm.py
```diff
@@ -288,19 +271,46 @@ def add_router(self, router: IPeerRouting) -> None:
...
+
+    def register_notifee(self, notifee: INotifee) -> None:
+        """
+        :param notifee: object implementing Notifee interface
+        :return: true if notifee registered successfully, false otherwise
+        """
+        self.notifees.append(notifee)
+
+    def notify_opened_stream(self, stream: INetStream) -> None:
+        asyncio.gather(
```
ralexstokes: something to think about here is what happens to these tasks after they are started by `asyncio.gather`.

afaiui, the tasks will start and then run concurrently w/ other tasks and hopefully these tasks complete. if we have to cancel the notification, we won't have a way to do it

not sure this change blocks merging in this PR as to my knowledge we aren't using the notifications system very much right now... shall we move this to an issue?

libp2p/security/security_multistream.py
```diff
@@ -31,15 +31,14 @@ class SecurityMultistream(ABC):
...
-        self, secure_transports_by_protocol: Mapping[TProtocol, ISecureTransport]
-    ) -> None:
+    def __init__(self, secure_transports_by_protocol: TSecurityOptions = None) -> None:
         self.transports = OrderedDict()
         self.multiselect = Multiselect()
         self.multiselect_client = MultiselectClient()
 
-        for protocol, transport in secure_transports_by_protocol.items():
-            self.add_transport(protocol, transport)
+        if secure_transports_by_protocol is not None:
```
ralexstokes: nitpick:

```suggestion
        if secure_transports_by_protocol:
```

ralexstokes: we could also just make this an empty tuple, then in _all_ cases, you just iterate over all of them

libp2p/__init__.py
```diff
@@ -114,7 +114,7 @@ def initialize_default_swarm(
     key_pair: KeyPair = None,
     swarm_opt: INetwork = None,
     transport_opt: Sequence[str] = None,
-    muxer_opt: Mapping[TProtocol, MuxerClassType] = None,
+    muxer_opt: Mapping[TProtocol, TMuxerClass] = None,
```
ralexstokes:
```suggestion
    muxer_opt: TMuxerOptions = None,
```

libp2p/__init__.py
```diff
@@ -114,7 +114,7 @@ def initialize_default_swarm(
...
     sec_opt: Mapping[TProtocol, ISecureTransport] = None,
```
ralexstokes:
```suggestion
    sec_opt: TSecurityOptions = None,
```

libp2p/stream_muxer/muxer_multistream.py
```diff
@@ -24,20 +22,19 @@ class MuxerMultistream:
...
-    def __init__(
-        self, muxer_transports_by_protocol: Mapping[TProtocol, MuxerClassType]
-    ) -> None:
+    def __init__(self, muxer_transports_by_protocol: TMuxerOptions = None) -> None:
         self.transports = OrderedDict()
         self.multiselect = Multiselect()
         self.multiselect_client = MultiselectClient()
-        for protocol, transport in muxer_transports_by_protocol.items():
-            self.add_transport(protocol, transport)
+        if muxer_transports_by_protocol is not None:
```
ralexstokes: see above comment about empty tuple as the default, or shortening to `if $VALUE`

libp2p/utils.py
```diff
@@ -73,9 +73,8 @@ def encode_delim(msg: bytes) -> bytes:
 
 async def read_delim(reader: Reader) -> bytes:
     msg_bytes = await read_varint_prefixed_bytes(reader)
-    # TODO: Investigate if it is possible to have empty `msg_bytes`
-    if len(msg_bytes) != 0 and msg_bytes[-1:] != b"\n":
-        raise ValueError(f'msg_bytes is not delimited by b"\\n": msg_bytes={msg_bytes}')
+    if len(msg_bytes) == 0 or msg_bytes[-1:] != b"\n":
+        raise ParseError(f'msg_bytes is not delimited by b"\\n": msg_bytes={msg_bytes}')
```
ralexstokes: the message being `0` bytes and the message not ending with a `\n` are _technically_  two different cases it could be handy to discriminate b/t

tests/factories.py
```diff
@@ -128,11 +141,37 @@ class Meta:
...
+
+
+async def mplex_stream_pair_factory(
+    is_secure: bool
+) -> Tuple[MplexStream, Swarm, MplexStream, Swarm]:
+    mplex_conn_0, swarm_0, mplex_conn_1, swarm_1 = await mplex_conn_pair_factory(
+        is_secure
+    )
+    stream_0 = await mplex_conn_0.open_stream()
+    await asyncio.sleep(0.01)
```
ralexstokes: do we need this sleep here?

tests/stream_muxer/test_mplex_stream.py
```diff
@@ -0,0 +1,182 @@
...
+    await stream_0.reset()
+
+
+@pytest.mark.asyncio
+async def test_mplex_stream_reset(mplex_stream_pair):
+    stream_0, stream_1 = mplex_stream_pair
+    await stream_0.reset()
+    await asyncio.sleep(0.01)
+
+    # Both sides are closed.
```
ralexstokes: love this testing on mplex stream :D

libp2p/security/security_multistream.py
mhchia: An empty default value sounds great to me. However, because `secure_transports_by_protocol` is of type `Mapping[TProtocol, ISecureTransport]`, we might still need the pattern for `dict` initialization.
```python
    if not secure_transports_by_protocol:
        secure_transports_by_protocol = {}
```
After tracing the code, I'd just remove the default `None` for `secure_transports_by_protocol` and `muxer_transports_by_protocol` since we didn't use and probably not even need it.

tests/factories.py
mhchia: Yes I think so. if we want to be more confident that the `mplex_conn_0` receives the `NewStream` message and accept the `MplexStream`.

tests/stream_muxer/test_mplex_stream.py
mhchia: I'm glad you like it. Now we have more detailed tests against `MplexStream`, which makes me a bit more confident in using it. 😂

libp2p/network/swarm.py
mhchia: Yes, you're right. I opened https://github.com/libp2p/py-libp2p/issues/309 to track this one.

mhchia:
@ralexstokes Big thanks for the review! I will fix https://github.com/libp2p/py-libp2p/issues/309 in the next PR.

PR #308, opened 2019-09-23
Move interop tests outside `tests`

### LINKED ISSUE #296: Remove py-libp2p-daemon
pypi rejects releases with this kind of dependency:

https://github.com/libp2p/py-libp2p/blob/811cd7813a74ef55110e1ae09dac7028c6caa0a4/setup.py#L12-L13

I had to manually remove that line from the tagged commit before releasing.

Some options:
- remove it (maybe it's cruft that is not necessary anymore?)
- inline it (maybe it's small enough that it's worth just bringing into this repo)
- do a proper release of `py-libp2p-daemon`

mhchia:
It is moved to the top level package `tests_interop`, to avoid circular dependency, with the dependency on `p2pclient` moved to `tox`.

Fixes #296 .


mhchia:
@carver 
I moved the interop tests to `tests_interop` and move the dependency on `p2pclient` from `setup.py` to `tox.ini`. It seems to work now. Could you have a look at this? Really appreciate!

mhchia:
Interop tests seemed to fail. I will fix it tomorrow.

tox.ini
```diff
@@ -27,11 +27,10 @@ skip_glob=
 
 [testenv]
 deps =
-passenv = CI TRAVIS TRAVIS_* GOPATH
+passenv = CI TRAVIS TRAVIS_*
 extras = test
 commands =
-    test: py.test --ignore=tests/interop
-    interop: py.test tests/interop
+    py.test tests/
```
carver: I can maybe imagine this command accidentally being run in future test environments that we didn't intend. One option is to re-add the `test:` prefix.

tox.ini
```diff
@@ -40,6 +39,15 @@ basepython = python3
...
+    black --check libp2p tests tests_interop examples setup.py
+    isort --recursive --check-only libp2p tests tests_interop examples setup.py
+    flake8 libp2p tests tests_interop examples setup.py
+
+[testenv:py37-interop]
+deps = p2pclient
+passenv = CI TRAVIS TRAVIS_* GOPATH
+extras = test
+commands =
+    py.test tests_interop/
```
carver: *nitpick* The recommendation from pytest is to run the command without a dot: https://github.com/pytest-dev/pytest/issues/1629

mhchia:
@carver Thanks for the review!

PR #310, opened 2019-09-24
Add the missing `await` for `stream.reset` in pubsub

mhchia:


mhchia:
@NIC619 Thanks for the review:)

PR #314, opened 2019-09-25
Add default protocols

ralexstokes:
Adds the option for all hosts to have some set of protocols by default.

This PR sets the infrastructure for the next PR which adds the `ping` and `identify` protocols for all hosts, by default.

### DIFF
--- a/libp2p/host/basic_host.py
+++ b/libp2p/host/basic_host.py
@@ -1,8 +1,9 @@
 import logging
-from typing import List, Sequence
+from typing import TYPE_CHECKING, List, Sequence
 
 import multiaddr
 
+from libp2p.host.defaults import DEFAULT_HOST_PROTOCOLS
 from libp2p.host.exceptions import StreamFailure
 from libp2p.network.network_interface import INetwork
 from libp2p.network.stream.net_stream_interface import INetStream
@@ -16,6 +17,9 @@ from libp2p.protocol_muxer.multiselect_communicator import MultiselectCommunicat
 from libp2p.typing import StreamHandlerFn, TProtocol
 
 from .host_interface import IHost
+
+if TYPE_CHECKING:
+    from collections import OrderedDict
 
 # Upon host creation, host takes in options,
 # including the list of addresses on which to listen.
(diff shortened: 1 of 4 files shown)


libp2p/protocol_muxer/multiselect.py
```diff
@@ -19,8 +19,8 @@ class Multiselect(IMultiselectMuxer):
 
     handlers: Dict[TProtocol, StreamHandlerFn]
 
-    def __init__(self) -> None:
-        self.handlers = {}
+    def __init__(self, default_handlers: Dict[TProtocol, StreamHandlerFn] = {}) -> None:
```
mhchia: Should we initialize `default_handlers = {}` in the function body to avoid one `{}` is shared in all `Multiselect` instance?

ralexstokes: yes, nice catch -- i was thinking python wouldn't do this for `{}` but it ends up as a reference type so we get the clobbering behavior

ralexstokes:
this is ready for merge, pending any more review...

i'll add testing for defaults implicitly in future PRs that add some default protocols

libp2p/host/basic_host.py
```diff
@@ -41,13 +45,18 @@ class BasicHost(IHost):
     multiselect: Multiselect
     multiselect_client: MultiselectClient
 
-    def __init__(self, network: INetwork, router: KadmeliaPeerRouter = None) -> None:
+    def __init__(
+        self,
+        network: INetwork,
+        router: KadmeliaPeerRouter = None,
+        default_protocols: "OrderedDict[TProtocol, StreamHandlerFn]" = DEFAULT_HOST_PROTOCOLS,
```
mhchia: Sorry, I found I missed this one when I reviewed last time. `DEFAULT_HOST_PROTOCOLS` seems fine to be shared across `BasicHost` instances since we are probably not going to modify it, but I'm a little bit worried `Multiselect` might accidentally modify it. IMO we have two options here: 1) still initialize `default_protocols` in the function body. 2) change `Multiselect`, make it not modify the passed `handlers`(i.e. we probably need to remove `add_handler`).

libp2p/protocol_muxer/multiselect.py
pipermerriam: I believe [`flake8-bugbear`](https://github.com/PyCQA/flake8-bugbear) would have caught this in CI.

ralexstokes: good to know, i'll make an issue to add it

PR #320, opened 2019-10-01
Remove forced debug level

moshemalawach:
As a best practice, the debug level should be set by the calling applications (example codes or unit tests fixtures if you want!).

Having it forced as debug was killing my whole logging system (as it is configurable, logger per logger or global).


mhchia:
It seems the failure is irrelevant to this PR. I will go ahead and merge it, and fix the mypy error later.

Issue #322, opened 2019-10-02
`mypy` failure: CI fails as mypy updates


mhchia:
Recently, `mypy` released `0.730`, which check the types differently for `asyncio.wait`. I'm figuring out what's wrong.

```bash
error: Argument 1 to "wait" has incompatible type "List[object]"; expected "Iterable[Union[Generator[Any, None, <nothing>], Awaitable[<nothing>]]]"
```

Ref: https://travis-ci.com/libp2p/py-libp2p

PR #323, opened 2019-10-02
Ignore typing for `asyncio.wait`

mhchia:
Fix https://github.com/libp2p/py-libp2p/issues/322. CI failed after `mypy` releasing `0.730`. It seems because in the new version of `mypy` we need one more `# type: ignore` to suppress the typing error of `asyncio.wait`(one is for "add typing for `done` and `pending` and the other is for "the first argument `fs` type mismatch".)

#### Note
This PR doesn't fix the root cause of the error. We should investigate later to solve the typing for `asyncio.wait` later.


mhchia:
I'll just go ahead to unblock fix the CI.

PR #325, opened 2019-10-14
Do not add addresses if they are already there in peerdata

moshemalawach:
In peerdata, if you call add_addrs, the function doesn't check if the addresses are already there.
This can result in the list growing huge if we reconnect often to the same hosts.

An alternative solution to the one proposed would be to change self.addrs to a set and not a list. But that would need some work in places where we use addrs[0] (like in swarm).


libp2p/peer/peerdata.py
```diff
@@ -26,7 +26,9 @@ def set_protocols(self, protocols: Sequence[str]) -> None:
         self.protocols = list(protocols)
 
     def add_addrs(self, addrs: Sequence[Multiaddr]) -> None:
-        self.addrs.extend(addrs)
+        for addr in addrs:
+            if addr not in self.addrs:
+                self.addrs.append(addr)
```
mhchia: What I'm thinking is probably we should change `addrs` from `List[Multiaddr]` to `Set[Multiaddr]` to reduce time complexity. However, this solution works for me so far. 👍

ralexstokes:
i'd suggest we use a `set` for this use case.

do you want to update the PR @moshemalawach ?

if not, i'll move this to a tracking issue and we can go ahead and merge this in.

ralexstokes:
and +1 on a test for this scenario :)

libp2p/peer/peerdata.py
Hiteon88: 0xCf5dE5aEbA9cf8503E6031deAf1279b4f6Db1223

pacrob:
closed by #485

PR #327, opened 2019-10-15
Add udp

### LINKED ISSUE #326: Add UDP
I have made a first draft for implementing UDP in Transport Layer.
It is supposed to be implemented in Go, however master branch is [empty](https://github.com/libp2p/go-udp-transport/). The only code we can find is in the second branch [feat/udp](https://github.com/libp2p/go-udp-transport/tree/feat/udp).

So, my approach is not based at all in Go's "implementation". I propose a unified interface for UDP and TCP Streams: [IStreamReader](https://github.com/aratz-lasa/py-libp2p/blob/5191704ff34ad9aef48f94e4b380d03ad1e5325e/libp2p/transport/stream_interface.py#L15) and [IStreamWriter](https://github.com/aratz-lasa/py-libp2p/blob/5191704ff34ad9aef48f94e4b380d03ad1e5325e/libp2p/transport/stream_interface.py#L21). Apart from getting a unified interface, I have also tried to keep it simple. Now, streams just implement four methods: **get_extra_info**, **close**, **read** and **write**. I may have made it too high-level abstracting from methods like **drain** or others. I let it for discussion ;) 

 It may look weird UDP implementing a Stream interface, however, in order to achieve modularity between transports, we must simulate a Session-like-stream on UDP. Due to this Unified Interface, every transport protocol will need to implement four interfaces: **ITransport**, **IListener** and **IStreamReader** and **IStreamWriter**. You can find the new Stream interfaces here: [UDP Streams](https://github.com/aratz-lasa/py-libp2p/blob/add-udp/libp2p/transport/udp/udp_stream.py) and [TCP Streams](https://github.com/aratz-lasa/py-libp2p/blob/add-udp/libp2p/transport/tcp/tcp_stream.py). On the other hand, I think this two last interfaces could be gatherer into one interface, however, I have prefered to divide them, for being compatible with current **RawConnection**, **Swarm**....
  
Of course, this code is just a playeable implementation, intended for encouraging discussions and give a new perspective. I have not made any tests for UDP yet, even though, TCP passes current tests while implementing this new Unified Interface.

aratz-lasa:
This is the code related to UDP implementation, where it is trying to Resolve #326 .

### DIFF
--- a/libp2p/transport/udp/udp.py
+++ b/libp2p/transport/udp/udp.py
@@ -1,0 +1,173 @@
+import asyncio
+from typing import Any, Dict, List, Tuple
+
+from multiaddr import Multiaddr
+
+import asyncio_dgram
+from asyncio_dgram.aio import DatagramStream
+from libp2p.network.connection.raw_connection import RawConnection
+from libp2p.network.connection.raw_connection_interface import IRawConnection
+from libp2p.transport.exceptions import OpenConnectionError
+from libp2p.transport.listener_interface import IListener
+from libp2p.transport.stream_interface import IStreamReader, IStreamWriter
+from libp2p.transport.transport_interface import ITransport
+from libp2p.transport.typing import THandler
+from libp2p.transport.udp.udp_stream import (
+    UDPServerStreamReader,
+    UDPServerStreamWriter,
+    UDPStreamReader,
+    UDPStreamWriter,
+)
+
+
+async def open_datagram_connection(
+    host: str, port: int
+) -> Tuple[IStreamReader, IStreamWriter]:
+    stream = await asyncio_dgram.connect((host, port))
+    reader, writer = UDPStreamReader(stream), UDPStreamWriter(stream)
+    return reader, writer
+
+
+def start_udp_server(handler: THandler, host: str, port: int) -> "UDPServer":
+    server = UDPServer(handler, host, port)
+    asyncio.create_task(server.listen())
+    return server
+
+
+class UDPServer:
+    _host: str
+    _port: int
+    _stream: DatagramStream
+    _handler_queues: Dict[Any, asyncio.Queue[bytes]]
+    _running_handlers: List[asyncio.Task[Any]]
+
+    def __init__(self, handler: THandler, host: str, port: int):
+        _handler: THandler  # in class-level, MyPy raises error, because it treats like a method
+        self._handler = handler
+        self._host = host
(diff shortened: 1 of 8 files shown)


libp2p/transport/udp/udp.py
```diff
@@ -0,0 +1,163 @@
...
+        self._stream = await asyncio_dgram.bind((self._host, self._port))
+        while True:
+            data, addr = await self._stream.recv()
+            await self._deliver_data(addr, data)
+
+    async def close(self) -> None:
+        self._stream.close()
+        # TODO: Correct way to destroy all handlers?
+        for task in self._running_handlers:
+            task.cancel()
```
pipermerriam: The *proper* pattern here according to the asyncio docs is:

```python
task.cancel()
try:
    await task
except asyncio.CancelledError:
    pass
```

This gives the loop the opportunity to inject the cancellation exception into the task so that it doesn't end up being reported as an unretrieved exception.

aratz-lasa:
This is still just a "prototype" of what could be a UDP implementation. I just made the pull request for being easily referenceable from the issue.
Are there any plans for implementing in Trio? I didn't know it. 
By the way I love trio. Even though it is not  easily compatible withasyncio compatible.

aratz-lasa:
Sorry for closing @pipermerriam I was on the phone, and I hit "Close and comment" accidentally.

libp2p/transport/udp/udp.py
aratz-lasa: Thanks! I missed it. And probably there are more things, I made without much revising.

pipermerriam:
>  Are there any plans for implementing in Trio? I didn't know it.

If you would like to work on that I'd be happy to help in any way I can.  That could be reviewing pull requests, funding a small grant, etc.

aratz-lasa:
> If you would like to work on that I'd be happy to help in any way I can. That could be reviewing pull requests, funding a small grant, etc.

We should first check how asyncio compatibility would work, because most people use asyncio. Or which limitations/difficulties may arise. However, I think, Trio's sockets API (apart from Nursery) are much better than asyncio's. And even more for getting a Unified Interface for TCP and UDP.

pipermerriam:
The plan would be to drop compatibility with `asyncio`, converting this to a pure `trio` based library.

pacrob:
Have decided we don't need UDP support, as QUIC is preferred.

PR #331, opened 2019-10-24
Add automatic docstring formatting

dmuhs:
This PR will
- introduce a `dev` setup dependency for [docformatter](https://github.com/myint/docformatter)
- format all docstrings we have so far to adhere to the standard formatting
- add automatic formatting to the main Makefile's `format` command

Looking through the code base, I have found that a lot of docstrings were inconsistent in line length, line formattling (i.e. one short description in the first line, followed by an empty line, a more verbose description, and optional parameter definitions). Similar to `black`, which we are already using, `docformatter` aims to automatically format docstrings according to [PEP257](https://www.python.org/dev/peps/pep-0257/).

It is worth noting that PEP257 is pretty hard to enforce programmatically, which is also why `docformatter` only implements a subset of the rules outlined above. This means that some manual intervention is required, however the [features](https://github.com/myint/docformatter#features) are valuable enough for maintaining docstring consistency across all code.

### DIFF
--- a/libp2p/__init__.py
+++ b/libp2p/__init__.py
@@ -48,7 +48,8 @@ def generate_peer_id_from(key_pair: KeyPair) -> ID:
 def initialize_default_kademlia_router(
     ksize: int = 20, alpha: int = 3, id_opt: ID = None, storage: IStorage = None
 ) -> KadmeliaPeerRouter:
-    """initialize kadmelia router when no kademlia router is passed in.
+    """
+    initialize kadmelia router when no kademlia router is passed in.
 
     :param ksize: The k parameter from the paper
     :param alpha: The alpha parameter from the paper
@@ -78,7 +79,8 @@ def initialize_default_swarm(
     peerstore_opt: IPeerStore = None,
     disc_opt: IPeerRouting = None,
 ) -> Swarm:
-    """initialize swarm when no swarm is passed in.
+    """
+    initialize swarm when no swarm is passed in.
 
     :param id_opt: optional id for host
     :param transport_opt: optional choice of transport upgrade
@@ -118,7 +120,8 @@ async def new_node(
     peerstore_opt: IPeerStore = None,
     disc_opt: IPeerRouting = None,
 ) -> IHost:
-    """create new libp2p node.
+    """
+    create new libp2p node.
 
     :param key_pair: key pair for deriving an identity
     :param swarm_opt: optional swarm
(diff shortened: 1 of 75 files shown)


libp2p/__init__.py
```diff
@@ -50,8 +48,8 @@ def generate_peer_id_from(key_pair: KeyPair) -> ID:
 def initialize_default_kademlia_router(
     ksize: int = 20, alpha: int = 3, id_opt: ID = None, storage: IStorage = None
 ) -> KadmeliaPeerRouter:
-    """
-    initialize kadmelia router when no kademlia router is passed in
+    """initialize kadmelia router when no kademlia router is passed in.
```
pipermerriam: I really dislike this format for docstrings, placing the first line on the same line as the triple quotes.

It appears that PEP257 allows for either via the `--pre-summary-newline` CLI option.  Thoughts on updating this to use that format?

> The summary line may be on the same line as the opening quotes or on the next line.

dmuhs: While this is mostly a matter of taste, I agree with the point because the previous format already included a pre-summary newline and I didn't come here to flip everything around completely :smile: Added the pre-summary newline in bafdd85 and updated the `make format` docformatter command

dmuhs: Updated the lintroll command and tox CI commands with a doc format check in 3439a2c :+1:

libp2p/kademlia/protocol.py
```diff
@@ -11,15 +11,11 @@
...
-    There are four main RPCs in the Kademlia protocol
-    PING, STORE, FIND_NODE, FIND_VALUE
-    PING probes if a node is still online
-    STORE instructs a node to store (key, value)
-    FIND_NODE takes a 160-bit ID and gets back
-    (ip, udp_port, node_id) for k closest nodes to target
-    FIND_VALUE behaves like FIND_NODE unless a value is stored
-    """
+    """There are four main RPCs in the Kademlia protocol PING, STORE,
+    FIND_NODE, FIND_VALUE PING probes if a node is still online STORE instructs
```
pipermerriam: This docstring got messed up.  Looks like it should be manually re-formatted so that the definitions of each message type are done using a bulletted list.

pipermerriam:
@dmuhs to avoid needing to look back over the whole thing I decided to go ahead and merge.  Would you mind submitting a small follow up PR fixing the docstring I commented on needing cleaning up?

PR #340, opened 2019-11-04
Register for disconnected event notification by pubsub

### LINKED ISSUE #306: Pubsub: handle stream closed
After reviewing the PR https://github.com/libp2p/py-libp2p/pull/299#discussion_r326230742,   I found that the stream exceptions are not handled completely in the current codebase. We should revisit it and add tests for the situation when a pubsub stream is closed/reset.

### Reference
- Go implementation: https://github.com/libp2p/go-libp2p-pubsub/blob/master/comm.go#L38-L46

NIC619:
Implement `disconnected()` in `PubsubNotifee` so pubsub can be notified of peer who has disconnected from us and remove the peer's info from pubsub records.
I believe this can close #306 (or it's already solved?).

### DIFF
--- a/libp2p/network/connection/swarm_connection.py
+++ b/libp2p/network/connection/swarm_connection.py
@@ -43,6 +43,9 @@ class SwarmConn(INetConn):
         # We *could* optimize this but it really isn't worth it.
         for stream in self.streams:
             await stream.reset()
+        # Force context switch for stream handlers to process the stream reset event we just emit
+        # before we cancel the stream handler tasks.
+        await asyncio.sleep(0.1)
 
         for task in self._tasks:
             task.cancel()
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -60,6 +60,7 @@ class Pubsub:
     router: "IPubsubRouter"
 
     peer_queue: "asyncio.Queue[ID]"
+    dead_peer_queue: "asyncio.Queue[ID]"
 
     protocols: List[TProtocol]
 
@@ -100,7 +101,10 @@ class Pubsub:
 
         # Register a notifee
         self.peer_queue = asyncio.Queue()
-        self.host.get_network().register_notifee(PubsubNotifee(self.peer_queue))
+        self.dead_peer_queue = asyncio.Queue()
+        self.host.get_network().register_notifee(
+            PubsubNotifee(self.peer_queue, self.dead_peer_queue)
+        )
 
         # Register stream handlers for each pubsub router protocol to handle
         # the pubsub streams opened on those protocols
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -253,10 +253,10 @@ class Pubsub:
         try:
             await self.continuously_read_stream(stream)
         except StreamEOF as error:
-            stream.close()
+            await stream.close()
             logger.debug("fail to read from peer %s, error=%s", peer_id, error)
         except (ParseError, IncompleteReadError, StreamReset) as error:
-            stream.reset()
+            await stream.reset()
             logger.debug("read corrupted data from peer %s, error=%s", peer_id, error)
 
     async def _handle_new_peer(self, peer_id: ID) -> None:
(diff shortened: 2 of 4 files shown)


libp2p/network/connection/swarm_connection.py
```diff
@@ -43,6 +43,9 @@ def __init__(self, muxed_conn: IMuxedConn, swarm: "Swarm") -> None:
         # We *could* optimize this but it really isn't worth it.
         for stream in self.streams:
             await stream.reset()
+        # Force context switch for stream handlers to process the stream reset event we just emit
+        # before we cancel the stream handler tasks.
+        await asyncio.sleep(0.1)
```
pipermerriam: This is a [code smell](https://ethpm-cli.readthedocs.io/en/latest/)

I would suggest looking into a mechanism to allow you to directly `await` this type of thing.  For example, `await stream.reset()` could return an object that could then be awaited and wouldn't return until the underlying handlers have processed the event.  Otherwise you're likely to continue to have to letter random sleeps into your code and I can say from experience in trinity that we don't want this.

libp2p/pubsub/pubsub.py
```diff
@@ -247,13 +246,18 @@ def get_msg_validators(self, msg: rpc_pb2.Message) -> Tuple[TopicValidator, ...]
...
-        except (StreamEOF, StreamReset) as error:
-            logger.debug("fail to read from stream, error=%s", error)
-            await stream.reset()
-            # TODO: what to do when the stream is terminated?
-            # disconnect the peer?
+        except StreamEOF as error:
+            stream.close()
+            logger.debug("fail to read from peer %s, error=%s", peer_id, error)
+        except (ParseError, IncompleteReadError, StreamReset) as error:
+            stream.reset()
```
pipermerriam: Why isn't this being awaited anymore?

libp2p/pubsub/pubsub.py
```diff
@@ -294,6 +309,21 @@ def get_msg_validators(self, msg: rpc_pb2.Message) -> Tuple[TopicValidator, ...]
...
+        """
+        while True:
+
+            peer_id: ID = await self.dead_peer_queue.get()
+
+            # Remove Peer
+            asyncio.ensure_future(self._handle_dead_peer(peer_id))
+
+            # Force context switch
+            await asyncio.sleep(0)
```
pipermerriam: Why do you need this?  The context switch will happen as soon as the next iteration of the loop occurs and you start awaiting the next dead peer?

libp2p/pubsub/pubsub.py
```diff
@@ -137,6 +141,7 @@ def __init__(
 
         # Call handle peer to keep waiting for updates to peer queue
         asyncio.ensure_future(self.handle_peer_queue())
+        asyncio.ensure_future(self.handle_dead_peer_queue())
```
pipermerriam: You aren't keeping track of this task I think implies that this might lead to unclean shutdowns?

libp2p/network/connection/swarm_connection.py
NIC619: `await stream.reset()` along with `await self.muxed_conn.close()` on L40 will `set` the `event`s for `Mplex` and `Mplex_stream`. This signals that `Mplex` and its Mplex_stream`s are shutting down. Then when stream handler try to read from stream it will catch `StreamReset` error and properly finishes.

So the problem is that changing the state of the stream does not guarantee that the stream handler task that's reading from the stream will detect the state change and finish right after. So in this case if we cancel tasks right after `await stream.reset()` the stream handler task will be cancelled instead of catching `StreamReset` error and properly finishes.

Maybe we can cancel other tasks and wait for stream handler task to finish?

libp2p/pubsub/pubsub_notifee.py
```diff
@@ -41,7 +49,14 @@ def __init__(self, initiator_peers_queue: "asyncio.Queue[ID]") -> None:
...
     async def disconnected(self, network: INetwork, conn: INetConn) -> None:
-        pass
+        """
+        Add peer_id to dead_peers_queue, so that pubsub and its router can
+        remove this peer_id and close the stream inbetween.
+
+        :param network: network the connection was opened on
+        :param conn: connection that was opened
+        """
+        await self.dead_peers_queue.put(conn.muxed_conn.peer_id)
```
mhchia: Can I understand `handle_dead_peer_queue` is called twice when a peer disconnects us?

NIC619: Yes, `_handle_dead_peer` might be called twice. Once when we disconnect from peer(`network` notify `pubsub`) and once when pubsub detect that the connection has closed.

Issue #333, opened 2019-10-25
Add more linting tooling


ralexstokes:
Tools to add/investigate:

https://github.com/PyCQA/flake8-bugbear

PR #341, opened 2019-11-04
Add more linting

hukkinj1:
* Add `warn_unreachable = True` config for mypy
* Add flake8-bugbear. No version restriction due to flake8-bugbear using https://calver.org/ not semantic versioning.

Possibly closes https://github.com/libp2p/py-libp2p/issues/333


ralexstokes:
thanks @hukkinj1 !

PR #343, opened 2019-11-06
Add pytest xdist and fix some issues w/ parallelizing tests

### LINKED ISSUE #342: Add `pytest-xdist` to parallelize tests
Let's add `pytest-xdist` to `setup.py` deps.

When I run it locally, I seem to get two failures in our kademlia code, possibly suggesting a test that isn't parallelizable (issue is extant allocation of port numbers).

We should debug and then run tests in parallel to save time!

ralexstokes:
Fixes #342.

Adding `pytest-xdist` reveals some issues where tests were not easily parallelizable w/ hard-code ports. Fix those tests by allowing the OS to choose a free port as needed.

### DIFF
--- a/libp2p/host/ping.py
+++ b/libp2p/host/ping.py
@@ -13,9 +13,8 @@ logger = logging.getLogger("libp2p.host.ping")
 
 
 async def _handle_ping(stream: INetStream, peer_id: PeerID) -> bool:
-    """
-    Return a boolean indicating if we expect more pings from the peer at ``peer_id``.
-    """
+    """Return a boolean indicating if we expect more pings from the peer at
+    ``peer_id``."""
     try:
         payload = await asyncio.wait_for(stream.read(PING_LENGTH), RESP_TIMEOUT)
     except asyncio.TimeoutError as error:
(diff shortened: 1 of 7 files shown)


libp2p/host/ping.py
```diff
@@ -13,9 +13,8 @@
 
 
 async def _handle_ping(stream: INetStream, peer_id: PeerID) -> bool:
-    """
-    Return a boolean indicating if we expect more pings from the peer at ``peer_id``.
-    """
+    """Return a boolean indicating if we expect more pings from the peer at
```
ralexstokes: comment here

ralexstokes:
anyone who wants to review: this should be ready to merge. the main thing is adding `pytest-xdist` and changing some of the Kademlia code to not conflict on ports. it seems like i had to include a linter fix that broke master in 700209c50a115d144f9efbb4117ac2718c1f5c88, waiting on some feedback from the contributors that put `docformatter` in place

tests/utils.py
```diff
@@ -45,7 +45,9 @@
     return tuple(nodes_list)
 
 
-async def set_up_routers(router_confs):
+async def set_up_routers(router_confs=[0, 0]):
```
mhchia: Should we avoid the mutable default value here though it's unlikely to be changed? It seems we can simply change to a tuple `(0, 0)` to avoid this.

ralexstokes: yep, good catch

ralexstokes:
see my edited comment above re: the docstring thing

i'll go ahead and merge this

pipermerriam:
Your assessment of the docformatter matches mine in that:

1. it is expected behavior that docstrings with only a single line do not get a leading newline.
2. it would be nice if they did but I can accept the status quo of them getting formatted this way because I like auto-formatting.

PR #345, opened 2019-11-06
Install default protocols

ralexstokes:
Adds `identify` and `ping` as default protocols.


ralexstokes:
Once #337 is merged I can rebase.

This PR is only the last two commits in the stream.

Issue #294, opened 2019-09-11
Add tooling for docs, release, etc


carver:
Want to be able to run `make release bump=minor` to trigger a release of the pip package, the docs build, etc.

carver:
Feel free to assign to me.

dmuhs:
I have had very good experiences with [bumpversion](https://pypi.org/project/bumpversion/). You supply it with some regex replacements, it automatically parses the version (format `{major}.{minor}.{patch}`), and bumps it automatically. You can configure it to automatically tag the branch and create a version bump commit as well. I use it to automatically bump the version, create all necessary commits and tags, and let my packages' CD pipeline handle all the publishing. With the tool installed, you simply type `bumpversion patch`, and `git push --tags && git push` and boom, there's the patch release.

An example `setup.cfg` with `bumpversion` enabled, and two regex replacements defined: https://github.com/dmuhs/mythx-cli/blob/master/setup.cfg#L1

carver:
Yup, bumpversion is the utility we tend to use in my team's projects!

Check out how we do releases in Trinity:  https://trinity-client.readthedocs.io/en/latest/contributing.html#releasing

Relevant Makefile excerpt:
https://github.com/ethereum/trinity/blob/24213d8f22a03c370313d3a990eea6a2fe1e76e4/Makefile#L72-L84

My plan was to basically duplicate that approach, since I am offering to help with releases, and it's the workflow I prefer.

mhchia:
Hey @carver , I just made a release https://github.com/libp2p/py-libp2p/releases/tag/v0.1.2 . Would you mind helping us to push the new package to PyPI?

> My plan was to basically duplicate that approach, since I am offering to help with releases, and it's the workflow I prefer.

Also, I opened https://github.com/libp2p/py-libp2p/issues/352 and was thinking it might probably be the time to have the release procedure discussed:). It will be great to have your PR for the plan you describe. I can also help to duplicate the approach if you don't have the bandwidth to handle it.

carver:
Yup, I'll get it out today. FWIW, the tagged build is failing (though it's only a `lint` failure): https://travis-ci.com/libp2p/py-libp2p

I'll start adding the tooling and dogfood it to run the release, then open a PR.

carver:
v0.1.2 is out: https://pypi.org/project/libp2p/#history

---

The goal is to have a release process very similar to: https://trinity-client.readthedocs.io/en/latest/contributing.html#releasing

[...]

PR #353, opened 2019-11-14
Pre-release testing tools, plus setup.py updates

carver:
First steps of #294 

There's still a bunch more to do, but I'd like to use it once myself before opening a PR for review.

Next time we need a release, don't tag it (the release process adds the tag) and I'll take care of it.

Some notable things left to add:
- sphinx docs
- bumpversion
- towncrier for release notes
- github templates for issue, PR, etc

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,15 @@ protobufs: $(PY)
 %_pb2.py: %.proto
 	protoc --python_out=. --mypy_out=. $<
 
-.PHONY: clean
+clean-proto:
+	rm -f $(PY) $(PYI)
 
 clean:
-	rm -f $(PY) $(PYI)
+	find . -name '__pycache__' -exec rm -rf {} +
+	rm -fr build/
+	rm -fr dist/
+	rm -fr *.egg-info
+
+package: clean
+	python setup.py sdist bdist_wheel
+	python scripts/release/test_package.py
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -27,7 +27,15 @@ protobufs: $(PY)
 %_pb2.py: %.proto
 	protoc --python_out=. --mypy_out=. $<
 
-.PHONY: clean
-
-clean:
+clean-proto:
 	rm -f $(PY) $(PYI)
+
+clean-build:
```
carver: I would prefer to just call this `clean` but since there was already a `clean` target, I didn't want to leave any ambiguity for people who were used to the old target.

Still, I think that we want `clean` to be something that we're comfortable doing before every release, and running the old `clean` definitely causes problems without doing some kind of follow-up commands (which I didn't have a chance to look into).

mhchia: Yeah, agree. Old `clean` actually removes the compiled `pb2.py` and the mypy stubs from the pb(protobuf) files. I'm for having `clean-build` as `clean` actually, since `rm -f $(PY) $(PYI)` is not frequently used from my experience.

carver:
> If we want to make a release, should we just do `bumpversion` with `tag = False` and then ping you for the later actions?

Ah, nope, the version bump itself happens at release time. Part of the reason is that documentation compilation happens just before the version bump.

Makefile
carver: Sold! I made the change.

PR #355, opened 2019-11-17
Check if entry exists in dictionary before delete

NIC619:
Ran into `KeyError` when our peer in a stream accidentally `close` and `reset` the stream because handlers for both will try to `del streams[stream_id]` without checking if the entry still exists.

### DIFF
--- a/libp2p/pubsub/mcache.py
+++ b/libp2p/pubsub/mcache.py
@@ -96,7 +96,8 @@ class MessageCache:
         last_entries: List[CacheEntry] = self.history[len(self.history) - 1]
 
         for entry in last_entries:
-            del self.msgs[entry.mid]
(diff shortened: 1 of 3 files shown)


libp2p/pubsub/mcache.py
```diff
@@ -96,7 +96,8 @@ def shift(self) -> None:
         last_entries: List[CacheEntry] = self.history[len(self.history) - 1]
 
         for entry in last_entries:
-            del self.msgs[entry.mid]
+            if entry.mid in self.msgs:
```
pipermerriam: It's a minimal improvement but you can also do `self.msgs.pop(encry.mid, None)`.  Slightly more robust as it's also threadsafe though admittedly that probably doesn't matter much here.

PR #357, opened 2019-11-20
Replace (check and) del pattern with pop method

NIC619:
Replace
```python
if A in B:
    del B[A]

# or simply
del B[A]
```
with
```python
B.pop(A, None)
```
to prevent `KeyError`.

This is the improvement brought up in https://github.com/libp2p/py-libp2p/pull/355#discussion_r347147489 .
I do a quick search over the code base and it doesn't seem like much work so I apply them.

### DIFF
--- a/libp2p/network/swarm.py
+++ b/libp2p/network/swarm.py
@@ -248,7 +248,7 @@ class Swarm(INetwork):
         # TODO: Should be changed to close multisple connections,
         #   if we have several connections per peer in the future.
         connection = self.connections[peer_id]
-        # NOTE: `connection.close` will perform `del self.connections[peer_id]`
+        # NOTE: `connection.close` will delete `peer_id` from `self.connections`
         # and `notify_disconnected` for us.
         await connection.close()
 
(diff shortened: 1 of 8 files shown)


libp2p/network/swarm.py
```diff
@@ -270,11 +270,9 @@ def remove_conn(self, swarm_conn: SwarmConn) -> None:
         """Simply remove the connection from Swarm's records, without closing
         the connection."""
         peer_id = swarm_conn.muxed_conn.peer_id
-        if peer_id not in self.connections:
-            return
         # TODO: Should be changed to remove the exact connection,
         #   if we have several connections per peer in the future.
-        del self.connections[peer_id]
+        self.connections.pop(peer_id, None)
```
pipermerriam: I would advocate to **only** use this in situations where you explicitely want graceful failure handling.  Otherwise this might hide bugs.  Fox example, this previously had no error handling around it so now this would silently pass when previously it would have raised an error which might be a good thing depending on the code path.

NIC619:
> ... only applying this to places where error handling is already in place.

@pipermerriam 
Not sure I follow, do you mean error handling for `KeyError`? or other errors?
And what would be the purpose of the error handling? Because IIUC `B.pop(A, None)` wouldn't raise any error?

NIC619:
Ahh, by error handling do you mean
```python
if A in B:
    del B[A]
```
?
We know this could raise error so we add a check. So we can replace this pattern with `B.pop(A, None)` but in other case we could suppress errors that we did not foresee.

NIC619:
@pipermerriam Updated the PR so that only the ones with error handling in place are replaced with `pop`.

PR #359, opened 2019-11-22
Update `protobuf` dep to `3.10.0`

### LINKED ISSUE #354: Fixed versions conflicts
In setup.py there are fixed versions :
 "protobuf==3.9.0",
        "coincurve>=10.0.0,<11.0.0",
        "fastecdsa==1.7.4",
        "pynacl==1.3.0",

Those conflicts with other projects I use, for example web3.py that requires protobuf >= 3.10.0

ralexstokes:
Fixes #354.

I re-ran the `protobuf` compiler and it yield no change so I think we can just update the version here to avoid conflicts w/ other packages in our ecosystem.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,7 +64,7 @@ setuptools.setup(
         "multiaddr>=0.0.8,<0.1.0",
         "rpcudp>=3.0.0,<4.0.0",
         "lru-dict>=1.1.6",
-        "protobuf==3.9.0",
+        "protobuf>=3.10.0,<4.0.0",
         "coincurve>=10.0.0,<11.0.0",
         "fastecdsa==1.7.4",
         "pynacl==1.3.0",


mhchia:
Would we want to pin `protobuf>=3.10.0,<4.0.0`, in case the other dependencies updating their required version of `protobuf` someday? This also allows us to be aware that we can make use of newer versions of `protobuf<4.0.0` when doing `pip install` IIUC?

ralexstokes:
i'm not opposed but given the sometimes chaotic nature of python deps, i'd lean towards just hardcoding a particular version we like and then managing updates as required.

i'm open to your suggestion though so just comment back if you want to move forward and i'll update this PR :)

carver:
+1 on the 3.* string that @mhchia gave.

Unless we have a specific reason to distrust that `protobuf` uses semantic versioning, I think loosening the constraint will help reduce the number of versioning issues we see. (like the one mentioned in gitter that opened the issue in the first place).

I feel differently about end-products, when they can justify their own isolated virtualenv, and think the "pin everything" approach can work pretty nicely there. But for libraries, I think the loose semver approach causes the least problems.

ralexstokes:
ok @mhchia @carver yall convinced me :D

carver:
Maybe now is a good time to do a tiny patch release with this fix for folks who are waiting on it. Any reason to wait for something else to be merged?

mhchia:
@carver +1 for having a release for this patch. Thank you. 🙏

PR #360, opened 2019-11-25
Store our keypair in peerstore

NIC619:
Store our public key/private key in peer store so we can use the key to sign `Pubsub` message.
ref in go-libp2p:
- store key pair in peer store during `NewNode`
    - https://github.com/libp2p/go-libp2p/blob/a091bcb087a87b16f9e37a9d248ed6ac2bb2cd66/config/config.go#L109-L110
- get the private key from peer store when initializing `Pubsub`
    - https://github.com/libp2p/go-libp2p-pubsub/blob/01b9825fbee1848751d90a8469e3f5f43bac8466/pubsub.go#L186
- storing in peer store(instead of e.g., like `Host` or `Network`) so it can be accessed from multiple places 
    - https://discuss.libp2p.io/t/what-is-the-purpose-of-having-map-peer-id-privatekey-in-peerstore/74

### DIFF
--- a/libp2p/__init__.py
+++ b/libp2p/__init__.py
@@ -105,6 +105,10 @@ def initialize_default_swarm(
     )
 
     peerstore = peerstore_opt or PeerStore()
+    # Store our key pair in peerstore
+    peerstore.add_pubkey(id_opt, key_pair.public_key)
+    peerstore.add_privkey(id_opt, key_pair.private_key)
+
     # TODO: Initialize discovery if not presented
     return Swarm(id_opt, peerstore, upgrader, transport)
 
@@ -151,9 +155,9 @@ async def new_node(
     # TODO routing unimplemented
     host: IHost  # If not explicitly typed, MyPy raises error
     if disc_opt:
-        host = RoutedHost(key_pair.public_key, swarm_opt, disc_opt)
+        host = RoutedHost(swarm_opt, disc_opt)
     else:
-        host = BasicHost(key_pair.public_key, swarm_opt)
+        host = BasicHost(swarm_opt)
 
     # Kick off cleanup job
     asyncio.ensure_future(cleanup_done_tasks())
--- a/libp2p/__init__.py
+++ b/libp2p/__init__.py
@@ -106,8 +106,7 @@ def initialize_default_swarm(
 
     peerstore = peerstore_opt or PeerStore()
     # Store our key pair in peerstore
-    peerstore.add_pubkey(id_opt, key_pair.public_key)
-    peerstore.add_privkey(id_opt, key_pair.private_key)
+    peerstore.add_key_pair(id_opt, key_pair)
 
     # TODO: Initialize discovery if not presented
     return Swarm(id_opt, peerstore, upgrader, transport)
--- a/libp2p/peer/addrbook_interface.py
+++ b/libp2p/peer/addrbook_interface.py
@@ -7,9 +7,6 @@ from .id import ID
 
 
 class IAddrBook(ABC):
-    def __init__(self) -> None:
-        pass
-
     @abstractmethod
     def add_addr(self, peer_id: ID, addr: Multiaddr, ttl: int) -> None:
         """
--- a/libp2p/peer/peermetadata_interface.py
+++ b/libp2p/peer/peermetadata_interface.py
@@ -5,9 +5,6 @@ from .id import ID
 
 
 class IPeerMetadata(ABC):
-    def __init__(self) -> None:
-        pass
-
     @abstractmethod
     def get(self, peer_id: ID, key: str) -> Any:
         """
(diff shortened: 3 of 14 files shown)


libp2p/__init__.py
```diff
@@ -105,6 +105,10 @@ def initialize_default_swarm(
     )
 
     peerstore = peerstore_opt or PeerStore()
+    # Store our key pair in peerstore
+    peerstore.add_pubkey(id_opt, key_pair.public_key)
+    peerstore.add_privkey(id_opt, key_pair.private_key)
```
ralexstokes: not blocking but if you like this here you go:

consider adding an `add_key_pair` method that takes a `key_pair` and adds both the pubkey and privkey internally? generally easier to get one call correct, rather than have to remember to call two methods when you need to do this.

libp2p/peer/peermetadata_interface.py
```diff
@@ -5,9 +5,6 @@
 
 
 class IPeerMetadata(ABC):
-    def __init__(self) -> None:
-        pass
-
```
ralexstokes: yay cleanup!

libp2p/peer/addrbook_interface.py
```diff
@@ -7,9 +7,6 @@
 
 
 class IAddrBook(ABC):
-    def __init__(self) -> None:
-        pass
-
```
ralexstokes: yay cleanup!

libp2p/peer/peerstore.py
```diff
@@ -10,90 +12,196 @@
 
 class PeerStore(IPeerStore):
 
-    peer_map: Dict[ID, PeerData]
+    peer_pubkey_map: Dict[ID, PublicKey]
+    peer_privkey_map: Dict[ID, PrivateKey]
+    peer_data_map: Dict[ID, PeerData]
```
ralexstokes: we are probably following the Go implementation on this but does it make sense to attach the pubkey/privkey to the `PeerData` directly?

so then you just have one map from `ID` to `PeerData` and then those fields are (optionally?) sitting on the instance of `PeerData`.

the general thinking here is the fewer places to synchronize the better.

libp2p/peer/peerstore.py
```diff
@@ -10,90 +12,196 @@
...
         """
-        if peer_id in self.peer_map:
-            return self.peer_map[peer_id]
+        if peer_id in self.peer_data_map:
+            return self.peer_data_map[peer_id]
         data = PeerData()
-        self.peer_map[peer_id] = data
-        return self.peer_map[peer_id]
+        self.peer_data_map[peer_id] = data
+        return self.peer_data_map[peer_id]
```
ralexstokes: this is super nit-picky but i expect the bytecode is marginally more efficient if we just return `data` rather than re-accessing the map. i would actually hope a "sufficiently smart compiler" would just do this for us but i don't know well enough to say

libp2p/peer/peerstore.py
```diff
@@ -10,90 +12,196 @@
...
         raise PeerStoreError("peer ID not found")
 
     def put(self, peer_id: ID, key: str, val: Any) -> None:
+        """
+        :param peer_id: peer ID to put peer data for
+        :param key:
+        :param value:
+        """
         # <<?>>
         # This can output an error, not sure what the possible errors are
```
ralexstokes: anything to resolve here before merge?

libp2p/peer/peerstore.py
```diff
@@ -10,90 +12,196 @@
...
+        """
         self.add_addrs(peer_id, [addr], ttl)
 
     def add_addrs(self, peer_id: ID, addrs: Sequence[Multiaddr], ttl: int) -> None:
+        """
+        :param peer_id: peer ID to add address for
+        :param addrs:
+        :param ttl: time-to-live for the this record
+        """
         # Ignore ttl for now
```
ralexstokes: would be handy to make an issue for this just so its recorded somewhere more accessible than deep in this code

libp2p/peer/peerstore.py
```diff
@@ -10,90 +12,196 @@
...
+        """
+        if peer_id in self.peer_pubkey_map:
+            return self.peer_privkey_map[peer_id]
+        raise PeerStoreError("peer ID not found")
+
+    def peers_with_keys(self) -> Set[ID]:
+        """
+        :return: all of the peer IDs which has pubkey/privkey stored in peer store
+        """
+        return set(self.peer_pubkey_map.keys()).union(self.peer_privkey_map.keys())
```
ralexstokes: i'm curious how this is used... assuming its something in the Go impl...

NIC619: It's not actually used in the Go implementation.🤔 Removed it.

libp2p/peer/peerstore.py
NIC619: Seems like an outdated comment and `__create_or_get_peer` is replaced by `defaultdict` so it should be safe now.

libp2p/peer/peerstore.py
NIC619: Good idea!

PR #365, opened 2019-11-26
Porting to trio

mhchia:
Port the current code to use trio. Also, make use of `async-service` to manage spawn tasks.

This PR is divided into 7 smaller PRs:
- Transport: #396 
- Network: #397 
- Host related: #398 
- Multiplexer and security stream: #399 
- Pubsub: #400 
- Interop test: #401 
- Settings and helpers: #402


libp2p/io/trio.py
```diff
@@ -0,0 +1,33 @@
...
+    stream: SocketStream
+
+    def __init__(self, stream: SocketStream) -> None:
+        self.stream = stream
+
+    async def write(self, data: bytes) -> None:
+        """Raise `RawConnError` if the underlying connection breaks."""
+        try:
+            await self.stream.send_all(data)
+        except (trio.ClosedResourceError, trio.BrokenResourceError) as error:
```
pipermerriam: Looks like this can also raise `trio.BusyResourceError`

libp2p/io/trio.py
```diff
@@ -0,0 +1,33 @@
...
+import trio
+from trio import SocketStream
+
+from libp2p.io.abc import ReadWriteCloser
+from libp2p.io.exceptions import IOException
+
+logger = logging.getLogger("libp2p.io.trio")
+
+
+class TrioReadWriteCloser(ReadWriteCloser):
```
pipermerriam: It *might* be better to do the work to just use the `trio.abc.Stream` API as it seems to encapsulate what you are trying to do here.

https://trio.readthedocs.io/en/latest/reference-io.html#abstract-base-classes

libp2p/io/trio.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def __init__(self, stream: SocketStream) -> None:
+        self.stream = stream
+
+    async def write(self, data: bytes) -> None:
+        """Raise `RawConnError` if the underlying connection breaks."""
+        try:
+            await self.stream.send_all(data)
+        except (trio.ClosedResourceError, trio.BrokenResourceError) as error:
+            raise IOException(error)
```
pipermerriam: If you do this as `raise IOException from error` then you'll preserve the stacktrace information of the original exception which is typically beneficial.

libp2p/network/connection/swarm_connection.py
```diff
@@ -45,16 +43,11 @@ def __init__(self, muxed_conn: IMuxedConn, swarm: "Swarm") -> None:
...
+        await trio.sleep(0.1)
 
-        for task in self._tasks:
-            task.cancel()
-            try:
-                await task
-            except asyncio.CancelledError:
-                pass
+        # FIXME: Now let `_notify_disconnected` finish first.
         # Schedule `self._notify_disconnected` to make it execute after `close` is finished.
```
pipermerriam: This comment looks out of date.

libp2p/network/swarm.py
```diff
@@ -208,15 +204,20 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
                 await self.add_conn(muxed_conn)
 
                 logger.debug("successfully opened connection to peer %s", peer_id)
+                # FIXME: This is a intentional barrier to prevent from the handler exiting and
+                #   closing the connection.
+                event = trio.Event()
+                await event.wait()
```
pipermerriam: This seems like you could just use `trio.wait_forever()`

libp2p/utils.py
```diff
@@ -95,3 +98,25 @@ def encode_fixedint_prefixed(msg_bytes: bytes) -> bytes:
...
+
+class IQueue(Generic[TItem]):
+    async def put(self, item: TItem):
+        ...
+
+    async def get(self) -> TItem:
+        ...
+
+
+class TrioQueue(IQueue):
```
pipermerriam: I'm inclined to :-1: this approach.

https://trio.readthedocs.io/en/latest/reference-io.html#trio.abc.Channel

First, the `trio.abc.Channel` API is this so I'd advocate using that API instead or at least that name.  You *may* have to actually create that class because I haven't looked to see if `trio` provides an implementation.

mhchia: Big thanks for the review! `TrioQueue` is used temporary just to make the code running. I found `trio.abc.Channel` is extremely helpful for us to replace the [`event.wait` patterns in our original implementation](https://github.com/libp2p/py-libp2p/commit/b6627d1ca32d13ad27fc96d723b5f0ab46eb6228#diff-d01e8ca42064b756a3da55aa43f9aff3L61-L103).

mhchia:
@pipermerriam Thanks for the review! Sorry, I'm still working on fixing the bugs and cleaning up. Therefore, many thing s are still workarounds. I will make them more appropriate for review soon.

One question: does the pattern using `async-service` look good to you? Really appreciate that the `async-service` is extracted to a standalone library and usable for us.

libp2p/network/connection/swarm_connection.py
```diff
@@ -15,21 +17,17 @@
...
     def __init__(self, muxed_conn: IMuxedConn, swarm: "Swarm") -> None:
         self.muxed_conn = muxed_conn
         self.swarm = swarm
         self.streams = set()
-        self.event_closed = asyncio.Event()
-
-        self._tasks = []
+        self.event_closed = trio.Event()
 
     async def close(self) -> None:
```
pipermerriam: This method looks out of place for a `Service`.  I would have at least expected `close()` to trigger `self.manager.cancel()`...

libp2p/network/connection/swarm_connection.py
```diff
@@ -45,16 +43,11 @@ def __init__(self, muxed_conn: IMuxedConn, swarm: "Swarm") -> None:
...
-                await task
-            except asyncio.CancelledError:
-                pass
+        # FIXME: Now let `_notify_disconnected` finish first.
         # Schedule `self._notify_disconnected` to make it execute after `close` is finished.
-        self._notify_disconnected()
+        await self._notify_disconnected()
 
     async def _handle_new_streams(self) -> None:
         while True:
```
pipermerriam: This should probably `while self.manager.is_running`

libp2p/network/swarm.py
```diff
@@ -231,15 +232,16 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
...
-
+        # await asyncio.gather(
+        #     *[listener.close() for listener in self.listeners.values()]
+        # )
+
+        # # Close connections
+        # await asyncio.gather(
+        #     *[connection.close() for connection in self.connections.values()]
+        # )
+        self.manager.stop()
```
pipermerriam: I believe that this should be awaited:

https://async-service.readthedocs.io/en/latest/async_service.html#managers

libp2p/io/trio.py
```diff
@@ -0,0 +1,37 @@
...
+
+    async def read(self, n: int = -1) -> bytes:
+        if n == 0:
+            # Check point
+            await trio.sleep(0)
```
pipermerriam: I think that the more correct way to do this is with `trio.hazmat.checkpoint()`

libp2p/network/connection/raw_connection.py
```diff
@@ -1,42 +1,26 @@
...
         self.is_initiator = initiator
 
-        self._drain_lock = asyncio.Lock()
-
     async def write(self, data: bytes) -> None:
         """Raise `RawConnError` if the underlying connection breaks."""
         try:
-            self.writer.write(data)
-        except ConnectionResetError as error:
+            await self.read_write_closer.write(data)
```
pipermerriam: maybe rather than `read_write_closer` this would be better just called `stream`?

libp2p/network/connection/swarm_connection.py
```diff
@@ -79,28 +78,26 @@ def __init__(self, muxed_conn: IMuxedConn, swarm: "Swarm") -> None:
...
-
-    async def start(self) -> None:
-        await self.run_task(self._handle_new_streams())
+    async def _notify_disconnected(self) -> None:
+        await self.swarm.notify_disconnected(self)
 
-    async def run_task(self, coro: Awaitable[Any]) -> None:
-        self._tasks.append(asyncio.ensure_future(coro))
+    async def run(self) -> None:
+        self.manager.run_task(self._handle_new_streams)
```
pipermerriam: I *think* that you could just replace the whole body of this function with `await self._handle_new_streams()`

libp2p/network/swarm.py
```diff
@@ -208,15 +203,19 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
                 await self.add_conn(muxed_conn)
 
                 logger.debug("successfully opened connection to peer %s", peer_id)
+                # FIXME: This is a intentional barrier to prevent from the handler exiting and
+                #   closing the connection. Probably change to `Service.manager.wait_finished`?
```
pipermerriam: Yes, I think that's a better pattern to use `self.manager.wait_finished()`

libp2p/network/swarm.py
```diff
@@ -229,17 +228,8 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
     async def close(self) -> None:
         # TODO: Prevent from new listeners and conns being added.
```
pipermerriam: I mentioned this in my other review but I think the way this can be accomplished is to implement an equivalent `external_api` decorator for the trio-based service and wrap all of the external methods of this class in it so that when called after closing they will fail.

libp2p/pubsub/floodsub.py
```diff
@@ -61,6 +63,8 @@ def remove_peer(self, peer_id: ID) -> None:
 
         :param rpc: rpc message
         """
+        # Checkpoint
+        await trio.sleep(0)
```
pipermerriam: Same recommendation to use `trio.hazmat.checkpoint`

libp2p/pubsub/pubsub.py
```diff
@@ -361,13 +368,16 @@ def handle_subscription(
...
                 # so add message to the subscription output queue
                 # for each topic
-                await self.my_topics[topic].put(publish_message)
+                await self.subscribed_topics_send[topic].send(publish_message)
 
-    async def subscribe(self, topic_id: str) -> "asyncio.Queue[rpc_pb2.Message]":
+    # TODO: Change to return an `AsyncIterable` to be I/O-agnostic?
+    async def subscribe(
+        self, topic_id: str
+    ) -> "trio.MemoryReceiveChannel[rpc_pb2.Message]":
```
pipermerriam: I'm thinking something like this.

```python
class SubscriptionAPI(AsyncContextManager[None], AsyncIterable[???]):
    async def __aenter__(self) -> 'SubscriptionAPI':
        ...
    async def __aexit__(self, ...):
        await self.cancel()
    async def cancel(self):
        # trigger call to `unsubscribe`
    async def __aiter__(self, ...):
        # yield results (from the trio channel)
```

This would be used like this:

```python
subscription = pubsub.subscribe(my_topic)
async for thing in subscription:
    ... 
await subscription.cancel()

# or better
async with subscription:
    async for thing in subscription:
        ...
```

This abstracts away the channel from the API.

libp2p/pubsub/pubsub.py
```diff
@@ -488,9 +504,20 @@ def handle_subscription(
 
         # TODO: Implement throttle on async validators
 
-        if len(async_topic_validator_futures) > 0:
-            results = await asyncio.gather(*async_topic_validator_futures)
-            if not all(results):
+        if len(async_topic_validators) > 0:
+            # TODO: Use a better pattern
```
pipermerriam: I think this is what you're looking for:

https://github.com/eabrouwer3/aiotoolz/blob/10790c9c5a8413502d8f35ce157966290492dbab/aiotoolz/functoolz.py#L536

I'd just vendor a copy of that function somewhere here in the codebase since that library looks dubiously maintained.

mhchia: Thanks for the direction. `pipe` looks nice to me for its clear syntax, but I also want to schedule all async validators at once and execute them *concurrently*. Does it seem that something like `gather` is quite suitable this case?

libp2p/network/connection/swarm_connection.py
mhchia: Sorry, I missed these feedback previously. Good point. Will address it in the next commits.

libp2p/network/connection/swarm_connection.py
mhchia: Good catch!

mhchia:
@pipermerriam Big thanks for giving the feedback. I've addressed some of them while haven't addressed the others. I will fix the rest soon. :pray:

mhchia:
Fixed all of the tests except for the ones in `kademlia` module. Gotta remove tests for `kademlia` by rebasing to `master`. The next steps is to clean up, reorganize the code, and apply the rest of the PR feedback.

mhchia:
@pipermerriam Should I split this PR to several small pieces by modules? I'm not sure what's the better condition for you to review. I'm still fixing `tests_interop` by porting `py-libp2p-daemon-bindings` to `anyio`, to make it I/O-agnostic. With this changing, I found `anyio` is quite promising, which can possibly be used in our repo in the future.

ShadowJonathan:
(+1 on anyio, I mention it in #383)

ShadowJonathan:
[contextlib.AsyncExitStack is only supported after version 3.7](https://docs.python.org/3/library/contextlib.html#contextlib.AsyncExitStack), this is why `py36-test` fails.

ShadowJonathan:
@mhchia [contextlib.AsyncExitStack is only supported after version 3.7](https://docs.python.org/3/library/contextlib.html#contextlib.AsyncExitStack), this is why `py36-test` fails.

ShadowJonathan:
I also suggest adding `trio_asyncio` to `known_third_parties` in the `isort` tab in `tox.ini`

mhchia:
@ShadowJonathan Thanks for the feedback. I'm fixing it.

mhchia:
@pipermerriam For the error in https://travis-ci.com/mhchia/py-libp2p/jobs/270085872, it seems my tests failed to pass after upgraded to `async-service>=0.1.0a3`, which has `iter_dag`. I'm still investigating the reason, but it will be great to have your input on this issue. :pray: Thank you.

mhchia:
@pipermerriam 

I've fixed the issue in https://travis-ci.com/mhchia/py-libp2p/jobs/270085872 and upgrade to `async-service>=0.1.0a4`. The issue is because the daemon tasks run by `run` are scheduled but not actually run before the tests finish. When the tests finish and cancel the service, `iter_dag` fails. 

Also, this PR is divided into smaller modules to make review easier. It will be great to have your review again. Thank you:).

libp2p/pubsub/pubsub.py
```diff
@@ -417,10 +438,14 @@ def handle_subscription(
...
+        if topic_id not in self.topic_ids:
             return
-        # Remove topic_id from map if present
-        del self.my_topics[topic_id]
+        # Remove topic_id from the maps before yielding
+        send_channel = self.subscribed_topics_send[topic_id]
+        del self.subscribed_topics_send[topic_id]
+        del self.subscribed_topics_receive[topic_id]
+        # Only close the send side
+        await send_channel.aclose()
```
pipermerriam: My intuition (which very well might be wrong) is that this should happen before removing the channels from the maps..

libp2p/pubsub/pubsub_notifee.py
```diff
@@ -56,7 +69,17 @@ def __init__(
...
+            try:
+                await self.dead_peers_queue.send(conn.muxed_conn.peer_id)
+            except (
+                trio.BrokenResourceError,
+                trio.ClosedResourceError,
+                trio.BusyResourceError,
+            ):
+                # Raised when the receive channel is closed.
+                # TODO: Do something with loggers?
+                ...
```
pipermerriam: probably want to get some form of logging in here before merging so this doesn't silently fail somehow....

libp2p/pubsub/pubsub.py
```diff
@@ -417,10 +438,14 @@ def handle_subscription(
...
         # Return if we already unsubscribed from the topic
-        if topic_id not in self.my_topics:
+        if topic_id not in self.topic_ids:
             return
-        # Remove topic_id from map if present
-        del self.my_topics[topic_id]
+        # Remove topic_id from the maps before yielding
+        send_channel = self.subscribed_topics_send[topic_id]
+        del self.subscribed_topics_send[topic_id]
+        del self.subscribed_topics_receive[topic_id]
```
pipermerriam: I think it would be ideal for the `SubscriptionAPI` to be in control of doing all of this cleanup in the `__aexit__`.  The way I've done this elsewhere is to have the subscription API work something like this.

```python
class Subscription:
    def __init__(self, unsubscribe_fn, ...)
        self._unsubscribe_fn = unsubscribe_fn

    def unsubscribe(self):
        self._unsubscribe_fn()

    def __aexit__(self, ...):
        self.unsubscribe()
```

Then in your function here where you create the subscription you do this.

```python
def subscribe(self, ...) -> Subscription:
    subscription = Subscription(
        unsubscribe_fn=functools.partial(self,unsubscribe, topic_id)
        ...
    )
    ...
    return subscription
```

And this inturn allows for usage like this.

```python
async with pubsub.subscribe(topic_id) as subscription:
    ...  # active here
# will be cleaned up and removed once the context exits
```

libp2p/pubsub/pubsub_notifee.py
```diff
@@ -29,7 +30,9 @@ def __init__(
         can process dead peers after we disconnect from each other
         """
         self.initiator_peers_queue = initiator_peers_queue
+        self.initiator_peers_queue_lock = trio.Lock()
         self.dead_peers_queue = dead_peers_queue
+        self.dead_peers_queue_lock = trio.Lock()
```
pipermerriam: I'm confused why these locks are needed...  don't the channels handle this for us?  Maybe deserves a code comment explaining why the lock is necessary.

libp2p/transport/tcp/tcp.py
```diff
@@ -1,64 +1,72 @@
...
         """
         put listener in listening mode and wait for incoming connections.
 
         :param maddr: maddr of peer
         :return: return True if successful
         """
-        self.server = await asyncio.start_server(
-            self.handler,
+
+        async def serve_tcp(
```
pipermerriam: What is the reason for this wrapper function?  Is it just for the logging statement?

mhchia: Yes, it's just for logging.

mhchia:
@pipermerriam 
Thanks for the review and the suggestions. I'm sorry for letting this PR grow this big.

libp2p/pubsub/pubsub_notifee.py
mhchia: Please correct me if I'm wrong: my thought is that `PubsubNotifee.connected` is possibly called multiple times by `Swarm` whenever there are new connections and it's possible for `self.initiator_peers_queue.send` being called concurrently and `trio.BusyResourceError` is raised.

libp2p/io/trio.py
```diff
@@ -0,0 +1,55 @@
...
+                await trio.hazmat.checkpoint()
+                return b""
+            max_bytes = n if n != -1 else None
+            try:
+                return await self.stream.receive_some(max_bytes)
+            except (trio.ClosedResourceError, trio.BrokenResourceError) as error:
+                raise IOException from error
+            except trio.BusyResourceError as error:
+                # This should never happen, since we already access streams with read/write locks.
+                raise Exception(
```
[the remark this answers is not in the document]
mhchia: Directly raising `trio.BusyResourceError` is more reasonable. Thank you for pointing it out.

libp2p/network/network_interface.py
```diff
@@ -70,3 +71,7 @@ def register_notifee(self, notifee: "INotifee") -> None:
     @abstractmethod
     async def close_peer(self, peer_id: ID) -> None:
         pass
+
+
+class INetworkService(INetwork, Service):
```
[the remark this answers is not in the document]
mhchia: Nice catch. I've changed it.

libp2p/pubsub/pubsub.py
mhchia: This is really nice! I will change it in the next commits.

libp2p/pubsub/pubsub_notifee.py
mhchia: Update: I removed the lock and changed the channel to be of size `math.inf`, to avoid `channel.send` from being blocked and prevent `BusyResourceError` from happening.

libp2p/pubsub/pubsub_notifee.py
mhchia: Update: here I only catch `trio.BrokenResourceError`, which occurs when `Pubsub` exits and the `receive_channel` is closed.

mhchia:
@pipermerriam 

Update: I fixed several things according to your suggestions. Thanks a lot for your review. CI started to fail in [`py37-test`](https://travis-ci.com/libp2p/py-libp2p/jobs/280527779) and [`py37-interop`](https://travis-ci.com/libp2p/py-libp2p/jobs/280527782), but still succeeds in [`py36-test`](https://travis-ci.com/libp2p/py-libp2p/jobs/280527778). I spent a whole day debugging but in vain. So far I only know this issue occurs in `3.7.6+` but not `3.7.4+`. Do you have any idea about this?

mhchia:
@ChihChengLiang 
Thanks a lot for the review. I reflected your PR suggestions in https://github.com/libp2p/py-libp2p/pull/396, https://github.com/libp2p/py-libp2p/pull/397, https://github.com/libp2p/py-libp2p/pull/398, and https://github.com/libp2p/py-libp2p/pull/402 in this PR: https://github.com/libp2p/py-libp2p/pull/365/commits/e57d01f360de9fb101adffa686cf5aff414fc8f8.

libp2p/pubsub/pubsub_notifee.py
pipermerriam: I don't think using `math.inf` for channel size is a good idea.  Channels are sized to 1) ensure that they cannot grow to some indefinite size and 2) to provide [backpressure](https://www.techopedia.com/definition/24131/backpressure) when the send side of the channel is operating faster than the receive side can consume.

Setting to `math.inf` effectively disables both of these mechanisms which doesn't seem like a good trade-off.

pipermerriam:
Let me know if you want me to give this another pass :eyes:

libp2p/pubsub/pubsub_notifee.py
mhchia: Alright, I agree with your point that`math.inf` is not suitable for `initiator_peers_queue` and `dead_peers_queue` here since the handle loop should be capable of catching up with the speed of peer churns. It seems to me that to make both sides(send and receive) synchronous I will need to add the locks back to prevent `BusyResourceError` from happening. Do you think it makes sense to add the locks back?

mhchia:
@pipermerriam It will be great to have your second review, though I haven't resolved the failed CI. Thank you!

libp2p/pubsub/pubsub_notifee.py
pipermerriam: According to the docs: 

https://trio.readthedocs.io/en/latest/reference-core.html#trio.MemorySendChannel

`MemorySendChannel` doesn't need a lock and allows multiple tasks to call `send` at the same time so I think you're safe without the lock.

pipermerriam:
I'd like to wait for tests to be green so ping me when that's the case (or when you at least know that whatever is needed to fix the tests isn't going to change much of the code).

libp2p/pubsub/pubsub_notifee.py
mhchia: My fault. I'm changing it back. Thank you for the reference.

mhchia:
@pipermerriam 

The reason why CIs using `3.7-dev`([py37-test](https://travis-ci.com/libp2p/py-libp2p/jobs/282731792) and [py37-interop](https://travis-ci.com/libp2p/py-libp2p/jobs/282731795)) failed is because of https://github.com/python/cpython/commit/b76d5e9ee64b8ac861e856d8e41289c0977c75e1. The `asynccontextmanager` in the package `async-generator` fails with python `3.7-dev`, but the one in `contextlib` works. 

The current solutions are 
1. Work around the issue: pin the version of python back from `3.7-dev` to `3.7`.
2. Solve the issue in `async-generator`. However, since `trio` also [depends](https://github.com/python-trio/trio/blob/master/setup.py#L84) on `async-generator`. To get the CI fixed, we need to make all the dependencies use the fixed `async-generator`.

My thought is to report this issue to the upstream. Then work around the issue, i.e. to use python `3.7` in CI. I will dig more into the issue and report it to the upstream now. Thoughts?

pipermerriam:
Using regular 3.7 seems fine (and reporting it upstream) seems fine.

mhchia:
Thanks for the feedback! I opened an issue in the upstream: https://github.com/python-trio/async_generator/issues/30.

mhchia:
@pipermerriam I changed to `3.7` and the CI are fixed. It will be great to have your second review. Thank you. :pray:

libp2p/host/ping.py
```diff
@@ -16,22 +17,23 @@
...
-        return False
-    except StreamReset as error:
-        logger.debug(
-            "Other side reset while waiting for ping from %s: %s", peer_id, error
-        )
-        raise
-    except Exception as error:
-        logger.debug("Error while waiting to read ping for %s: %s", peer_id, error)
-        raise
+    with trio.fail_after(RESP_TIMEOUT):
```
pipermerriam: I think this should be moved inside the `try` block.

libp2p/network/connection/swarm_connection.py
```diff
@@ -19,22 +20,27 @@ class SwarmConn(INetConn):
     muxed_conn: IMuxedConn
     swarm: "Swarm"
     streams: Set[NetStream]
-    event_closed: asyncio.Event
-
-    _tasks: List["asyncio.Future[Any]"]
+    event_started: trio.Event
```
pipermerriam: Since this is declared on the parent class it seems un-necessary to specify it here too.

libp2p/network/connection/swarm_connection.py
```diff
@@ -45,43 +51,32 @@ def __init__(self, muxed_conn: IMuxedConn, swarm: "Swarm") -> None:
...
         while True:
             try:
                 stream = await self.muxed_conn.accept_stream()
             except MuxedConnUnavailable:
-                # If there is anything wrong in the MuxedConn,
-                # we should break the loop and close the connection.
                 break
             # Asynchronously handle the accepted stream, to avoid blocking the next stream.
-            await self.run_task(self._handle_muxed_stream(stream))
+            self.swarm.manager.run_task(self._handle_muxed_stream, stream)
```
pipermerriam: If you update to the latest `async-service` release (which you should), this *should* cause `mypy` to complain.  The `run_task` APIs are intentionally not exposed on the public API of `ManagerAPI`.

My recommendation would be **not** to use the `ManagerAPI` but instead to just open a local nursery here in this method and just use the raw nursery API to run this in the backgroung.

libp2p/network/connection/swarm_connection.py
```diff
@@ -45,43 +51,32 @@ def __init__(self, muxed_conn: IMuxedConn, swarm: "Swarm") -> None:
...
-            self.remove_stream(net_stream)
-
     async def _handle_muxed_stream(self, muxed_stream: IMuxedStream) -> None:
         net_stream = self._add_stream(muxed_stream)
         if self.swarm.common_stream_handler is not None:
-            await self.run_task(self._call_stream_handler(net_stream))
+            try:
+                await self.swarm.common_stream_handler(net_stream)
+            # TODO: More exact exceptions
+            except Exception:
```
pipermerriam: Are you intentionally silencing this exception?  This looks like you may want to just do a `try/finally` and allow the exception to bubble.

libp2p/network/swarm.py
```diff
@@ -238,14 +229,19 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
...
+                await self.manager.wait_finished()
+
             try:
                 # Success
                 listener = self.transport.create_listener(conn_handler)
                 self.listeners[str(maddr)] = listener
-                await listener.listen(maddr)
+                # TODO: `listener.listen` is not bounded with nursery. If we want to be
+                #   I/O agnostic, we should change the API.
+                await listener.listen(maddr, self.manager._task_nursery)  # type: ignore
```
pipermerriam: We should not be accessing this private API.  Instead, this class should create a new nursery in the `run` method and that should be used here.

libp2p/network/swarm.py
```diff
@@ -320,19 +301,21 @@ def register_notifee(self, notifee: INotifee) -> None:
...
-        )
+        for notifee in self.notifees:
+            self.manager.run_task(notifee.opened_stream, self, stream)
 
     # TODO: `notify_closed_stream`
 
     def notify_connected(self, conn: INetConn) -> None:
-        asyncio.gather(*[notifee.connected(self, conn) for notifee in self.notifees])
+        for notifee in self.notifees:
+            self.manager.run_task(notifee.connected, self, conn)
```
pipermerriam: It seems odd to me that this method returns before the notifications have been sent (they will only have been queued)...  Is that intentional?

libp2p/pubsub/gossipsub.py
```diff
@@ -75,16 +78,19 @@ def __init__(
...
         self.mcache = MessageCache(gossip_window, gossip_history)
 
         # Create heartbeat timer
         self.heartbeat_initial_delay = heartbeat_initial_delay
         self.heartbeat_interval = heartbeat_interval
 
+    async def run(self) -> None:
+        if self.pubsub is None:
+            raise NoPubsubAttached
+        self.manager.run_task(self.heartbeat)
```
pipermerriam: Should this be a `daemon` task?  Is it considered an error if the heartbeat task exited?

libp2p/pubsub/pubsub_notifee.py
```diff
@@ -56,10 +60,14 @@ def __init__(
...
-        await self.dead_peers_queue.put(conn.muxed_conn.peer_id)
+        try:
+            await self.dead_peers_queue.send(conn.muxed_conn.peer_id)
+        except trio.BrokenResourceError:
+            # The receive channel is closed by Pubsub. We should do nothing here.
+            ...
 
     async def listen(self, network: INetwork, multiaddr: Multiaddr) -> None:
-        pass
+        ...
```
pipermerriam: Should this raise `NotImplementedError` or something?

libp2p/stream_muxer/mplex/mplex.py
```diff
@@ -61,15 +64,16 @@ def __init__(self, secured_conn: ISecureConn, peer_id: ID) -> None:
...
         self.streams = {}
-        self.streams_lock = asyncio.Lock()
-        self.new_stream_queue = asyncio.Queue()
-        self.event_shutting_down = asyncio.Event()
-        self.event_closed = asyncio.Event()
-
-        self._tasks = []
+        self.streams_lock = trio.Lock()
+        self.streams_msg_channels = {}
+        channels = trio.open_memory_channel[IMuxedStream](math.inf)
```
pipermerriam: We shouldn't be using `math.inf` for channel sizes...

libp2p/stream_muxer/mplex/mplex.py
```diff
@@ -104,9 +109,13 @@ def _get_next_channel_id(self) -> int:
         return next_id
 
     async def _initialize_stream(self, stream_id: StreamID, name: str) -> MplexStream:
-        stream = MplexStream(name, stream_id, self)
+        # Use an unbounded buffer, to avoid `handle_incoming` being blocked when doing
+        # `send_channel.send`.
```
pipermerriam: I really don't think this is a good idea.  It will hide a real problem by allowing unbounded memory growth...

libp2p/transport/tcp/tcp.py
pipermerriam: Worth adding a comment saying this explicitly

libp2p/network/connection/swarm_connection.py
mhchia: I intended to catch exceptions raised in the `common_stream_handler` to avoid the whole `Swarm` broken. After second thoughts, it seems not what we should take care of. And it seems I forgot to do `remove_stream` for the success case. Nice catch!

libp2p/network/swarm.py
mhchia: Hmm, you're right. I checked out the implementation in [go](https://github.com/libp2p/go-libp2p-swarm/blob/master/swarm.go#L459). All Notifiees should finish before the function exits.

libp2p/pubsub/pubsub_notifee.py
mhchia: `PubsubNotifee.listen` just do *nothing* here. Since it is an asynchronous function, I'm going to add checkpoints here.

mhchia:
Fixed most of the issue here, except for `math.inf`. Will keep fixing it tmr.

libp2p/pubsub/pubsub_notifee.py
pipermerriam: Lets reserve `...` for `@abstractmethod` bodies.

I think anytime something is intentionally doing nothing it's probably good to leave a comment so the intention is clear.

libp2p/stream_muxer/mplex/mplex.py
mhchia: Do you think it makes sense to use a bounded buffer for `streams_msg_channels`, i.e. `trio.open_memory_channel[bytes](BUFFER_SIZE)`? Or do you expect all the channels to be unbuffered, i.e. `trio.open_memory_channel[bytes](0)`?

libp2p/pubsub/pubsub_notifee.py
mhchia: Sounds reasonable to me. Will change it.

libp2p/stream_muxer/mplex/mplex.py
pipermerriam: Bounded to some reasonable upper bound seems fine.  I don't have context to know what an appropriate value would be here but it's easy to change so no need to stress over it too much.

mhchia: I've changed the sizes of
- `Mplex.new_stream_send_channel` to `0`. Its consumer is `SwarmConn`, which should be able to consume the data pretty soon in another task.
https://github.com/libp2p/py-libp2p/blob/5b03a7ad9f801fe7fa26c70515dadbe2e9bf799e/libp2p/stream_muxer/mplex/mplex.py#L70-L71
- `Mplex.streams_msg_channels` to `MPLEX_MESSAGE_CHANNEL_SIZE = 8`, conforming to the go implementation. If the channel is full and new messages arrive, the stream is reset.
https://github.com/libp2p/py-libp2p/blob/5b03a7ad9f801fe7fa26c70515dadbe2e9bf799e/libp2p/stream_muxer/mplex/mplex.py#L113-L115
- `Pubsub.subscribed_topics_send[topic]` to `SUBSCRIPTION_CHANNEL_SIZE = 32`, conforming to the go implementation. If the channel is full and new messages arrive, the messages are just dropped.
https://github.com/libp2p/py-libp2p/blob/ddbedc6c154cff5c513ff149c0948c2394b0e6d8/libp2p/pubsub/pubsub.py#L398-L400

mhchia:
@pipermerriam 
I've removed the `math.inf` in the codebase, except for the ones in the tests. Please have another pass of review if you have time. I will merge the PR tomorrow if there is no further feedback. Really appreciate your review and suggestions!

PR #366, opened 2019-11-26
Splice in project template

carver:
This has the benefit of making future upgrades from the template project pretty painless.

Of interest are the four latest commits: 27da3122850e4...f84edfc13377

- Merge in the template, fixing up lots of conflicts
- Match linting rules used by the local repository
- Fill in template variables (like the readthedocs name, module name, etc)
- A fixup from #353 -- Added a missing script for testing the release package

This supplies the bulk of #294 (maybe all of it), but I won't be confident about completion until running a release using this code.

---

Side note: sorry for the ugly history. It's the only way we have figured to keep template upgrades easy in projects that depend on them. See: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/template.md#repository-template


carver:
Note that master is currently failing on Travis, soo I guess I can/should ignore a failure here? :grimacing: 

https://github.com/libp2p/py-libp2p/runs/321090268

setup.py
```diff
@@ -1,62 +1,62 @@
...
+        "pytest-xdist>=1.30.0,<2",
         "pytest-asyncio>=0.10.0,<1.0.0",
-        "pytest-xdist>=1.30.0",
+        "factory-boy>=2.12.0,<3.0.0",
     ],
     "lint": [
-        "mypy>=0.701,<1.0",
+        "flake8==3.7.9",  # flake8 is not semver: it has added new warnings at minor releases
+        "isort==4.3.21",
+        "mypy==0.740",  # flake8 is not semver: it has added new warnings at minor releases
```
ralexstokes: spurious comment?

carver: Ah thanks, I forgot to reword the comment (roughly the same thing applies to mypy).

PR #372, opened 2019-12-03
Add py36 compatibility

ralexstokes:
## What was wrong?

We only supported python 3.7.

## How was it fixed?

The main thing was adding backfills for APIs that changed in python 3.7 compared to earlier versions.

I had to make some changes to some tests to account for subtleties in the tests that interacted w/ differences in semantics b/t python 3.6 and 3.7.

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,30 +48,24 @@ jobs:
       - image: circleci/python:3.6
         environment:
           TOXENV: lint
-  py36-core:
+  py36-test:
     <<: *common
     docker:
       - image: circleci/python:3.6
         environment:
-          TOXENV: py36-core
-  py37-core:
+          TOXENV: py36-test
+  py37-test:
     <<: *common
     docker:
       - image: circleci/python:3.7
         environment:
-          TOXENV: py37-core
-  pypy3-core:
-    <<: *common
-    docker:
-      - image: pypy
-        environment:
-          TOXENV: pypy3-core
+          TOXENV: py37-test
+
 workflows:
   version: 2
   test:
     jobs:
       - docs
       - lint
-      - py36-core
-      - py37-core
-      - pypy3-core
+      - py36-test
+      - py37-test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,24 +48,30 @@ jobs:
       - image: circleci/python:3.6
         environment:
           TOXENV: lint
-  py36-test:
+  py36-core:
     <<: *common
     docker:
       - image: circleci/python:3.6
         environment:
-          TOXENV: py36-test
-  py37-test:
+          TOXENV: py36-core
+  py37-core:
     <<: *common
     docker:
       - image: circleci/python:3.7
         environment:
-          TOXENV: py37-test
-
+          TOXENV: py37-core
+  pypy3-core:
+    <<: *common
+    docker:
+      - image: pypy
+        environment:
+          TOXENV: pypy3-core
 workflows:
   version: 2
   test:
     jobs:
       - docs
       - lint
-      - py36-test
-      - py37-test
+      - py36-core
+      - py37-core
+      - pypy3-core
--- a/libp2p/network/connection/raw_connection.py
+++ b/libp2p/network/connection/raw_connection.py
@@ -53,5 +53,6 @@ class RawConnection(IRawConnection):
 
     async def close(self) -> None:
         self.writer.close()
-        if sys.version_info[0:2] > (3, 6):
-            await self.writer.wait_closed()
+        if sys.version_info < (3, 7):
+            return
+        await self.writer.wait_closed()
--- a/libp2p/network/connection/raw_connection.py
+++ b/libp2p/network/connection/raw_connection.py
@@ -1,4 +1,5 @@
 import asyncio
+import sys
 
 from .exceptions import RawConnError
 from .raw_connection_interface import IRawConnection
@@ -52,4 +53,5 @@ class RawConnection(IRawConnection):
 
     async def close(self) -> None:
         self.writer.close()
-        await self.writer.wait_closed()
+        if sys.version_info[0:2] > (3, 6):
+            await self.writer.wait_closed()
--- a/libp2p/tools/factories.py
+++ b/libp2p/tools/factories.py
@@ -1,6 +1,8 @@
 import asyncio
 from typing import Any, AsyncIterator, Dict, Tuple, cast
 
+# NOTE: import ``asynccontextmanager`` from ``contextlib`` when support for python 3.6 is dropped.
+from async_generator import asynccontextmanager
 import factory
 
 from libp2p import generate_new_rsa_identity, generate_peer_id_from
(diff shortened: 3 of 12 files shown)


.circleci/config.yml
```diff
@@ -48,30 +48,24 @@ jobs:
       - image: circleci/python:3.6
```
ralexstokes: changed this file but it is actually not being used right now...

libp2p/network/connection/raw_connection.py
```diff
@@ -52,4 +53,5 @@ def __init__(
 
     async def close(self) -> None:
         self.writer.close()
-        await self.writer.wait_closed()
+        if sys.version_info[0:2] > (3, 6):
```
ralexstokes: a bit clunky for the `wait_closed` API only in `3.7+` -- open to suggestions on how to handle this kind of thing...

libp2p/tools/pubsub/floodsub_integration_test_settings.py
```diff
@@ -180,59 +188,45 @@
...
-        for neighbor_id in adj_list[start_node_id]:
-            # Create neighbor if neighbor does not yet exist
-            if neighbor_id not in node_map:
-                await add_node(neighbor_id)
-            tasks_connect.append(
-                connect(node_map[start_node_id], node_map[neighbor_id])
-            )
-    # Connect nodes and wait at least for 2 seconds
-    await asyncio.gather(*tasks_connect, asyncio.sleep(2))
+    # NOTE: the test using this routine will fail w/o these sleeps...
```
ralexstokes: had some trouble w/ this suite of tests under python 3.6.... seems like something changed behind the scenes w/ `gather`.... ultimately tried to streamline the implementation and found that the `sleep`s are important...

given that we are moving towards a `trio` refactor i don't think its worth the resources right now to debug further....

setup.py
```diff
@@ -80,7 +82,7 @@
     url="https://github.com/libp2p/py-libp2p",
     include_package_data=True,
     install_requires=install_requires,
-    python_requires=">=3.7,<4",
+    python_requires=">=3.6,<4",
```
carver: :+1: reminder to add to the `classifiers` below, also.

libp2p/network/connection/raw_connection.py
carver: One feature I aim for is: the ability to simply delete lines when we drop support for an old python version.

Which makes me think of:
```py
if sys.version_info < (3, 7):  # one bonus of this construction is that you don't have to slice the version_info
  return
await self.writer.wait_closed()
```

I don't love early returns, but might personally favor that to get the "delete to drop support" feature.

Some other options:
1. subclass the writer with an empty `wait_closed()` if using python3.6 (should reduce the number of different files touched, which is nice)
2. basically the same as ^ but monkey-patch the method onto the `StreamWriter` instead of subclassing
3. a `hasattr(writer, 'wait_closed')` check first (tiny performance penalty, and doesn't meet the goal of "delete to drop support")

I personally have higher tolerance for "uglier" patch solutions here, since it will be deleted when we drop support for old versions. (Like 1 or 2 above)

Any of these (including status quo) is fine, just offering alternatives that you might like better.

.circleci/config.yml
```diff
@@ -48,30 +48,24 @@ jobs:
...
+
 workflows:
   version: 2
   test:
     jobs:
       - docs
       - lint
-      - py36-core
-      - py37-core
-      - pypy3-core
```
carver: If we can get pypy3 support, it's nice to have this in to prevent accidentally losing support. Since it's not blocking this PR (not running Circle yet), I'm :+1: for leaving it in for now. (especially the test definition, even if it's removed from the `workflows`)

... but if you know pypy3 support is not possible currently, and would take a lot of work, then no objection to dropping it now.

libp2p/tools/pubsub/floodsub_integration_test_settings.py
carver: Maybe add a tracking issue to revisit this after switching to trio?

carver:
I don't have enough context to comment on the changes in the `libp2p/tools` and the sleeps, etc. I'll take your word that it will change significantly soon, and isn't worth debugging now. :)

tests/pubsub/test_pubsub.py
```diff
@@ -58,11 +58,11 @@
     await connect(pubsubs_fsub[0].host, pubsubs_fsub[1].host)
     await pubsubs_fsub[0].subscribe(TESTING_TOPIC)
     # Yield to let 0 notify 1
-    await asyncio.sleep(0.1)
+    await asyncio.sleep(1)
     assert pubsubs_fsub[0].my_id in pubsubs_fsub[1].peer_topics[TESTING_TOPIC]
     await pubsubs_fsub[0].unsubscribe(TESTING_TOPIC)
     # Yield to let 0 notify 1
-    await asyncio.sleep(0.1)
+    await asyncio.sleep(1)
```
mhchia: Does the test fail w/o increasing the sleep time?

tox.ini
```diff
@@ -1,8 +1,9 @@
 # Reference: https://github.com/ethereum/ethereum-python-project-template/blob/master/tox.ini
 
-# TODO: consider py36 and pypy3 support
+# TODO: consider pypy3 support
 [tox]
 envlist =
+    py36-test
```
mhchia: nitpick: it seems `py36-test` and `py37-test` can be combined as `py{36,37}-test`

.circleci/config.yml
carver: Yeah, it's on my to-do list to switch to CircleCI. (It is way faster)

libp2p/tools/factories.py
```diff
@@ -33,6 +32,12 @@
 )
 from .utils import connect, connect_swarm
 
+try:
+    from contextlib import asynccontextmanager
+except ImportError:
+    # NOTE: mypy complains about a duplicate import without the following ``# type: ignore``
+    from async_generator import asynccontextmanager  # type: ignore
```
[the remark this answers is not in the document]
carver: If we do that, then it's worth leaving a note to change it back when we drop py3.6

libp2p/network/connection/raw_connection.py
ralexstokes: yeah i like the early return version actually; the goal of being able to just delete lines to drop support for older versions does seem like a worthwhile thing to aim for!

.circleci/config.yml
ralexstokes: good point, i don't know about `pypy3` support and i'm actually inclined to just revert changes to this file

tests/pubsub/test_pubsub.py
ralexstokes: they do -- which is part of the issue but i opened a tracking issue in #378.

ralexstokes:
@carver @mhchia i've applied your PR feedback. thanks!

carver:
I took care of it for this release, but just a reminder to add a blurb to `newsfragments/` that will be included with the release notes. See some examples here: https://github.com/libp2p/py-libp2p/commit/23fa86979db43102e44c300b75008c6394c49b5f

and how they turned out:
https://py-libp2p.readthedocs.io/en/latest/release_notes.html#libp2p-v0-1-4-2019-12-12

PR #377, opened 2019-12-05
Remove kademlia module

### LINKED ISSUE #374: Remove kademlia module
## What was wrong?

Currently, Kademlia is a blocker for further development. Since it is not used in Eth2.0, we have minimized the maintenance of this module for a while. But it also created a vicious cycle that the less we maintain it, the harder we maintain it.

For example, it has no typing so adding new features could be difficult. For another, it's heavily based on asyncio, adding extra work to port libp2p to Trio.

I would suggest removing Kademlia first, and then rework a new one when the py-libp2p library has been rearchitected to some degree. Desirably when we have a sans-io version of py-libp2p.

### Concerns of removing Kademlia

- Some users might have already been using it.

## How can it be fixed?

Remove `libp2p/kademlia` and `tests/kademlia`

ChihChengLiang:
## What was wrong?

Fix #374,Remove kademlia module

## How was it fixed?

- Remove kademlia and routing/kademlia
- Fix routed_host test

#### Cute Animal Picture


libp2p/tools/utils.py
```diff
@@ -95,3 +76,29 @@
...
+    _routing_table: Dict[ID, PeerInfo]
+
+    def __init__(self) -> None:
+        self._routing_table = dict()
+
+    async def find_peer(self, peer_id: ID) -> PeerInfo:
+        return self._routing_table.get(peer_id, None)
+
+
+async def set_up_routed_hosts() -> Tuple[BasicHost, BasicHost]:
```
mhchia:
```suggestion
async def set_up_routed_hosts() -> Tuple[RoutedHost, RoutedHost]:
```

libp2p/tools/utils.py
```diff
@@ -58,24 +57,6 @@
     return tuple(nodes_list)
```
mhchia: Should the type here also be `RoutedHost`?

libp2p/tools/utils.py
```diff
@@ -95,3 +76,29 @@
     # Associate the peer with local ip address (see default parameters of Libp2p())
     node_a.get_peerstore().add_addrs(node_b.get_id(), node_b.get_addrs(), 10)
     return node_a, node_b
+
+
+class DummyRouter(IPeerRouting):
```
mhchia: I like this nice and simple implementation of `IPeerRouting `:).

libp2p/tools/utils.py
ChihChengLiang: This function is only used once and does not ease any cognitive load. Let me just remove it.

PR #381, opened 2019-12-16
[minor] Prevent extras_require overlap

ShadowJonathan:
## What was wrong?

`extras_require` in `setup.py` was defining `"dev"` as a list of requirements, but then redefinining and squashing it with all other extra requirements later on, with the *same key*.

## How was it fixed?

Rename `"dev"` from the list to `"dev_core"` (while keeping the squashed extras as `"dev"`), keeping everything bug-free.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

(a small cat for a small PR)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ extras_require = {
         "sphinx_rtd_theme>=0.4.3,<=1",
         "towncrier>=19.2.0, <20",
     ],
-    "dev": [
+    "dev_core": [
         "bumpversion>=0.5.3,<1",
         "pytest-watch>=4.1.0,<5",
         "wheel",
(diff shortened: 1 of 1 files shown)


ShadowJonathan:
> Minor, but this ends up leaving a `dev_core` extra requirement that isn't really a useful install target...

Current `[dev]` extras adds *all* extras, a huge amount of dependencies that mostly go to linting and testing.

This is more for efficiency and convenience, the difference between a "normal" installation, a "development" installation, and a "dev suite" installation is that the last one intends to use every aspect of development, while the second one might just need a minimal running environment. (for version bumping or wheel compilation)

By the way, on that note: Shouldn't `tox`+`pytest-watch` be moved to `"test"`, and `wheel`+`setuptools` be moved to `setup_requires`?

ShadowJonathan:
I rebased the branch so it comes from `master`, and doesn't include a PR merge commit

ralexstokes:
@ShadowJonathan do you feel strongly about getting this merged in? otherwise i'll close the PR

PR #385, opened 2019-12-17
fixes #197

### LINKED ISSUE #197: Update digest hash function to sha256
#### What's wrong?
https://github.com/libp2p/py-libp2p/blob/1727ba48d9e5cff4fb00d3972aee0a7cd1871eb0/libp2p/peer/id.py#L87-L90
sha1 is used in calculating a DHT ID
1. XOR ID of a peer
https://github.com/libp2p/py-libp2p/blob/1727ba48d9e5cff4fb00d3972aee0a7cd1871eb0/libp2p/peer/id.py#L37-L38
2. key in DHT
https://github.com/libp2p/py-libp2p/blob/1727ba48d9e5cff4fb00d3972aee0a7cd1871eb0/libp2p/kademlia/network.py#L143

#### How can it be fixed?
Change it from sha1 to sha256
Go:
https://github.com/libp2p/go-libp2p-kbucket/blob/3752ea0128fd84b4fef0a66739b8ca95c8a471b6/util.go#L43-L46

ShadowJonathan:
## What was wrong?

Issue #197

## How was it fixed?

Change peerID digest algorithm to sha256 and clarify function language.

Note: This function should probably be moved to a general-use package like `tools` or something.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/libp2p/peer/id.py
+++ b/libp2p/peer/id.py
@@ -44,7 +44,7 @@ class ID:
     @property
     def xor_id(self) -> int:
         if not self._xor_id:
-            self._xor_id = int(digest(self._bytes).hex(), 16)
+            self._xor_id = int(sha256_digest(self._bytes).hex(), 16)
         return self._xor_id
 
     def to_bytes(self) -> bytes:
@@ -89,7 +89,7 @@ class ID:
         return cls(mh_digest.encode())
 
 
-def digest(data: Union[str, bytes]) -> bytes:
+def sha256_digest(data: Union[str, bytes]) -> bytes:
     if isinstance(data, str):
         data = data.encode("utf8")
-    return hashlib.sha1(data).digest()
+    return hashlib.sha256(data).digest()


libp2p/peer/id.py
```diff
@@ -89,7 +89,7 @@ def from_pubkey(cls, key: PublicKey) -> "ID":
         return cls(mh_digest.encode())
 
 
-def digest(data: Union[str, bytes]) -> bytes:
+def sha256_digest(data: Union[str, bytes]) -> bytes:
```
ralexstokes: i don't think we need to block merging for this but something to consider is the name of this function...

making it `sha256_digest` makes it very specific to producing a SHA-256 hash, which is what we want to do here. however, consider what happens in the future if the kademlia code changes again to use a different hash -- we will need to go and find all of the uses of `sha256_digest` and update them. also in terms of communicating information, the name `sha256_digest` doesn't really add more information than what is already provided by `hashlib.sha256`.

in general, i aim to name things based on what they provide, not what they do. this generally means better maintenance over time and hopefully better communicates intent to myself and others down the line. in this case, i'd think to name it something like `dht_digest` and also probably move it to the (future) `libp2p.kademlia` module. my two cents on naming :)

ralexstokes:
it looks like github merged this in but did not update this PR state. i'll go ahead and close and just note this PR was merged into `master` at `1355fbae99a81dc73e5f431951e8744f603a41eb`

PR #386, opened 2019-12-17
fixes #384

### LINKED ISSUE #384: [bug] Swarm.dial_peer only tries first multiaddress before failing
## What was wrong?

`Swarm.dial_peer()` only tries the first out of the multiaddresses provided, it does not interate over them till a connection is established.

(As a side-effect, the selection uses the `0` key, even though further up there is a dodgy and cryptic `not` check, this is still a KeyError that can be unaccounted for)

### Code that ~~produced~~contains the ~~error~~bug

https://github.com/libp2p/py-libp2p/blob/d503950179c810a881a31f3b273021de137d83d8/libp2p/network/swarm.py#L96

## How can it be fixed?

Iterate over all known multiaddresses, continue to the next one when one attempt fails or returns nothing, and stop when a connection has finally been established.

ShadowJonathan:
also adds MultiError to libp2p/exceptions.py

    and an additional fixme I have noticed

## What was wrong?

Issue #384

## How was it fixed?

Adding an `for .. in` iteration to the multiaddresses, also splitting the function into `dial_peer` and `dial_addr` for convenience and clarity.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/libp2p/exceptions.py
+++ b/libp2p/exceptions.py
@@ -8,3 +8,9 @@ class ValidationError(BaseLibp2pError):
 
 class ParseError(BaseLibp2pError):
     pass
+
+
+class MultiError(BaseLibp2pError):
+    """Raised with multiple exceptions."""
+
+    # todo: find some way for this to fancy-print all encapsulated errors
--- a/libp2p/network/swarm.py
+++ b/libp2p/network/swarm.py
@@ -19,6 +19,7 @@ from libp2p.transport.transport_interface import ITransport
 from libp2p.transport.upgrader import TransportUpgrader
 from libp2p.typing import StreamHandlerFn
 
+from ..exceptions import MultiError
 from .connection.raw_connection import RawConnection
 from .connection.swarm_connection import SwarmConn
 from .exceptions import SwarmException
@@ -90,14 +91,49 @@ class Swarm(INetwork):
         except PeerStoreError:
             raise SwarmException(f"No known addresses to peer {peer_id}")
 
-        if not addrs:
+        if len(addrs) == 0:
             raise SwarmException(f"No known addresses to peer {peer_id}")
 
-        multiaddr = addrs[0]
+        exceptions: List[SwarmException] = []
+
+        # Try all known addresses
+        for multiaddr in addrs:
+            try:
+                return await self.dial_addr(multiaddr, peer_id)
+            except SwarmException as e:
+                exceptions.append(e)
+                logger.debug(
+                    "encountered swarm exception when trying to connect to %s, "
+                    "trying next address...",
+                    multiaddr,
+                    exc_info=e,
+                )
+
+        # Tried all addresses, raising exception.
+        if len(exceptions) > 0:
+            raise SwarmException(
+                "unable to connect to %s, all addresses failed to dial (with exceptions)",
+                peer_id,
+            ) from MultiError(exceptions)
+        else:
+            raise SwarmException(
+                "unable to connect to %s, all addresses failed to dial", peer_id
+            )
+
+    async def dial_addr(self, addr: Multiaddr, peer_id: ID) -> INetConn:
+        """
+        dial_addr try to create a connection to peer_id with addr.
+
+        :param addr: the address we want to connect with
+        :param peer_id: the peer we want to connect to
+        :raises SwarmException: raised when an error occurs
+        :return: muxed connection
+        """
+
         # Dial peer (connection to peer does not yet exist)
         # Transport dials peer (gets back a raw conn)
         try:
-            raw_conn = await self.transport.dial(multiaddr)
+            raw_conn = await self.transport.dial(addr)
         except OpenConnectionError as error:
             logger.debug("fail to dial peer %s over base transport", peer_id)
             raise SwarmException(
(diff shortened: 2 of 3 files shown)


libp2p/exceptions.py
```diff
@@ -8,3 +8,9 @@ class ValidationError(BaseLibp2pError):
 
 class ParseError(BaseLibp2pError):
     pass
+
+
+class MultiError(BaseLibp2pError):
+    """Raised with multiple exceptions."""
+
+    # todo: find some way for this to fancy-print all encapsulated errors
```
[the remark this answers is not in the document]
ShadowJonathan: That's what i'm implying with that comment right there, but I have no idea how to make stack trace printing of these exceptions not botch itself with `__str__`, but instead play nice with existing libraries that extract stack trace information, which I don't know how to do, thus the comment.

libp2p/network/swarm.py
```diff
@@ -90,14 +91,49 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
         except PeerStoreError:
             raise SwarmException(f"No known addresses to peer {peer_id}")
 
-        if not addrs:
+        if len(addrs) == 0:
             raise SwarmException(f"No known addresses to peer {peer_id}")
```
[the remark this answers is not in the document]
ShadowJonathan: It's actually still possible for IPeerStore to return an empty set, though this would not alter behavior much, iteration will not start, and it'll cascade direction to an exception. Though I would have to change wording for that to make sense.

libp2p/network/swarm.py
```diff
@@ -87,17 +88,53 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
...
+            )
+
+    async def dial_addr(self, addr: Multiaddr, peer_id: ID) -> INetConn:
+        """
+        dial_addr try to create a connection to peer_id with addr.
+
+        :param addr: the address we want to connect with
+        :param peer_id: the peer we want to connect to
+        :raises SwarmException: raised when an error occurs
+        :return: muxed connection
```
mhchia:
```suggestion
        :return: network connection
```

libp2p/network/swarm.py
```diff
@@ -87,17 +88,53 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
...
+        # Tried all addresses, raising exception.
+        if len(exceptions) > 0:
+            raise SwarmException(
+                "unable to connect to %s, no addresses established a successful connection "
+                "(with exceptions)",
+                peer_id,
+            ) from MultiError(exceptions)
+        else:
+            raise SwarmException(
+                "unable to connect to %s, no addresses established a successful connection "
```
mhchia: ditto

libp2p/network/swarm.py
```diff
@@ -87,17 +88,53 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
...
+                    "encountered swarm exception when trying to connect to %s, "
+                    "trying next address...",
+                    multiaddr,
+                    exc_info=e,
+                )
+
+        # Tried all addresses, raising exception.
+        if len(exceptions) > 0:
+            raise SwarmException(
+                "unable to connect to %s, no addresses established a successful connection "
```
mhchia: Is this a typo? I think `SwarmException` is not able to resolve the format string. Besides, we favor f-string and `.format` over `%s` style format string.

libp2p/network/swarm.py
mhchia: Besides, we should never enter here, right? If `len(exceptions) <= 0`, it means we got no exception when performing `dial_addr`. The only possible situation is `len(addrs) == 0`, which can be checked by `if not addrs:`.

libp2p/network/swarm.py
```diff
@@ -87,17 +88,53 @@ def set_stream_handler(self, stream_handler: StreamHandlerFn) -> None:
...
-            raise SwarmException(f"No known addresses to peer {peer_id}")
+        except PeerStoreError as error:
+            raise SwarmException(f"No known addresses to peer {peer_id}") from error
 
-        if not addrs:
-            raise SwarmException(f"No known addresses to peer {peer_id}")
+        exceptions: List[SwarmException] = []
+
+        # Try all known addresses
+        for multiaddr in addrs:
```
mhchia: Since you add the case we can try all addresses, would you mind adding a test for it?

libp2p/network/swarm.py
ShadowJonathan: My thought was that you guys used `%s` for exception and log formatting, I'll correct this.

libp2p/network/swarm.py
ShadowJonathan: Yup, will do, I'll add test cases for:
1. No addresses
2. Multiple addresses
3. Explicitly one address
4. One address (that doesn't work)
5. Multiple addresses (that don't work)

libp2p/network/swarm.py
ShadowJonathan: I'll change the `addr` variable to have a type hinting of it being an `Multiaddr` iterative/generator. I'll make `exceptions` be the check as to if an iteration has been run at all, and if not, change the wording on the second one as to point directly to that. (With a note in a comment that tells why exactly the program only gets there when it did not encounter any addresses in the generator)

mhchia: > I'll change the addr variable to have a type hinting of it being an Multiaddr iterative/generator.

Could you elaborate more on the reason that `addrs` should be an iterator/generator? IMO the deleted `if not addrs:` above can already sufficiently replace the `else` branch here and it's more explicit to me.

ShadowJonathan: Alright, then I'll do that.

ShadowJonathan:
I'm planning to make those first before marking it as complete and requesting reviews, yeah. 👍

PR #387, opened 2019-12-17
Store peer ids in set instead of list and check if peer id exist before access

NIC619:
## What was wrong?

Possible issue: https://github.com/ethereum/trinity/issues/1395

## How was it fixed?

- store peer ids in set so there's no duplicated peer id due to re-adding peer.
- check if peer id exist in the dict before accessing.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/libp2p/pubsub/gossipsub.py
+++ b/libp2p/pubsub/gossipsub.py
@@ -246,7 +246,7 @@ class GossipSub(IPubsubRouter):
                         # Combine fanout peers with selected peers
                         fanout_peers.update(
                             self._get_in_topic_gossipsub_peers_from_minus(
-                                topic, self.degree - fanout_size, fanout_peers
+                                topic, self.degree - fanout_size, list(fanout_peers)
                             )
                         )
                 self.fanout[topic] = fanout_peers
@@ -278,7 +278,7 @@ class GossipSub(IPubsubRouter):
             # Selects the remaining number of peers (D-x) from peers.gossipsub[topic].
             if topic in self.pubsub.peer_topics:
                 selected_peers = self._get_in_topic_gossipsub_peers_from_minus(
-                    topic, self.degree - fanout_size, fanout_peers
+                    topic, self.degree - fanout_size, list(fanout_peers)
                 )
                 # Combine fanout peers with selected peers
                 fanout_peers.update(selected_peers)
@@ -417,7 +417,7 @@ class GossipSub(IPubsubRouter):
             if num_mesh_peers_in_topic < self.degree_low:
                 # Select D - |mesh[topic]| peers from peers.gossipsub[topic] - mesh[topic]
                 selected_peers = self._get_in_topic_gossipsub_peers_from_minus(
-                    topic, self.degree - num_mesh_peers_in_topic, self.mesh[topic]
+                    topic, self.degree - num_mesh_peers_in_topic, list(self.mesh[topic])
                 )
 
                 for peer in selected_peers:
@@ -430,7 +430,7 @@ class GossipSub(IPubsubRouter):
             if num_mesh_peers_in_topic > self.degree_high:
                 # Select |mesh[topic]| - D peers from mesh[topic]
                 selected_peers = GossipSub.select_from_minus(
-                    num_mesh_peers_in_topic - self.degree, self.mesh[topic], []
+                    num_mesh_peers_in_topic - self.degree, list(self.mesh[topic]), []
                 )
                 for peer in selected_peers:
                     # Remove peer from mesh[topic]
@@ -470,10 +470,10 @@ class GossipSub(IPubsubRouter):
                     selected_peers = self._get_in_topic_gossipsub_peers_from_minus(
                         topic,
                         self.degree - num_fanout_peers_in_topic,
-                        self.fanout[topic],
+                        list(self.fanout[topic]),
                     )
                     # Add the peers to fanout[topic]
-                    self.fanout[topic].Update(selected_peers)
+                    self.fanout[topic].update(selected_peers)
 
     def gossip_heartbeat(self) -> DefaultDict[ID, Dict[str, List[str]]]:
         peers_to_gossip: DefaultDict[ID, Dict[str, List[str]]] = defaultdict(dict)
@@ -484,7 +484,7 @@ class GossipSub(IPubsubRouter):
                 if topic in self.pubsub.peer_topics:
                     # Select D peers from peers.gossipsub[topic]
                     peers_to_emit_ihave_to = self._get_in_topic_gossipsub_peers_from_minus(
-                        topic, self.degree, self.mesh[topic]
+                        topic, self.degree, list(self.mesh[topic])
                     )
 
                     msg_id_strs = [str(msg_id) for msg_id in msg_ids]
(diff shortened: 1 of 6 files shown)


libp2p/pubsub/gossipsub.py
```diff
@@ -499,7 +502,7 @@ def gossip_heartbeat(self) -> DefaultDict[ID, Dict[str, List[str]]]:
                 if topic in self.pubsub.peer_topics:
                     # Select D peers from peers.gossipsub[topic]
                     peers_to_emit_ihave_to = self._get_in_topic_gossipsub_peers_from_minus(
-                        topic, self.degree, self.fanout[topic]
+                        topic, self.degree, list(self.fanout[topic])
```
mhchia: ditto

libp2p/pubsub/pubsub.py
```diff
@@ -316,8 +315,7 @@ def _handle_dead_peer(self, peer_id: ID) -> None:
 
         for topic in self.peer_topics:
             if peer_id in self.peer_topics[topic]:
-                # Delete the entry if no other peers left
-                self.peer_topics[topic].remove(peer_id)
+                self.peer_topics[topic].discard(peer_id)
```
mhchia: Nit: as mentioned above.

libp2p/pubsub/gossipsub.py
```diff
@@ -469,10 +472,10 @@ def fanout_heartbeat(self) -> None:
                     selected_peers = self._get_in_topic_gossipsub_peers_from_minus(
                         topic,
                         self.degree - num_fanout_peers_in_topic,
-                        self.fanout[topic],
+                        list(self.fanout[topic]),
```
mhchia: ditto

libp2p/pubsub/gossipsub.py
```diff
@@ -416,24 +419,24 @@ def mesh_heartbeat(
...
+                    self.mesh[topic].add(peer)
 
                     # Emit GRAFT(topic) control message to peer
                     peers_to_graft[peer].append(topic)
 
             if num_mesh_peers_in_topic > self.degree_high:
                 # Select |mesh[topic]| - D peers from mesh[topic]
                 selected_peers = GossipSub.select_from_minus(
-                    num_mesh_peers_in_topic - self.degree, self.mesh[topic], []
+                    num_mesh_peers_in_topic - self.degree, list(self.mesh[topic]), []
```
mhchia: Same as the note about `minus: Sequence`

libp2p/pubsub/gossipsub.py
```diff
@@ -215,44 +215,47 @@ def _get_peers_to_send(
...
                 if not topic_in_fanout or (
                     topic_in_fanout and fanout_size < self.degree
                 ):
                     if topic in self.pubsub.peer_topics:
                         # Combine fanout peers with selected peers
-                        fanout_peers += self._get_in_topic_gossipsub_peers_from_minus(
-                            topic, self.degree - fanout_size, fanout_peers
+                        fanout_peers.update(
+                            self._get_in_topic_gossipsub_peers_from_minus(
+                                topic, self.degree - fanout_size, list(fanout_peers)
```
mhchia: Instead of performing `list` on `self.mesh[topic]`, does it make sense to change  `minus: Sequence`  in `_get_in_topic_gossipsub_peers_from_minus` to `Iterable` and pass `self.mesh[topic]` directly?

libp2p/pubsub/gossipsub.py
```diff
@@ -416,24 +419,24 @@ def mesh_heartbeat(
             if num_mesh_peers_in_topic < self.degree_low:
                 # Select D - |mesh[topic]| peers from peers.gossipsub[topic] - mesh[topic]
                 selected_peers = self._get_in_topic_gossipsub_peers_from_minus(
-                    topic, self.degree - num_mesh_peers_in_topic, self.mesh[topic]
+                    topic, self.degree - num_mesh_peers_in_topic, list(self.mesh[topic])
```
mhchia: ditto

libp2p/pubsub/gossipsub.py
```diff
@@ -266,25 +269,25 @@ def _get_peers_to_send(
...
+        fanout_peers: Set[ID] = self.fanout[topic] if topic_in_fanout else set()
         fanout_size = len(fanout_peers)
         if not topic_in_fanout or (topic_in_fanout and fanout_size < self.degree):
             # There are less than D peers (let this number be x)
             # in the fanout for a topic (or the topic is not in the fanout).
             # Selects the remaining number of peers (D-x) from peers.gossipsub[topic].
             if topic in self.pubsub.peer_topics:
                 selected_peers = self._get_in_topic_gossipsub_peers_from_minus(
-                    topic, self.degree - fanout_size, fanout_peers
+                    topic, self.degree - fanout_size, list(fanout_peers)
```
mhchia: ditto

libp2p/pubsub/gossipsub.py
```diff
@@ -215,44 +215,47 @@ def _get_peers_to_send(
         :param origin: peer id of the peer the message originate from.
         :return: a generator of the peer ids who we send data to.
         """
-        send_to: List[ID] = []
+        send_to: Set[ID] = set()
```
mhchia: IMO the return value of `_get_peers_to_send` can be changed to `Set[ID]` accordingly.

libp2p/pubsub/gossipsub.py
```diff
@@ -75,14 +73,11 @@ def __init__(
         self.fanout = {}
 
         # Create peer --> protocol mapping
-        self.peers_to_protocol = {}
+        self.peer_protocol = {}
```
ChihChengLiang: Are we assuming each peer only supports one protocol here?

libp2p/pubsub/gossipsub.py
```diff
@@ -121,17 +116,13 @@ def add_peer(self, peer_id: ID, protocol_id: TProtocol) -> None:
...
-            self.peers_gossipsub.append(peer_id)
-        elif protocol_id == floodsub.PROTOCOL_ID:
-            self.peers_floodsub.append(peer_id)
-        else:
+        if protocol_id not in (PROTOCOL_ID, floodsub.PROTOCOL_ID):
             # We should never enter here. Becuase the `protocol_id` is registered by your pubsub
             #   instance in multistream-select, but it is not the protocol that gossipsub supports.
             #   In this case, probably we registered gossipsub to a wrong `protocol_id`
             #   in multistream-select, or wrong versions.
             raise Exception(f"Unreachable: Protocol={protocol_id} is not supported.")
```
ChihChengLiang: We might able to make this unreachable exception a ValueError now
```suggestion
             raise ValueError(f"Protocol={protocol_id} is not supported.")
```

libp2p/pubsub/gossipsub.py
```diff
@@ -603,6 +586,12 @@ def _get_in_topic_gossipsub_peers_from_minus(
         rpc_msg: bytes = packet.SerializeToString()
 
         # 3) Get the stream to this peer
+        if sender_peer_id not in self.pubsub.peers:
```
ChihChengLiang: Feel like this is a "look before you leap" style, the downside is that we have to check every time we access the `self.pubsub.peers`. In a future refactor, raising and catching PeerNotFound exception might look nicer.

libp2p/pubsub/gossipsub.py
```diff
@@ -603,6 +586,12 @@ def _get_in_topic_gossipsub_peers_from_minus(
...
+            logger.debug(
+                "Fail to responed to iwant request from %s: peer disconnected",
```
ChihChengLiang: The wording "peer disconnected" might be misleading here. The symptom is we can't find the peer in `pubsub.peers` but the wording assumes the cause of it. Maybe just say `"can't find {sender_peer_id} in pubsub.peers"`?

libp2p/pubsub/gossipsub.py
NIC619: Yes, a peer supports either floodsub or gossipsub.

PR #388, opened 2019-12-17
Remove all asyncio.sleep statements from tests

ShadowJonathan:
## What was wrong?

Currently nearly all tests under `tests/` arbitrarily wait with `asyncio.sleep(0.01)` (or a greater number) to effectively hide the race conditions that some underlying library code are intentionally or unintentionally making for the tests, or to loop through one event loop iteration.

Talking to @mhchia while going through tests and making tests available to windows (in preparation for after #380), it's been agreed to remove these `sleep` statements and replace them with logic that cascades only when a specific requisite for that tests has been asynchronously set in place. (actual test intentions will also be written in docstring, if not clear)

## Why is this change needed?

`asyncio.sleep()` delays per wall clock, it:
1. Does not care what it's underlying system clock speed is, tests could heavily differ per machine speed.
2. Waits *arbitrarily*, it does not wait for the *specific hinted requisite/state* that that tests needs from the event loop before continuing.

### To-Do

- [ ] Work through all todo items and clear up sleep statement per test.
- [ ] Confirm testing works as intended

- [ ] Clean up commit history

### DIFF
--- a/tests/examples/test_chat.py
+++ b/tests/examples/test_chat.py
@@ -47,7 +47,7 @@ async def connect_write(host_a, host_b):
         await stream.write(message.encode())
 
     # Reader needs time due to async reads
-    await asyncio.sleep(2)
+    await asyncio.sleep(2)  # todo: remove sleep or justify existence
 
     await stream.close()
     assert received == messages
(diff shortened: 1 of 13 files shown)


ShadowJonathan:
I will close this as there has been no activity on this repository for a year.

Issue #382, opened 2019-12-16
Uniform formatting format across library


ShadowJonathan:
The current library uses a mix of `"" %`, `"%s".format` and `"{}".format` everywhere.

`"" %` formatting is heavily outdated (up for debate), alongside it being relatively un-pythonic.

Since python 3.6 (the current minimally supported version of this library), "f-strings" were introduced, this formalizes down from `"{}".format` into `f"{}"`:
```py
value = "supercalifragilisticexpialidocious"

# Percentage formatting with symbol
"%s" % value

# Percentage formatting with format call
"%s".format(value)

# Brace formatting with format call
"{0!s}".format(value)

# Brace formatting with f-string
f"{value!s}"
```

Even if f-strings would not be a good, viable or a desirable candidate, I at *least* am asking if there would be a *consistent* formatting style across the library. With a coding style document somewhere to note this down, and/or linters to enforce this (`flake8-pep3101` , `flake8-string-format`).

pipermerriam:
Note that there are [good reasons](https://github.com/ethereum/snake-charmers-tactical-manual/issues/19#issuecomment-442616682) why logging statements should use `%s` style formatting.  :+1: to all of the others being consistent.  `f-strings` are the current standard we're using elsewhere.

ShadowJonathan:
`flake-pep3101` still reports that there are modulo formatting operators used in some files, and if f-strings are the standard, it should be indicated in some coding style document somewhere, I'm still seeing some normal `.format` calls here and then.

ShadowJonathan:
But I'll get started on checking and correcting formatting mistakes, I'm even going to make a linter that enforces f-strings to incorporate in this library (as I don't see any made specifically to make f-strings a norm in the files that it would check)

mhchia:
After searching, we only have one `.format` here: https://github.com/libp2p/py-libp2p/blob/4b01c33d54d88c1d8e9e8e5247b0c97b1d45ab15/libp2p/host/basic_host.py#L96.

ShadowJonathan:
So to summarize:

f-strings everywhere *except* for log entries, which use `%s`, correct?

ShadowJonathan:
@pipermerriam 

Because if so, then I'm thinking of making another branch + draft PR that:
1. Changes everything to use f-strings (except logging)
2. Adds a styling document or a notation in `README.md` about this styling (and a note that clears up logging using `%s`)
[...]

PR #389, opened 2019-12-18
Apply f-string formatting to everything (except logging)

ShadowJonathan:
## What was wrong?

Issue #382 

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


libp2p/transport/tcp/tcp.py
```diff
@@ -91,5 +100,52 @@ def create_listener(self, handler_function: THandler) -> TCPListener:
...
+        # ipv6: (host, port, flowinfo, scopeid)
+        addr, port = sock.getsockname()[:2]
+        ip = p_code(P_IP6).name
+
+    else:
+        raise NotImplementedError(
+            f"Cannot convert socket to multiaddr, socket family is of {sock.family}"
+        )
+
+    return Multiaddr(f"/{ip}/{addr}/{t_proto}/{port}")
```
ShadowJonathan: I started changing `Multiaddr("/ip4/%s/tcp/%s" % socket.getsockname())` till I realized it needed an actual working function, so I made one.

pipermerriam: Typically we would request that this type of substantive change to functionality be submitted separately from the other changes.  Do you think you could extract this refactor into it's own PR that we'll review and merge separately and this branch can be based on top of that other branch?

ShadowJonathan: Yes, I'll do that.

libp2p/crypto/ecc.py
```diff
@@ -32,7 +32,7 @@ def get_type(self) -> KeyType:
         return KeyType.ECC_P256
 
     def verify(self, data: bytes, signature: bytes) -> bool:
-        raise NotImplementedError
+        raise NotImplementedError()
```
mhchia: Actually `raise NotImplementedError` works: https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement

libp2p/tools/interop/daemon.py
```diff
@@ -56,7 +56,7 @@ def __init__(
         is_pubsub_signing: bool = False,
         is_pubsub_signing_strict: bool = False,
     ) -> None:
-        args = [f"-listen={str(control_maddr)}"]
+        args = [f"-listen={control_maddr!s}"]
```
mhchia: Note: IMO `str(control_maddr)` is more explicit, but never mind.

ShadowJonathan: More bytes. 

Also `!s` is proper for brace formatting over `str`, and they're essentially the same thing internally.

libp2p/crypto/ecc.py
ShadowJonathan: Huh, thanks for telling. I personally like putting braces after them (to be more explicit), but I'll reverse these changes if that's requested.

mhchia:
Would you mind fixing the conflict so that I can merge it? Thanks!

ShadowJonathan:
> Would you mind fixing the conflict so that I can merge it? Thanks!
Fixed 👍

mhchia:
Sorry, that I failed to merge it previously. Would you mind fixing the conflict again? 😢 Thank you!

ShadowJonathan:
@mhchia fixed it 👍

mhchia:
Thank you! Please ignore the failure of `tests_interop` for now, due to some breaking changes in the dependency `p2pclient`.

PR #390, opened 2019-12-19
Cherry-pick changes mentioned from #389 into transport/tcp.py

ShadowJonathan:
## Describe the fix.

`transport/tcp.py::_multiaddr_from_socket()` assumes that the socket uses `AF_INET` as family and `SOCK_STREAM` as a socket type, everything is also hardcoded to ipv4.

This fix should make this function more smarter in handling ipv4+6, and allows it to detect if the socket is UDP or TCP.

Also in general it's recommended to move this to something like `transport/utils.py` or similar.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/libp2p/transport/tcp/tcp.py
+++ b/libp2p/transport/tcp/tcp.py
@@ -1,9 +1,11 @@
 import asyncio
-from socket import socket
+import socket
 import sys
-from typing import List
+from typing import List, Optional
 
 from multiaddr import Multiaddr
+from multiaddr.protocols import P_IP4, P_IP6, P_TCP, P_UDP
+from multiaddr.protocols import protocol_with_code as p_code
 
 from libp2p.network.connection.raw_connection import RawConnection
 from libp2p.network.connection.raw_connection_interface import IRawConnection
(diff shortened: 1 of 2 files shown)


ShadowJonathan:
Will do! 👍

ShadowJonathan:
Looks like Travis doesn't support ipv6 on some containers: 
    https://docs.travis-ci.com/user/reference/overview/

ShadowJonathan:
I split the tests so when they *are* ran on hosts with ipv6 enabled, they will be run, otherwise, they won't.

Currently, ipv6 is not enabled on travis ci, but when ran locally on a computer with ipv6, or when switching to another CI host with ipv6, these will be automatically ran.

ShadowJonathan:
It looks like `socket.has_ipv6` is only true when its compiled with support for ipv6, not when the system has ipv6 support.

ShadowJonathan:
One small last opinion: we should not build a multiaddress from a raw sockets, as it's possible that a socket type/family does not equate it's underlying transport (e.g. QUIC on TCP), and so an erroneous/false multiaddress is built from a raw socket. "Generating" multiaddresses should only happen in a transport listener. 

If this function is specific to testing (a specific transport), it should be moved to the `tests/` folder and removed from the main library.

ralexstokes:
@ShadowJonathan want to resolve the merge conflicts here? otherwise i'll close as the PR has gotten stale.

PR #394, opened 2019-12-24
Fix `close` an already reset connection

NIC619:
## What was wrong?

https://github.com/ethereum/trinity/issues/1422
## How was it fixed?

Catch `ConnectionResetError` error when trying to close a connection.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/libp2p/network/connection/raw_connection.py
+++ b/libp2p/network/connection/raw_connection.py
@@ -58,5 +58,5 @@ class RawConnection(IRawConnection):
         try:
             await self.writer.wait_closed()
         # In case the connection is already reset.
-        except (ConnectionResetError, RawConnError):
+        except ConnectionResetError:
             return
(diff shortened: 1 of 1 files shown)


libp2p/network/connection/raw_connection.py
```diff
@@ -55,4 +55,8 @@ def __init__(
         self.writer.close()
         if sys.version_info < (3, 7):
             return
-        await self.writer.wait_closed()
+        try:
+            await self.writer.wait_closed()
+        # In case the connection is already reset.
+        except (ConnectionResetError, RawConnError):
```
ChihChengLiang: The writer is an `asyncio.StreamWriter` type, so we don't need to handle `RawConnError` here right?

DonaldTsang:
@ShadowJonathan any comments?

ShadowJonathan:
I'm thinking about a proposal to go over the transport-connection-protocol stack to make it interoperable upwards and downwards with these kind of close/reset events, but yeah, this looks good to me.

PR #398, opened 2020-01-19
Trio: fix host, peer, protocols, and examples

mhchia:
Part of the https://github.com/libp2p/py-libp2p/pull/365:
- Host
- PeerInfo
- Protocol muxer
- Examples

### DIFF
--- a/examples/chat/chat.py
+++ b/examples/chat/chat.py
@@ -1,11 +1,10 @@
 import argparse
-import asyncio
 import sys
-import urllib.request
 
 import multiaddr
+import trio
 
-from libp2p import new_node
+from libp2p import new_host_trio
 from libp2p.network.stream.net_stream_interface import INetStream
 from libp2p.peer.peerinfo import info_from_p2p_addr
 from libp2p.typing import TProtocol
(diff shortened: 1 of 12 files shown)


examples/chat/chat.py
```diff
@@ -26,53 +25,48 @@
...
+            # Associate the peer with local ip address
+            await host.connect(info)
+            # Start a stream with the destination.
+            # Multiaddress of the destination peer is fetched from the peerstore using 'peerId'.
+            stream = await host.new_stream(info.peer_id, [PROTOCOL_ID])
 
-        localhost_opt = " --localhost" if localhost else ""
+            nursery.start_soon(read_data, stream)
+            nursery.start_soon(write_data, stream)
+            print("Connected to peer %s" % info.addrs[0])
```
ChihChengLiang: nit: let's do f-string here

libp2p/__init__.py
```diff
@@ -43,28 +33,27 @@ def generate_peer_id_from(key_pair: KeyPair) -> ID:
...
 def initialize_default_swarm(
-    key_pair: KeyPair,
-    id_opt: ID = None,
-    transport_opt: Sequence[str] = None,
+    key_pair: KeyPair = None,
     muxer_opt: TMuxerOptions = None,
     sec_opt: TSecurityOptions = None,
     peerstore_opt: IPeerStore = None,
-) -> Swarm:
+) -> INetworkService:
```
ChihChengLiang: Why are we changing to a more abstract type to the function output here?

mhchia: I intended to use interfaces over implementations. I can change it back if you think it's more reasonable.

ChihChengLiang: Using interfaces sounds good 👍

pipermerriam: Agreed.  This has helped us a lot in both the trinity and py-evm codebases to establish concrete interface definitions and to use them ubiquitously in our type hints.

mhchia:
Closed by #365

PR #396, opened 2020-01-19
Trio: transport

mhchia:
The transport part of https://github.com/libp2p/py-libp2p/pull/365.

PR #402, opened 2020-01-19
Trio: fix settings, factories, and helpers

mhchia:
The rest of the code in #365, including the settings, factories, and the helper functions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,8 +7,8 @@ from setuptools import find_packages, setup
 extras_require = {
     "test": [
         "pytest>=4.6.3,<5.0.0",
-        "pytest-xdist>=1.30.0,<2",
-        "pytest-asyncio>=0.10.0,<1.0.0",
+        "pytest-xdist>=1.30.0",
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -74,6 +74,9 @@
     "pynacl==1.3.0",
     "dataclasses>=0.7, <1;python_version<'3.7'",
     "async_generator==1.10",
+    "trio>=0.13.0",
```
ChihChengLiang: Missing `trio_typing`

mhchia: Could you point out where we require `trio_typing`? I'm not able to find it used in the codebase.

ChihChengLiang: Oh, please check here. https://github.com/libp2p/py-libp2p/pull/396/files#r369460417

mhchia: Nice catch! Fixed in https://github.com/libp2p/py-libp2p/pull/365/commits/e57d01f360de9fb101adffa686cf5aff414fc8f8.

mhchia:
Closed by #365

PR #405, opened 2020-02-07
Noise: skeleton of transport and connection

mhchia:
### What's inside this PR
* security.noise
    * transport
        * `Transport(ISecureTransport)`
            * Note: Given options, use the desired pattern to upgrade an `IRawConnection` to `NoiseConnection`.
    * patterns
        * `IPattern(ABC)`
            * Note: Handle the whole handshake process.
            * `PatternXX(IPattern)`
                * Note: Not complete yet. Will finish authentication of libp2p pubkey in the next PR.
            * (Other patterns in the future)
    * connection
        * `NoiseConnection(ISecureConn)`
            * Note: It is just a fake implementation now. Will support [en|de]crypt in the latter PRs.
    * pb
        * Note: The wire message described in protobuf. It is not used currently. Will be used in the next PR.
* tests
    * `noise_conn_factory`
        * Note: upgrading `IRawConnection` to `NoiseConnection`.
    * test_noise
        * Note: Add a simple test for both `noise.Transport` and `NoiseConnection`.


libp2p/security/noise/transport.py
```diff
@@ -0,0 +1,69 @@
+from libp2p.crypto.keys import KeyPair, PrivateKey
+from libp2p.io.msgio import ReadWriteCloser
+from libp2p.network.connection.raw_connection_interface import IRawConnection
+from libp2p.peer.id import ID
+from libp2p.security.base_session import BaseSession
+from libp2p.security.secure_conn_interface import ISecureConn
+from libp2p.security.secure_transport_interface import ISecureTransport
+from libp2p.typing import TProtocol
+
+PROTOCOL_ID = TProtocol("/")
```
ralexstokes:
```suggestion
PROTOCOL_ID = TProtocol("/noise")
```

mhchia:
@ralexstokes No problem! Thanks for the review. 🙏

mhchia:
This one is ready for review. Most of the details are implemented in #406

libp2p/network/stream/net_stream.py
```diff
@@ -51,14 +51,14 @@ def set_protocol(self, protocol_id: TProtocol) -> None:
         except MuxedStreamReset as error:
             raise StreamReset() from error
 
-    async def write(self, data: bytes) -> int:
+    async def write(self, data: bytes) -> None:
```
ralexstokes: what happens if we can't write all of `data` to the underlying stream? the idea w/ this interface is to allow a caller to detect this scenario and re-attempt the write if that happens

libp2p/security/noise/patterns.py
```diff
@@ -0,0 +1,116 @@
...
+            NoiseKeypair.STATIC, self.noise_static_key.to_bytes()
+        )
+        return noise_state
+
+
+class PatternXX(BasePattern):
+    def __init__(
+        self, local_peer: ID, libp2p_privkey: PrivateKey, noise_static_key: PrivateKey
+    ) -> None:
+        self.protocol_name = b"Noise_XX_25519_ChaChaPoly_SHA256"
```
ralexstokes: is this the specific noise suite we are using in `libp2p-noise` according to that spec?

mhchia: Yes, it's the protocol we use in the XX pattern. Ref: https://github.com/libp2p/specs/tree/master/noise#valid-noise-protocol-names

libp2p/network/stream/net_stream.py
mhchia: Good point. I'm changing this because I didn't see this use case(returns before all of the data is written) present in our codebase and also the underlying transport e.g. [`trio.SocketStream.send_all`](https://trio.readthedocs.io/en/stable/reference-io.html#trio.abc.SendStream.send_all). `write` is asynchronous and we probably can guarantee that `write` returns only when all of the data is written. I can also change it back if we would prefer to allow `write` to return if it only writes part of the data.

ralexstokes: we should avoid underlying implementations from guiding our interface choices but i suppose we can leave this alone for now just to bias towards speed :)

Issue #375, opened 2019-12-05
Implement noise-libp2p secure channel handshake


hwwhww:
## What was wrong?

Spec: https://github.com/libp2p/specs/blob/master/noise/README.md

Noise protocol Python implementations:
1. https://github.com/tgalal/dissononce
2. https://github.com/plizonczyk/noiseprotocol

## How can it be fixed?

Fill this section in if you know how this could or should be fixed.

mhchia:
Forgot to assign to myself previously. I will have a PR for it soon.

tintinweb:
Hey @mhchia @hwwhww, just a quick question: what is the status of the noise implementation? While secio works noise seems to break when talking to a nimbus node.

this is my setup (taken from the trinity host implementation):

https://github.com/ethereum/trinity/blob/81beae51ab34fe2ac57de3cf301730b7522fbcf0/trinity/nodes/beacon/host.py#L58-L76

```python
        muxer_transports_by_protocol = {MPLEX_PROTOCOL_ID: Mplex}
        noise_key = ed25519.create_new_key_pair()
        security_transports_by_protocol = {
            TProtocol(secio.ID): secio.Transport(
                key_pair
            ),
            TProtocol(noise.PROTOCOL_ID): noise.Transport(
                key_pair, noise_key.private_key
            )
        }
        upgrader = TransportUpgrader(
            security_transports_by_protocol, muxer_transports_by_protocol
        )

```

```
TRC 2020-08-05 10:28:10+02:00 upgrading incoming connection              topics="switch" tid=12728997 file=switch.nim:345 conn= oid=5f2a6d9a121ed2455a7ef81a
TRC 2020-08-05 10:28:10+02:00 initiating handshake                       topics="multistream" tid=12728997 file=multistream.nim:48 codec="\x13/multistream/1.0.0\n"
TRC 2020-08-05 10:28:10+02:00 registering proto handler                  topics="switch" tid=12728997 file=multistream.nim:179 codec=/noise
TRC 2020-08-05 10:28:10+02:00 handle: starting multistream handling      topics="multistream" tid=12728997 file=multistream.nim:115
TRC 2020-08-05 10:28:10+02:00 handle: got request for                    topics="multistream" tid=12728997 file=multistream.nim:121 ms=/noise
TRC 2020-08-05 10:28:10+02:00 found handler for                          topics="multistream" tid=12728997 file=multistream.nim:143 protocol=/noise
[...]

PR #406, opened 2020-02-15
Noise: `PatternXX` and testing with go

mhchia:
## What was wrong?

#375 

## How was it fixed?
Based on #405 

* Feature
    * Noise
        * Handshake
            * `Pattern XX`
        * Encryption/decryption
            * `SecureSession`(a buffered stream reader originally used in secio) is refactored and used by both `Noise` and `Secio` now.
* Refactor
    * msgio: dedup the original similar implementations.
        * `MsgReadWriter`
        * `FixedSizeLenMsgReadWriter`
            * `NoiseReadWriter`
        * `VarIntLengthMsgReadWriter`
            * `PlaintextHandshakeReadWriter`
* Fix
    * Plaintext: use `VarIntLengthMsgReadWriter` over `FixedSizeLenMsgReadWriter`. Ref: https://github.com/libp2p/go-libp2p-core/pull/74/
* Tests
    * Transport, encryption/decryption, security multistream.
    * Interop test with go implementation
        * through the modified [`go-libp2p-daemon`](https://github.com/mhchia/go-libp2p-daemon/commit/558b58db41c1821ce9d2537235ca1bbe0030ce3e) against different security protocol.
            * Plaintext
            * Noise: [`go-libp2p-noise`](https://github.com/libp2p/go-libp2p-noise)
                * Issue: msg#3(initiator sends the payload to the responder) is different from the one in go implementation. Workaround this through ignoring the first `32` bytes for now. Will investigate what is wrong later.
* Misc
    * Re-compile protobufs
        * `noise.proto`
        * `plaintext.proto`

### TODO
- [ ] Confirm why msg#3 is different between py and go.
- [ ] Add more exceptions and handling.
- [ ] Add more comments.

### Reference
- [noise spec](https://noiseprotocol.org/noise.html#message-format)
- [noise-libp2p spec](https://github.com/libp2p/specs/tree/master/noise#xx)
- [go-libp2p-noise](https://github.com/libp2p/go-libp2p-noise)

#### Cute Animal Picture


mhchia:
Though there are things to be added and confirmed, I think this should be ready for review. We can talk with go implementation through `noise` now. 🎉

libp2p/io/msgio.py
```diff
@@ -5,80 +5,85 @@
...
-async def read_length(reader: Reader) -> int:
-    length_bytes = await read_exactly(reader, SIZE_LEN_BYTES)
+async def read_length(reader: Reader, size_len_bytes: int) -> int:
+    length_bytes = await read_exactly(reader, size_len_bytes)
     return int.from_bytes(length_bytes, byteorder=BYTE_ORDER)
 
 
-def encode_msg_with_length(msg_bytes: bytes) -> bytes:
-    len_prefix = len(msg_bytes).to_bytes(SIZE_LEN_BYTES, "big")
+def encode_msg_with_length(msg_bytes: bytes, size_len_bytes: int) -> bytes:
```
pipermerriam: Why is the size passed in here instead of measuring it from `len(msg_bytes)`?

libp2p/io/msgio.py
```diff
@@ -5,80 +5,85 @@
...
 
+    @abstractmethod
+    async def next_msg_len(self) -> int:
+        ...
 
-class MsgIOReader(ReadCloser):
-    read_closer: ReadCloser
-    next_length: Optional[int]
+    @abstractmethod
+    def encode_msg(self, msg: bytes) -> bytes:
```
pipermerriam: should this be public?  Is it used as an external API or only called internally?

mhchia: It's only called internally, but I expect subclasses to implement it to make `write_msg ` work. So as `next_msg_len` to make `read_msg` work.

libp2p/io/msgio.py
mhchia: `size_len_bytes` means the size of the length field in the message. `len(msg_bytes)` is encoded and put in the length field.

mhchia:
I'm merging this PR since it gets too large and it already works fine with the go implementation. Will open a new one if everything gets updated.

Issue #349, opened 2019-11-11
Exploring Sans-io


ChihChengLiang:
This is a different approach to fix #174 

We first refactor classes using the [Sans-io](https://sans-io.readthedocs.io/) pattern. Separate out all async functions and then support both trio and even asyncio if we want to.

Here's an example how we can refactor the class GossipSub

before: The async function publish writes some data to peers
```python
class GossipSub:
    async def publish(self, msg_forwarder: ID, pubsub_msg: rpc_pb2.Message) -> None:
        ...
        for peer_id in peers_gen:
            stream = self.pubsub.peers[peer_id]
            await stream.write(encode_varint_prefixed(rpc_msg.SerializeToString()))
```
after: maximally abstracting out the async part 
```python
class GossipSub:
   def publish(self, msg_forwarder: ID, pubsub_msg: rpc_pb2.Message) -> PublishEvent:
        ...
        return [PublishEvent(peer_id=peer_id, data=data) for peer_id in peers_gen]
            ...
            await stream.write(encode_varint_prefixed(rpc_msg.SerializeToString()))

class TrioGossipSubService:
    async def publish(self):
        events = self.gossipsub.publish(...)
        for e in events:
            stream = get_stream(e.peer_id)
            await stream.write(e.data)
```

- Pros:
  - Clean separation of IO
  - Better testability
- Cons:
  - Might change the interface
  - Takes more effort to refactor
  - The refactor might obscure the code logic

ralexstokes:
you could go a step further and put the calls to `get_stream` and `stream.write` into a(n async) closure that is returned from the `publish` function.

we then get a `TrioHost` or something that is basically just `await`ing things and handling continuations between tasks

ralexstokes:
i like all of these pros :)

i agree that this is the more heavy-handed approach in that it is a more serious rewrite of the library, rather than just `s/asyncio/trio`.

ShadowJonathan:
This requires large parts of the library to be synchronous, which could take away advantages in `trio` which use `await` keywords as "checkpoints" to allow other code to run, making logic code actually asynchronous.

With this implimentation, logic code could run blocking indefinitely till it gives control back to its async-envoked entry point, after which only *then* async code is allowed to run (such as vital parts like ping, heartbeats, or connection management)

Unless these synchronous functions are ran in an alternative thread, and all calls and accesses to the *other* thread with the event loop are thread-safe and without race conditions, I personally don't think this is a good idea.

PR #407, opened 2020-02-26
Removed explicit 'Nursery' from Listener

aratz-lasa:
## What was wrong?
The [IListener](https://github.com/libp2p/py-libp2p/blob/99f505d6d782b363cfcaf9ebc18acd3ca462a506/libp2p/transport/listener_interface.py#L10) interface is tightly coupled to the I/O implementation (Trio). This makes the application not to be agnostic to the I/O implementation.

Even if it goes quite against Trio's Nursery philosophy, we must decide whether we prioritize the abstractions and I/O agnostic implementations, or to the contrary, Trio's structured concurrent philosophy.
In order to understand better trio's Philosophy you can check: [Trio's philosphy]( https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/ )

## How was it fixed?

Created a static class [TrioGlobalNursery](https://github.com/aratz-lasa/py-libp2p/blob/795d53f19f399f563015ee29326eab9cdf1bf3e6/libp2p/io/trio.py#L45) with two methods:
1.  **set_global_nursery_and_run**: which sets the specified `nursery` as 'Global' and runs an asynchronous function, so that if you later before the function returns `get_global_nursery`, it returns the 'Global Nursery'
2.  **get_global_nursery**: returns the current 'Global Nursery'

This way, for running the IListener `listen` you should do it by using:

`await  TrioGlobalNursery.set_global_nursery_and_run(nursery, listener.listen, maddr)`

Instead of: 

`await listener.listen(maddr, nursery)`

### To-Do

- [x] Check why py37-interop fails

### DIFF
--- a/libp2p/network/swarm.py
+++ b/libp2p/network/swarm.py
@@ -40,7 +40,6 @@ def create_default_stream_handler(network: INetworkService) -> StreamHandlerFn:
 
 
 class Swarm(Service, INetworkService):
-
     self_id: ID
     peerstore: IPeerStore
     upgrader: TransportUpgrader
@@ -276,7 +275,9 @@ class Swarm(Service, INetworkService):
                 #   I/O agnostic, we should change the API.
                 if self.listener_nursery is None:
                     raise SwarmException("swarm instance hasn't been run")
-                await listener.listen(maddr, self.listener_nursery)
+                await self.listener_nursery.start(
+                    listener.listen, maddr  # type: ignore
+                )
 
                 # Call notifiers since event occurred
                 await self.notify_listen(maddr)
--- a/libp2p/pubsub/gossipsub.py
+++ b/libp2p/pubsub/gossipsub.py
@@ -391,7 +391,7 @@ class GossipSub(IPubsubRouter, Service):
             await trio.sleep(self.heartbeat_interval)
 
     def mesh_heartbeat(
-        self
+        self,
     ) -> Tuple[DefaultDict[ID, List[str]], DefaultDict[ID, List[str]]]:
         peers_to_graft: DefaultDict[ID, List[str]] = defaultdict(list)
         peers_to_prune: DefaultDict[ID, List[str]] = defaultdict(list)
--- a/libp2p/tools/factories.py
+++ b/libp2p/tools/factories.py
@@ -67,7 +67,7 @@ def security_transport_factory(
 
 @asynccontextmanager
 async def raw_conn_factory(
-    nursery: trio.Nursery
+    nursery: trio.Nursery,
 ) -> AsyncIterator[Tuple[IRawConnection, IRawConnection]]:
     conn_0 = None
     conn_1 = None
@@ -81,7 +81,7 @@ async def raw_conn_factory(
 
     tcp_transport = TCP()
     listener = tcp_transport.create_listener(tcp_stream_handler)
-    await listener.listen(LISTEN_MADDR, nursery)
+    await nursery.start(listener.listen, LISTEN_MADDR)
     listening_maddr = listener.get_addrs()[0]
     conn_0 = await tcp_transport.dial(listening_maddr)
     await event.wait()
@@ -351,7 +351,7 @@ async def swarm_pair_factory(
 
 @asynccontextmanager
 async def host_pair_factory(
-    is_secure: bool
+    is_secure: bool,
 ) -> AsyncIterator[Tuple[BasicHost, BasicHost]]:
     async with HostFactory.create_batch_and_listen(is_secure, 2) as hosts:
         await connect(hosts[0], hosts[1])
@@ -370,7 +370,7 @@ async def swarm_conn_pair_factory(
 
 @asynccontextmanager
 async def mplex_conn_pair_factory(
-    is_secure: bool
+    is_secure: bool,
 ) -> AsyncIterator[Tuple[Mplex, Mplex]]:
     muxer_opt = {MPLEX_PROTOCOL_ID: Mplex}
     async with swarm_conn_pair_factory(is_secure, muxer_opt=muxer_opt) as swarm_pair:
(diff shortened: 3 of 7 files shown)


mhchia:
@aratz-lasa No worries about the failure of `py37-interop`. I overwrote the branch `test/add-options` of my cloned repo [`mhchia/go-libp2p-daemon`](https://github.com/libp2p/py-libp2p/blob/006002f68706d67263324e8c2c44737d50aaf0e1/tests_interop/go_pkgs/install_interop_go_pkgs.sh#L17), which is needed by the CI test. This branch depends on a library that only exists in `go==1.13`. I am fixing it in #406. If needed, I can cherrypick the commits to another PR to make the CI happy.

aratz-lasa:
@mhchia Okay, thank you!
Should I ping someone in order to discuss this Trio Nursery work-around?

pipermerriam (changes requested):
I'm :-1: 

It accomplishes removal of the `trio` specific nursery argument for global application state which in my opinion simply hides the problem behind a common antipattern. The `Listener` is still coupled with the trio nursery, only now it is hidden via a global embedded within the class.

I don't think this improves the current state of things.

ralexstokes (commented):
thanks for this PR!

i agree we should remove the reference to the `trio.Nursery` from the `IListener` interface.

but given how lightweight trio nurseries are, we can just do something like

```python
class TrioTCPListener(IListener):
    async def listen(self, maddr):
            # do stuff to start the actions of this listener
            ...
```

if we do in fact need to control the nursery, then the instantiator of `TrioTCPListener` can provide one in the `__init__`.

```python
class TrioTCPListener(IListener):
    def __init__(self, nursery):
        # we can use this nursery inside e.g. `listen` if we need to use a specific nursery
        self._nursery = nursery
```

it seems to be adding an extra level of indirection to introduce the `TrioGlobalNursery` when we can just take advantage of hiding these implementation details behind our nice `IListener` interface...

aratz-lasa:
@ralexstokes You are completely right and that was my first thought, however, we need to control the nursery because it is the one used in the `Swarm`. Also there is no possibility to put the Nursery on the `__init__`  because in order to create a `Listener` we use `transport.create_listener(conn_handler)`, instead of directly creating it.

aratz-lasa:
@pipermerriam I understand it, and I am not sure if this the best way to go. I have made a pull request to discuss about it

pipermerriam:
@aratz-lasa apologies if my tone came across negative.  I do appreciate the PR and your contribution so not trying to shut you down.  I'm very open to considering alternatives.

Short term, I'm personally fine with the trio dependency but that is primarily because transitioning to trio is inline with the use case I specifically have for this library and I want to get it off the ground and working in the wild with minimal delays.  Medium/long term, I'm very game for this library to move towards being agnostic to asyncio/trio.

aratz-lasa:
> @aratz-lasa apologies if my tone came across negative. I do appreciate the PR and your contribution so not trying to shut you down. I'm very open to considering alternatives.
> 
> Short term, I'm personally fine with the trio dependency but that is primarily because transitioning to trio is inline with the use case I specifically have for this library and I want to get it off the ground and working in the wild with minimal delays. Medium/long term, I'm very game for this library to move towards being agnostic to asyncio/trio.

@pipermerriam Okay, okay, I understand you. And what if for now, we change the `nursery` for `task_status`. This way, on the one hand, there is no `nursery` explicit passing as argument, and on the other hand, it is an optional argument where you cannot use it if you don't want to.

ralexstokes:
@aratz-lasa 

> Also there is no possibility to put the Nursery on the __init__ because in order to create a Listener we use transport.create_listener(conn_handler), instead of directly creating it.

why can't we just directly create the `TCPListener`?

aratz-lasa:
@ralexstokes 
> why can't we just directly create the `TCPListener`?

Well, even if right now is only implemented TCP, the `Swarm` must be Transport layer agnostic, that is why we invoke  `transport.create_listener(conn_handler)`, instead of directly creating a `TCPListener`. 

Or at least that is what I understood reading the source code, I may be wrong. I am still quite new to this project :)

ralexstokes:
@aratz-lasa so this quickly becomes a bigger question about API and the higher-order structure of this library but if we want to remove _concrete_ IO from our _abstract_ interfaces (i think a worthy goal), then a straightforward way to do this would be to have `create_listener` return some opaque `Task` type (like in `async-service`, e.g. https://github.com/ethereum/async-service/blob/master/async_service/base.py#L100) which is the passed up to outer layers of the libp2p library until it hits a "task runner", aka a thing that can drive the concurrency here... this was kind of the thinking that lead to the discussion in this issue: https://github.com/libp2p/py-libp2p/issues/349

right now a `Swarm` (via inheriting `async_service.Service`) is the "task runner" so it seems clear how we can deploy this strategy here...

ralexstokes:
it also isn't immediately clear to me we need a separate `listener_nursery` but this is likely a different conversation :)

the fewer places we can "effect" the concurrency the simpler time we will have debugging/maintaining/ etc

aratz-lasa:
@ralexstokes I think you are right, do I commit a version with `create_listener` returning a `Task`? Or do we leave it for now? 
If we leave it, I would recommend changing  `nursery` by `task_status` from `IListener` interface, as in my last commit on this pull request.

aratz-lasa:
> it also isn't immediately clear to me we need a separate `listener_nursery` but this is likely a different conversation :)
> 
> the fewer places we can "effect" the concurrency the simpler time we will have debugging/maintaining/ etc

Yeah, I also thought about it. We should have a clear strategy for when to use `Nursery`, and what is every `nursery`'s scope. Because depending on the strategy, as you said, maybe there is no need for having a nursery on the `Swarm`.

acul71:
Closing as superseded. The nursery parameter was removed from `IListener.listen()` in #1308 (which also adopted #1055). Thank you for the early design work.

Issue #344, opened 2019-11-06
Full-feature `ping` implementation


ralexstokes:
Reference: https://github.com/libp2p/py-libp2p/pull/315#pullrequestreview-309232342

We have the basic viable implementation of `ping`. We should add features like ~sharing our local view of our address and~ measuring RTTs to remote peers.

edit: i mixed up which features are in `ping` and which are in `identify`

mystical-prog:
hey @dhuseby 
can I work on this issue ? there are some clarifications that I think we should discuss if this is assigned to me

dhuseby:
Absolutely!! Take it!! 🚀

mystical-prog:
> Absolutely!! Take it!! 🚀 

Actually @dhuseby it has been solved and PR merged, so you can close this issue now

PR #409, opened 2020-02-27
Implemented PING fully-featured

aratz-lasa:
First draft of PingService, which executes pings and implements RTT calculation feature.

## What was wrong?
There was no PingService that implemented the RTT feature, nor an easy way for creating a Ping requests Loop.

Issue #344 

## How was it fixed?

Created a class Called [PingService](https://github.com/aratz-lasa/py-libp2p/blob/ef2e2487af9ccc8260bfcb8a8ecf51120a482847/libp2p/host/ping.py#L69), which implements two methods:
1. `ping(peer_id)`: method for executing a ping. It opens a stream, pings it, closes the stream, and returns the RTT.
2. `ping_loop(peer_id, ping_amount)`: method that returns a Ping iterator. If no `ping_amount` is passed as argument, it creates an infinite iterator. Every iteration returns the RTT.

### To-Do

- [ ] Decide whether it is an intuitive design, and if it should be coupled to `handle_ping`
- [ ] Decide where to place the service. Maybe as an attribute of the `Host`?
- [ ] Check if the RTT calculation is accurate and if its format should be in miliseconds
- [ ] Document it well enough

### DIFF
--- a/libp2p/host/ping.py
+++ b/libp2p/host/ping.py
@@ -1,7 +1,13 @@
 import logging
+import math
+import secrets
+import time
+from typing import Union
 
 import trio
 
+from libp2p.exceptions import ValidationError
+from libp2p.host.host_interface import IHost
 from libp2p.network.stream.exceptions import StreamClosed, StreamEOF, StreamReset
 from libp2p.network.stream.net_stream_interface import INetStream
 from libp2p.peer.id import ID as PeerID
@@ -12,6 +18,21 @@ PING_LENGTH = 32
 RESP_TIMEOUT = 60
 
 logger = logging.getLogger("libp2p.host.ping")
+
+
+async def handle_ping(stream: INetStream) -> None:
+    """``handle_ping`` responds to incoming ping requests until one side errors
+    or closes the ``stream``."""
+    peer_id = stream.muxed_conn.peer_id
+
+    while True:
+        try:
+            should_continue = await _handle_ping(stream, peer_id)
+            if not should_continue:
+                return
+        except Exception:
+            await stream.reset()
+            return
 
 
 async def _handle_ping(stream: INetStream, peer_id: PeerID) -> bool:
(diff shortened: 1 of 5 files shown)


libp2p/host/ping.py
```diff
@@ -45,16 +66,65 @@
...
+        try:
+            return await _ping(self._stream)
+        except trio.EndOfChannel:
+            await _close_stream(self._stream)
+            raise StopAsyncIteration
+
+
+async def _ping(stream: INetStream) -> int:
+    ping_bytes = secrets.token_bytes(PING_LENGTH)
+    before = int(time.time() * 10 ** 6)  # convert float of seconds to int miliseconds
```
ralexstokes: what do you think about just keeping the native `time.time()` value as a `float` for as long as possible (possibly just letting the consumer of the rtt convert to `int` if _they_ want...)

libp2p/host/ping.py
```diff
@@ -45,16 +66,65 @@
...
+    if ping_bytes != pong_bytes:
+        raise ValidationError("Invalid PING response")
+    return rtt
+
+
+async def _close_stream(stream: INetStream) -> None:
+    try:
+        await stream.close()
+    except Exception:
+        pass
```
ralexstokes: worth at least logging this exception, if not letting bubble up to some caller

libp2p/host/ping.py
```diff
@@ -45,16 +66,65 @@
...
+    ) -> "PingIterator":
+        stream = await self._host.new_stream(peer_id, (ID,))
+        ping_iterator = PingIterator(stream, ping_amount)
+        return ping_iterator
+
+
+class PingIterator:
+    def __init__(self, stream: INetStream, ping_amount: Union[int, float]):
+        self._stream = stream
+        self._ping_limit = ping_amount
```
ralexstokes: i think `ping_limit` is much clearer than `ping_amount`.

want to just use `ping_limit` everywhere?

libp2p/host/ping.py
```diff
@@ -14,6 +20,21 @@
...
+    peer_id = stream.muxed_conn.peer_id
+
+    while True:
+        try:
+            should_continue = await _handle_ping(stream, peer_id)
+            if not should_continue:
+                return
+        except Exception:
+            await stream.reset()
+            return
```
ralexstokes: any particular reason to move this chunk of code up here?

just curious.... at first pass on this review i figured you had changed some of the logic but it seems to be the same(!)

libp2p/host/ping.py
```diff
@@ -45,16 +66,65 @@
...
         except Exception:
-            await stream.reset()
-            return
+            await _close_stream(stream)
+            raise
+
+    async def ping_loop(
+        self, peer_id: PeerID, ping_amount: Union[int, float] = math.inf
+    ) -> "PingIterator":
+        stream = await self._host.new_stream(peer_id, (ID,))
```
ralexstokes: would be helpful to leave a docstring describing why we have `ping` and `ping_loop`.

it turns out that `ping` is just `ping_loop` specialized to `ping_amount == 1`. which also suggest we may be able to simplify the implementation here...

libp2p/host/ping.py
aratz-lasa: First of all thank you for your comments. I learn a lot from them!

I think it is a good idea to keep it as float. Do you think it is better to leave it also as native seconds? I think I prefer to return it as a miliseconds float

libp2p/host/ping.py
aratz-lasa: I like it! I'll change it to `ping_limit`

libp2p/host/ping.py
aratz-lasa: Oh I'm sorry, I think this happened because at first I moved it inside `PingService` as a method. But at the end I left it out, and I think I moved it accidentally

libp2p/host/ping.py
aratz-lasa: Okay, it is a good idea to write a docstring!

The reason why I thought of separating into two methods is because `ping_loop` is a way to use it as an intuitive `for loop`. However, it is not so straightforward, you must generate the iterator with `ping_loop` and then construct a `for loop`. So I created the second method `ping` for when you just want a quick, simple and one-line ping (maybe for a rapid keep-alive, or something else). 

If am open to discussion whether it really is worth it or useful or not :)

libp2p/host/ping.py
aratz-lasa: I think you are right, it is better to let bubble up to the caller

libp2p/host/ping.py
aratz-lasa: Also the loop is useful in case you want to add some more functionality between each ping. (or maybe it is not a good feature?)

However, I have been thinking and maybe it is a good idea to add a second argument to the `ping(peer_id)`, which specifies the amount of pings and return a list of `RTTs`. Something like `rtts = ping_service.ping(peer_id, amount=5)`.

aratz-lasa:
@ralexstokes I have updated the PingService, whenever you can let me know what you think

mystical-prog:
@dhuseby I think this PR can be closed (without merging)

pacrob:
closed by #480

PR #424, opened 2020-08-25
Interop tests updated and fixed.

g-r-a-n-t:
## What was wrong?

Noise was not working with the most recent versions of other noise-libp2p implementations.

For some context, this is required for eth2 networking.

## How was it fixed?

- [x] upgraded the versions of go-libp2p that we test against.
- [x] modified the way we set security on the go libp2p daemon, as this is now supported out of the box and we don't need to use a special-purpose branch.
- [x] removed prefixes on noise handshake payloads, it looks like this was not desired in the go implementation and has since been removed.
- [x] added explicit params to `key_pair.public.public_bytes`.
- [x] removed interop testing of the plaintext protocol, as it's not something that's meant to be used in prod.
- [x] added interop testing for Secio.
- [ ] fixed failing remote reset tests.

I've added skip tags to the failing tests mentioned above (`test_net_stream_write_after_remote_reset` and `test_net_stream_read_after_remote_reset`). The tests themselves aren't actually running, because setup fails and returns the error:

```
RuntimeError: <fixture 'p2pds'> cancelled the test but didn't raise an error
```

Digging a little deeper, the crash has something to do with this [workaround](https://github.com/libp2p/py-libp2p/blob/9d68de8c21a9d11e40ff253698715ed63b749d9f/tests_interop/conftest.py#L132-L138), which makes use of the Python daemon bindings, which I suspect is expecting an older p2pd version.

It looks like I may not be able to get this workaround working without making some changes to the Python bindings, so I'm inclined to just leave these tests out. I'd be happy to investigate further if someone has any suggestions.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


g-r-a-n-t:
Hey @mhchia, are you available to take a look at this?

mhchia:
@g-r-a-n-t Sure. Sorry I can't review it right now, but I can do it in the next few days.

g-r-a-n-t:
No rush, thank you very much!

mhchia:
>  removed prefixes on noise handshake payloads, it looks like this was not desired in the go implementation and has since been removed.

Do I understand it correctly that it's the prefixes that make py-libp2p unable to talk with go and rust, and removing the prefixes makes it interoperable again?

g-r-a-n-t:
That's correct. I did a manual test with Rust too, it worked after removing this prefix. Judging by this [comment](https://github.com/libp2p/py-libp2p/blob/36a4a9150dcc53b42315b5c6868fccde5083963b/libp2p/security/noise/io.py#L35-L36), I figured the prefix was added to accommodate unspecified behavior in the Go impl.

mhchia:
> That's correct. I did a manual test with Rust too, it worked after removing this prefix. Judging by this comment, I figured the prefix was added to accommodate unspecified behavior in the Go impl.

Got it. It's nice that they had this fixed.

mhchia (approved):
Awesome work! I didn't dig into why the skipped tests(i.e. `test_net_stream_read_after_remote_reset` and `test_net_stream_write_after_remote_reset`) fail but they should not be a problem IMO.

Just a reference: I suggested that there is no way for users to reset opened streams previously: https://github.com/libp2p/go-libp2p-daemon/issues/141 . We can enable these tests after they support this feature someday.

PR #425, opened 2020-08-27
Fix broken release notes link

MohamedAlFahim:


McSinyx:
The check seems to failed because of unrelated reasons, I think this can be merged.

mhchia:
Thanks for the contribution!

PR #426, opened 2021-02-23
Use trio.lowlevel instead of trio.hazmat

McSinyx:
Since trio 0.15.0, hazmat has been deprecated (python-trio/trio#476).  trio-typing and mypy are bumped to support newer trio and each other.

I suppose that this needs not a news fragment, but feel free to tell me to add one.

### Cute Animal Motion Picture

https://user-images.githubusercontent.com/13689192/108862799-29135d80-7623-11eb-93e4-57f0f8776886.mp4


mhchia (approved):
Looks good to me! Thanks for the contribution.

Issue #428, opened 2021-03-12
Signature verification using Ed25519 public key fails


wolfgang:
## What was wrong?
I expected to verify a signature using a Ed25519 public key, but that threw an exception.

### Code that produced the error

```py
from libp2p import crypto
from base64 import b64decode 

public_key_b64 = 'CAESIIsxrBKQAeAtxOF9U5RBGG53bN15kHwUVyldC0PaPFIT'
payload_b64 = 'c29tZSBzdHJpbmc='
signature_b64 = 'YXnK4XqfaIjsZLpUhTv15+fBjUHdKYw95L7eWAziQfH22h8fDpr6lgqAs5xOZNo9fnuyA0rldk1XhxtdHpGLBQ=='

public_key_data = b64decode(public_key_b64)
public_key = crypto.serialization.deserialize_public_key(public_key_data);
payload = b64decode(payload_b64)
signature = b64decode(signature_b64)

result = public_key.verify(payload, signature)
print("verify result", result)

```

### Full error output 

```sh
Traceback (most recent call last):
  File "src/sandbox/libp2p-error.py", line 14, in <module>
    result = public_key.verify(payload, signature)
  File "/Users/wodev/miniconda3/envs/ipfs-3.7/lib/python3.7/site-packages/libp2p/crypto/ed25519.py", line 29, in verify
    verify_key.verify(h, signature)
  File "/Users/wodev/miniconda3/envs/ipfs-3.7/lib/python3.7/site-packages/nacl/signing.py", line 107, in verify
    smessage = signature + smessage
TypeError: can't concat SHA256Hash to bytes

```

### Expected Result
```sh
verify result True
``` 

### Environment

```sh
Python version:
3.7.6 (default, Jan  8 2020, 13:42:34) 
[Clang 4.0.1 (tags/RELEASE_401/final)]

Operating System: Darwin-19.6.0-x86_64-i386-64bit

pip freeze result:
aiohttp==3.6.2
aioipfs-api==0.1.3
asn1crypto==1.4.0
async-exit-stack==1.0.1
async-generator==1.10
async-service==0.1.0a11
async-timeout==3.0.1
attrs==19.3.0
base58==1.0.3
certifi==2020.4.5.1
cffi==1.14.5
chardet==3.0.4
coincurve==10.0.0
cryptography==3.4.6
cytoolz==0.11.0
docopt==0.6.2
eth-hash==0.3.1
eth-typing==2.2.2
eth-utils==1.10.0
[...]

PR #431, opened 2021-03-16
Don't hash data passed to nacl library in Ed25519PublicKey.verify

wolfgang:
## What was wrong?

Issue #428 

## How was it fixed?

Don't hash the data parameter passed to nacl library, as it expects both inputs to be bytes.


mhchia:
Thanks for fixing!:)

PR #453, opened 2024-04-08
Flesh out docs

pacrob:
## What was wrong?

Docs didn't contain much useful info, and too much was crammed into the main README.

 - Moved contributing information in the docs Contributing section.

### To-Do

- [ ] Clean up commit history

* [ ] Add entry to the [release notes](https://github.com/libp2p/py-libp2p/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -1,4 +1,51 @@
 Contributing
 ============
 
-TODO - instructions for contributing
+Development
+-----------
+
+py-libp2p requires Python 3.8+.
+
+To get started, fork the repository to your own github account, then clone it to your
+development machine:
+
+```sh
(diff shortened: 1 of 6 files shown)


docs/contributing.rst
```diff
@@ -1,4 +1,51 @@
...
+```
+
+then install the development dependencies. We recommend using a virtual environment,
+such as [`virtualenv`](https://virtualenv.pypa.io/en/stable/)
+
+```sh
+cd py-libp2p
+virtualenv -p python venv
+. venv/bin/activate
+python -m pip install -e ".[dev]"
```
pacrob, line 23: Adding the double-quotes lets it work in bash or zsh

pacrob:
All this and more in #454

PR #480, opened 2024-10-25
added PingService

mystical-prog:
## Ping Protocol Implementation

Issue #344 

## How was it fixed?

I have created a new PingService class inside the [`libp2p/host/ping.py`](https://github.com/libp2p/py-libp2p/blob/main/libp2p/host/ping.py) file

### To-Do

- [ ] Write tests for the new `PingService` implementation
- [ ] Run make lint

#### Cute Animal Picture

### DIFF
--- a/libp2p/host/ping.py
+++ b/libp2p/host/ping.py
@@ -1,6 +1,7 @@
 import logging
-
+import time
 import trio
+import secrets
 
 from libp2p.network.stream.exceptions import (
     StreamClosed,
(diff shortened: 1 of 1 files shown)


kazzmir:
tangentially related, lint errors in the ping.py example
```
examples/ping/ping.py:29:24: E711 comparison to None should be 'if cond is not None:'
examples/ping/ping.py:35:9: B001 Do not use bare `except:`, it also catches unexpected events like memory errors, interrupts, system exit, and so on.  Prefer `except Exception:`.  If you're sure what you're doing, be explicit and write `except BaseException:`.
examples/ping/ping.py:35:9: E722 do not use bare 'except'
```

pacrob (approved):
looks good!

PR #481, opened 2024-10-25
replace display_version with two other options

kazzmir:
## What was wrong?

html build uses a deprecated sphinx_rtd_theme option: `display_version`
```
$ make html
preparing documents... WARNING: unsupported theme option 'display_version' given
```

## How was it fixed?

Replace `display_version` in conf.py with `language_selector` and `version_selector`, both set to false.

https://sphinx-rtd-theme.readthedocs.io/en/stable/configuring.html

### To-Do

- [ ] Clean up commit history

* [ ] Add entry to the [release notes](https://github.com/libp2p/py-libp2p/blob/main/newsfragments/README.md)

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -17,5 +17,6 @@ python:
         - docs
         - test
 
-# Build all formats for RTD Downloads - htmlzip, pdf, epub
-formats: all
+formats:
+  - epub
+  - htmlzip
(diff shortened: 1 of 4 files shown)


pacrob:
Thanks, @kazzmir ! I threw in dropping the pdf docs build - I think it broke a while ago, doesn't really seem necessary.

PR #482, opened 2024-10-26
Fixed Failing Tests due to RaisesGroup of Trio Package

mystical-prog:
## Issue with `RaisesGroup`

The failed tests were a result of deprecated keyword `strict` for `trio.testing.RaisesGroup`.

## How was it fixed?

I updated the keywords from `strict` to `allow_unwrapped` and `flatten_subgroups`, and assigning value `True` to both, this fixed the failing tests but `protobuf` warnings still remain but it doesn't break the tests.

#### Cute Animal Picture

### DIFF
--- a/newsfragments/482.internal.rst
+++ b/newsfragments/482.internal.rst
@@ -1,0 +1,1 @@
+Update ``trio`` package version dependency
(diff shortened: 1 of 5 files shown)


pacrob:
Looks good, @mystical-prog ! I just merged @kazzmir 's pr to fix the docs issue, so you if you rebase off `main` you should see those pass.

It looks like `trio` made the change in `0.26.0`, so go ahead and move the pin in `setup.py` to `>=0.26.0`. Add a newsfragment, and it should be good to go!

mystical-prog:
I have made the necessary updates @pacrob 
Let me know if there is anything else

PR #484, opened 2024-11-06
reject messages claiming to be from ourselves but not locally published

### LINKED ISSUE #413: Pubsub should not return messages from local peer.
It looks like pubsub w/ gossipsub router yields messages we publish on the local peer to the local peer. This seems to be because we are using the same infrastructure to publish data as we are to read incoming data.

In particular, review use of `self.subscribed_topics_receive[topic_id]`. It seems like this is the only way right now to get gossip messages from others, but given the call to notify subscriptions in `push_msg`, we end up getting a reference to the message back locally.

mystical-prog:
## What was wrong?

There was a missing check in `push_msg` function inside the `pubsub` module wherein we weren't rejecting messages claiming to be from ourselves but not locally published

Closes #413 

## How was it fixed?

I took a reference from [`go-libp2p-pubsub`](https://github.com/libp2p/go-libp2p-pubsub/blob/c06df2f9a38e9382e644b241adf0e96e5ca00955/pubsub.go#L1154) implementation and added the missing check

### To-Do

- [X] Ran `tests`
- [X] Ran `make lint`
- [X] Added `newsfragment`

#### Cute Animal Picture

### DIFF
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -643,6 +643,19 @@ class Pubsub(Service, IPubsub):
             return
 
         self._mark_msg_seen(msg)
+
+        # reject messages claiming to be from ourselves but not locally published
+        self_id = self.host.get_id()
+        if (
+            base58.b58encode(msg.from_id).decode() == self_id
+            and msg_forwarder != self_id
+        ):
+            logger.debug(
+                "dropping message claiming to be from self but forwarded from %s",
+                msg_forwarder,
+            )
+            return
+
         self.notify_subscriptions(msg)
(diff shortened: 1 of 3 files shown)


pacrob:
Thanks, @mystical-prog ! Would it be possible to add a test showing that such a message is properly dropped?

mystical-prog:
Sure, I will add a test and get back to you.

mystical-prog:
hey @pacrob 
I have added the required testcase to an existing flow - let me know if there are any changes

pacrob:
Looks good, @mystical-prog ! Thanks!

PR #479, opened 2024-10-24
WIP - no duplicate addresses in peerdata

pacrob:
## What was wrong?

Updated from PR #325

## How was it fixed?

Summary of approach.

### To-Do

- [ ] Clean up commit history

* [ ] Add entry to the [release notes](https://github.com/libp2p/py-libp2p/blob/main/newsfragments/README.md)

#### Cute Animal Picture

image

PR #485, opened 2024-11-21
No Duplicate Peers in Peerdata

mystical-prog:
## What was wrong?

The `add_addrs` function lacked checking if the address already exists in the peerdata, necessary tests have been added.

updated from #325 & #479 

#### Cute Animal Picture

### DIFF
--- a/libp2p/peer/peerdata.py
+++ b/libp2p/peer/peerdata.py
@@ -55,7 +55,9 @@ class PeerData(IPeerData):
         """
         :param addrs: multiaddresses to add
         """
-        self.addrs.extend(addrs)
+        for addr in addrs:
+            if addr not in self.addrs:
+                self.addrs.append(addr)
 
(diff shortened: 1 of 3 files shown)


pacrob:
Just a note on the tests pulled from #479 - we should be using pytest, no unittest. That's my mistake.

mystical-prog:
Yes, I also found it weird but I thought it was intentional, let me update this quickly.

mystical-prog:
> Just a note on the tests pulled from #479 - we should be using pytest, no unittest. That's my mistake.

I have made the necessary changes, please review and let me know @pacrob

pacrob:
Looks good, thanks @mystical-prog !

moshemalawach:
It took 5 years but finally!

mystical-prog:
Better late then never 😅

PR #486, opened 2024-11-30
Fixed Doc Comments

mystical-prog:
## What was wrong?

There were some missing comments for params in `abc.py` under the pubsub folder.
There was an outdated TODO which was to check whether the interface was correct or not, I reviewed the usage of `handle_rpc` function in both `py-libp2p` and `go-libp2p`, the `sender_peer_id` is a necessary param which allows us to keep context for who sent this message to us. Thus, the TODO is resolved and removed.

I have added the newsfragment for the same.

#### Cute Animal Picture


pacrob:
Looks great, @mystical-prog! I just updated the docstrings to match across all instances. Thank you!

PR #498, opened 2025-01-25
update Feature Breakdown to more closely match official implementations list of features

pacrob:
## What was wrong?

List of features in README had a number of `unplanned` features that were not present in the libp2p implementations list - https://libp2p.io/implementations/

## How was it fixed?

Updated the list items and statuses to more closely match.

Add @seetadev to the current maintainers too.

Summary of approach.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,102 +21,100 @@ Read more in the [documentation on ReadTheDocs](https://py-libp2p.readthedocs.io
 
 ## Maintainers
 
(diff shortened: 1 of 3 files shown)


pacrob:
@dhuseby I've tried to clean up the Features list to more closely resemble the list at libp2p.io, but wasn't sure where the ones I've left at the bottom fit - the `Protocol Muxers` and `Switch (Swarm)` sections. Are those non-standard, or are they under different names in the groups above?

PR #514, opened 2025-02-23
fix: broken import in the examples after net_stream_interface rename

### LINKED ISSUE #513: Fix broken import in the examples after net_stream_interface rename
## Description:
The examples (`examples/ping/ping.py`, `examples/chat/chat.py` and `examples/echo/echo.py`) is currently broken due to an import path change from `libp2p.network.stream.net_stream_interface` to `libp2p.network.stream.net_stream`. When running `python -m examples.ping.ping`, users get this error:

```python
ModuleNotFoundError: No module named 'libp2p.network.stream.net_stream_interface'
```

This appears to be due to a module rename/restructure where the interface was moved from net_stream_interface.py to net_stream.py, but the example code wasn't updated to reflect this change.

paschal533:
## Description:
 
The examples was broken due to an old import path that no longer exists. This PR:

- Updates the import statement in examples to use the new path
- Adds a newsfragment documenting the fix

The examples now runs correctly without the ModuleNotFoundError.

Fixes #513


pacrob (approved):
Looks great, thanks @paschal533 !

Issue #526, opened 2025-03-02
py-ibp2p vulnerable to attack using large RSA keys（vulnerability）


scacaca:
### Summary

There is no length limit for RSA keys, and a malicious node can exploit large RSA keys to launch a resource exhaustion attack.
A malicious peer can use large RSA keys to run a resource exhaustion attack & force a node to spend time doing signature verification of the large key.
there is a similar [problem](https://github.com/advisories/GHSA-876p-8259-xjgg)
The relevant codes are in [file1](https://github.com/libp2p/py-libp2p/blob/main/libp2p/crypto/rsa.py)(RSA key generation and parsing functions）

### Expected behavior

refuse large rsa key

### Actual behavior

There is no length limit for RSA keys.

### Relevant log output

```shell

```

### Possible Solution

The vulnerability can be fixed by restricting the length of RSA keys. it is similar to [this](https://github.com/libp2p/go-libp2p/pull/2454/files).
The fix could be to limit the RSAkeys length.

### Environment

```shell

```

### Would you like to work on fixing this bug ?

Yes
# Credit
The vulnerability was discovered by chaojie Xiong (FUDAN University, sslwork2023@163.com) ,mingda Guo(FUDAN University,22110240064@m.fudan.edu.cn) ,zhang Lei(FUDAN University,zxl@fudan.edu.cn),DongLai Zhu(FUDAN University) YangChao Liu（FUDAN University,23210860056@m.fudan.edu.cn).

Khwahish29:
Hey! I'd like to work on this issue

scacaca:
thank you for your fix

pacrob:
Closed by #531

PR #531, opened 2025-03-05
Rufuse large RSA keys

Khwahish29:
## What was wrong?
There is no length limit for RSA keys, which allows a malicious peer to generate excessively large RSA keys and exploit this to launch a resource exhaustion attack. This forces nodes to spend unnecessary time verifying large signatures, potentially degrading network performance.
Issue #526 

## How was it fixed?
The fix involves imposing a limit on the maximum allowed RSA key length when generating or parsing RSA keys. This prevents the possibility of abuse by enforcing a reasonable upper bound of 4KB on key size.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses:max_bytes(150000):strip_icc()/GettyImages-185743593-2000-507c6c8883a44851885ea4fbc10a2c9e.jpg)

### DIFF
--- a/libp2p/crypto/rsa.py
+++ b/libp2p/crypto/rsa.py
@@ -9,6 +9,9 @@ from Crypto.Signature import (
     pkcs1_15,
 )
 
+from libp2p.crypto.exceptions import (
+    CryptographyError,
+)
 from libp2p.crypto.keys import (
     KeyPair,
     KeyType,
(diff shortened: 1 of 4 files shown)


pacrob:
Thanks, @Khwahish29! Couple thoughts:

 - Can we make it so we only need to have that error message once? That is, allow the validation function to be used in every case where we want to validate the size?
 - While we're validating, we might as well check other cases like negative numbers.
 - We'll need some tests that fail if the key size is invalid.

pacrob:
Looks great, thanks!

Issue #589, opened 2025-05-08
Fix check-rst-files pre-commit hook to work cross-platform


paschal533:
Currently, the `check-rst-files` hook uses `sh -c '...'` which fails on Windows since Windows doesn't have `sh` available by default.

This issue causes pre-commit to fail on Windows with:

```
Executable sh not found
```

Let's modify the hook to use Python instead of shell commands, making it work consistently across all platforms.

PR #590, opened 2025-05-08
Fix check-rst-files hook to work cross-platform

paschal533:
## What was wrong?

Currently, the `check-rst-files` hook uses `sh -c '...'` which fails on Windows since Windows doesn't have `sh` available by default.

Issue #589

## How was it fixed?

This PR fixes the `check-rst-files` pre-commit hook to work on all platforms, including Windows.

The issue was that the previous implementation used `sh` which isn't available by default on Windows systems. The new implementation uses Python instead, which works everywhere pre-commit runs.

The functionality remains exactly the same - it checks for .rst files in the top-level directory and fails if it finds any.

Summary of approach.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,7 +55,7 @@ repos:
     hooks:
     -   id: check-rst-files
         name: Check for .rst files in the top-level directory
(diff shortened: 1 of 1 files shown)


pacrob (approved):
works for me!

Issue #412, opened 2020-05-13
Upgrade `gossipsub` to v1.1


ralexstokes:
Spec: https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/README.md

Ref. impl.: https://github.com/libp2p/go-libp2p-pubsub

ralexstokes:
v1.1 spec: https://github.com/libp2p/specs/blob/master/pubsub/gossipsub/gossipsub-v1.1.md

mystical-prog:
@pacrob I will like to work on this issue and make PRs on pragmatic basis

mystical-prog:
As I am working towards upgrading the Gossipsub to V1.1 but along the way I am also trying to fill the gaps in existing v1.0 and Pubsub, for this I am taking reference from go-libp2p and libp2p specs.

### TODOs

- [ ] Implement `seen_ttl` for seen_messages cache in Pubsub Router
- [ ] Add optional `D_LAZY` param coming from gossipsub v1.0

I will keep on updating and adding to this list

mystical-prog:
I have broken down the entire upgrade process into multiple steps -

- [X] Direct Peers
- [x] Peer Exchange & Prune Back Off
- [ ] Activation Function
- [ ] Peer Scoring


Direct Peers PR has been merged and currently I am working towards Peer Exchange & Prune Back Off, while going through the go-libp2p implementation I have found that py-libp2p doesn't have CertifiedAddrBook implementation in PeerStore.

seetadev:
@mystical-prog : Hi Vraj. 

Wish to also share to keep in sync with Gossipsub interop efforts, which are nearing production at https://github.com/libp2p/test-plans/pull/648.  Please also visit https://github.com/ppopth/pubsub-shadow/tree/main .

PR #594, opened 2025-05-11
Direct Peers : Gossipsub V1.1

mystical-prog:
## What was wrong?

This PR is part of a long working Gossipsub upgrade process, I have added Direct Peers (AKA Explicit Peers) in this contribution. There are 4 major changes that need to be implemented to introduce the complete functionality of Direct Peering -

- [x] adding direct peers to peerstore while invoking router.attach method
- [x] including direct peers in peers_gen while invoking router.publish method
- [x] rejecting GRAFT requests from direct peers and responding with an appropriate PRUNE msg
- [x] running a direct peer heartbeat which constantly checks for connections with direct peers
- [x] test file to cover all possible cases of Direct Peering

part of #412 

### To-Do

- [X] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/libp2p/pubsub/gossipsub.py
+++ b/libp2p/pubsub/gossipsub.py
@@ -7,6 +7,7 @@ from collections import (
 from collections.abc import (
     Iterable,
     Sequence,
+    Set,
 )
 import logging
 import random
@@ -28,6 +29,12 @@ from libp2p.network.stream.exceptions import (
 )
 from libp2p.peer.id import (
     ID,
+)
+from libp2p.peer.peerinfo import (
+    PeerInfo,
+)
+from libp2p.peer.peerstore import (
+    PERMANENT_ADDR_TTL,
 )
 from libp2p.pubsub import (
     floodsub,
@@ -82,17 +89,24 @@ class GossipSub(IPubsubRouter, Service):
     heartbeat_initial_delay: float
     heartbeat_interval: int
 
+    direct_peers: Set[PeerInfo]
+    direct_connect_initial_delay: int
+    direct_connect_interval: int
+
     def __init__(
         self,
         protocols: Sequence[TProtocol],
         degree: int,
         degree_low: int,
         degree_high: int,
+        direct_peers: Set[PeerInfo],
         time_to_live: int = 60,
         gossip_window: int = 3,
         gossip_history: int = 5,
         heartbeat_initial_delay: float = 0.1,
         heartbeat_interval: int = 120,
+        direct_connect_initial_delay: int = 1,
+        direct_connect_interval: int = 300,
     ) -> None:
         self.protocols = list(protocols)
         self.pubsub = None
@@ -119,6 +133,11 @@ class GossipSub(IPubsubRouter, Service):
         self.heartbeat_initial_delay = heartbeat_initial_delay
         self.heartbeat_interval = heartbeat_interval
 
+        # Create direct peers
+        self.direct_peers = direct_peers
+        self.direct_connect_interval = direct_connect_interval
+        self.direct_connect_initial_delay = direct_connect_initial_delay
+
     async def run(self) -> None:
         if self.pubsub is None:
             raise NoPubsubAttached
@@ -141,6 +160,12 @@ class GossipSub(IPubsubRouter, Service):
         :param pubsub: pubsub instance to attach to
         """
         self.pubsub = pubsub
+
+        if len(self.direct_peers) > 0:
+            for pi in self.direct_peers:
+                self.pubsub.host.get_peerstore().add_addrs(
+                    pi.peer_id, pi.addrs, PERMANENT_ADDR_TTL
+                )
 
         logger.debug("attached to pusub")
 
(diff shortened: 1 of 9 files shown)


mystical-prog:
hey @seetadev @pacrob 

I would like to give an update on this, me and @Khwahish29  connected today and I have give her the full context of what I am trying to achieve in this PR, she is helping in trying to write test cases for the code changes I have introduced and we plan to close this before this weekend.

seetadev:
> hey @seetadev @pacrob
> 
> I would like to give an update on this, me and @Khwahish29 connected today and I have give her the full context of what I am trying to achieve in this PR, she is helping in trying to write test cases for the code changes I have introduced and we plan to close this before this weekend.

@mystical-prog : Great to hear. Thank you so much for the update. Appreciate the great contribution by both you and @Khwahish29 .

mystical-prog:
Hey @pacrob 

I think this PR is ready to be reviewed now !

libp2p/pubsub/gossipsub.py
```diff
@@ -82,17 +88,24 @@ class GossipSub(IPubsubRouter, Service):
...
+    direct_connect_initial_delay: float
+    direct_connect_interval: int
+
     def __init__(
         self,
         protocols: Sequence[TProtocol],
         degree: int,
         degree_low: int,
         degree_high: int,
+        direct_peers: Sequence[PeerInfo],
```
pacrob, line 101: Should `direct_peers` be required? Or can we give it a default?

libp2p/tools/constants.py
```diff
@@ -26,11 +32,14 @@ class GossipsubParams(NamedTuple):
     degree: int = 10
     degree_low: int = 9
     degree_high: int = 11
+    direct_peers: Sequence[PeerInfo] = list()
     time_to_live: int = 30
     gossip_window: int = 3
     gossip_history: int = 5
     heartbeat_initial_delay: float = 0.1
     heartbeat_interval: float = 0.5
+    direct_connect_init_delay: float = 0.1
```
pacrob, line 41: In `GossipSub.__init__`, you use `initial`, here you have `init` - probably good to stick to one.

libp2p/tools/constants.py
```diff
@@ -26,11 +32,14 @@ class GossipsubParams(NamedTuple):
     degree: int = 10
     degree_low: int = 9
     degree_high: int = 11
+    direct_peers: Sequence[PeerInfo] = list()
```
pacrob, line 35: You don't want to use a mutable object as a default here, otherwise every instance will share the same list. I'm not sure there's a good way to do with with `NamedTuple` - maybe a `dataclass`?

libp2p/pubsub/gossipsub.py
```diff
@@ -654,6 +704,14 @@ async def handle_graft(
 
         # Add peer to mesh for topic
         if topic in self.mesh:
+            for _peer in self.direct_peers:
```
pacrob, line 707: A `_`-prefixed variable means that it isn't used, but it is directly used in the next line.

libp2p/pubsub/gossipsub.py
```diff
@@ -425,6 +457,24 @@ async def heartbeat(self) -> None:
 
             await trio.sleep(self.heartbeat_interval)
 
+    async def direct_connect_heartbeat(self) -> None:
+        """
+        Connect to direct peers.
+        """
+        await trio.sleep(self.direct_connect_initial_delay)
+        while True:
+            for _peer in self.direct_peers:
```
pacrob, line 466: Same comment, should not use `_` prefix because the variable is used.

pacrob:
Looks good! Just a few comments to clean up, then good to go!

libp2p/pubsub/gossipsub.py
mystical-prog: I have set the default value as None and added logic to handle it

libp2p/tools/constants.py
mystical-prog: Since, the default value is None now, I can simply put in None as input instead of any other class or data structure

libp2p/tools/constants.py
mystical-prog: I will stick with init since if I go with initial then lint check fails as accepted length of a line is only 88 characters

mystical-prog:
I have made the suggested changes @pacrob 
Let me know if something else seems off or needs to be changed.

libp2p/tools/constants.py
pacrob: It's ok have a longer line, you can use `# noqa: E501` at the end to have lint ignore it. Better that than have 2 different variable names for the same thing.

mystical-prog:
I have made the final changes but I guess CI is blocked due to too many invocations on this PR

pacrob:
> I have made the final changes but I guess CI is blocked due to too many invocations on this PR

I see it, I'll give it a bit more time, but will get merged today.

pacrob (approved):

PR #617, opened 2025-05-18
Feature/kademlia dht

acul71:
## What was wrong?
doc issues in `kad_dht.py` and `kad_dht/peer_routing.py`

## How was it fixed?

Just added `-`


acul71:
Didn't know how to make https://github.com/libp2p/py-libp2p/pull/617/commits/ab3c88c633c372bf3f29835ff72dbec49902d20d available for Sumanjeet so I did like this

Issue #619, opened 2025-05-19
Store pubkey and peer_id to PeerStore


sukhman-sukh:
### Description

Currently, the system does not store the public key (pubkey) and peer_id in the PeerStore. This enhancement will ensure that both values are properly saved to the PeerStore whenever a new peer is added or updated. 

### Motivation

This is important for peer identification and secure communication within the network. 

### Current Implementation

This requires changes in `security/insecure/transport.py` file

### Are you planning to do it yourself in a pull request ?

Yes

sukhman-sukh:
@seetadev @acul71  Please assign me this issue.

PR #631, opened 2025-05-29
feat: store pubkey and peerid in peerstore

Winter-Soren:
## What was wrong?
The system was not storing the public key (pubkey) and peer_id in the PeerStore during the handshake process.

Issue #619 

## How was it fixed?
1. Modified `InsecureTransport` class to accept an optional `peerstore` parameter
2. Updated the `run_handshake` function to store the received public key and peer ID in the peerstore if available
3. Updated all relevant code to pass the peerstore to the InsecureTransport constructor:
   - Updated `libp2p/__init__.py` to pass peerstore_opt
   - Added proper type annotations to ensure type safety
   - Updated test files to properly handle the new parameter

### To-Do
- [x] Clean up commit history

#### Cute Animal Picture
image


seetadev:
@Winter-Soren : This is indeed an important improvement for peer tracking and validation. Could you please share a small demo or a set of screenshots that demonstrate the handshake flow changes — ensuring pubkey and peer_id are now reliably stored in the PeerStore. 

Like the approach of making peerstore optional in InsecureTransport — keeps flexibility while allowing proper state capture when needed.

pacrob:
Looks like a good start, @Winter-Soren. I'd like a newsfragment and a specific test for this, but then good to go.

For the test, I'm looking for something that would fail without these updates, but passes with them, if that helps.

seetadev:
@pacrob : Thank you Paul for the feedback and pointers. Appreciate it.

@Winter-Soren has now added both the newsfragment and a targeted test that confirms the fix — exactly in line with your suggestion. The test does fail without the update and passes with it, so we’re covered on that front.

Wish to share that I did discuss with @Winter-Soren that he could add a third test that checks whether the peerstore remains unchanged if the handshake fails or if the peer ID doesn't match expectations. This enables edge behavior coverage.

Looks good to go from my side — appreciate the thoughtful review as always!

pacrob:
> he could add a third test that checks whether the peerstore remains unchanged if the handshake fails or if the peer ID doesn't match expectations. This enables edge behavior coverage.

PR does look good, but I'm always a fan of more testing! If you have another case in mind, @Winter-Soren, go ahead and add it. I'll merge once it's in. Thanks!

Winter-Soren:
> > he could add a third test that checks whether the peerstore remains unchanged if the handshake fails or if the peer ID doesn't match expectations. This enables edge behavior coverage.
> 
> PR does look good, but I'm always a fan of more testing! If you have another case in mind, @Winter-Soren, go ahead and add it. I'll merge once it's in. Thanks!

@pacrob @seetadev I have incorporated a test case to verify that the peer store remains unchanged when the handshake fails. I’ve also added a test case from a security perspective to ensure that if a peer ID already exists in the peer store (but without a public key), the handshake correctly adds the missing public key to that entry.

seetadev:
> > > he could add a third test that checks whether the peerstore remains unchanged if the handshake fails or if the peer ID doesn't match expectations. This enables edge behavior coverage.
> > 
> > 
> > PR does look good, but I'm always a fan of more testing! If you have another case in mind, @Winter-Soren, go ahead and add it. I'll merge once it's in. Thanks!
> 
> @pacrob @seetadev I have incorporated a test case to verify that the peer store remains unchanged when the handshake fails. I’ve also added a test case from a security perspective to ensure that if a peer ID already exists in the peer store (but without a public key), the handshake correctly adds the missing public key to that entry.

@Winter-Soren : This looks great — thanks for taking the time to add those thorough test cases. 🙌

Appreciate the attention to detail here: verifying that the peer store remains unchanged on handshake failure helps ensure stability, and the security-focused test for backfilling the missing public key is a solid addition. These are exactly the kinds of edge cases that strengthen the robustness of the handshake logic.

Everything looks good to merge — nice work.

Issue #633, opened 2025-05-30
Add Support for mDNS-Based Local Peer Discovery in py-libp2p


seetadev:
### Description

The `py-libp2p` implementation currently does not support [Multicast DNS (mDNS)](https://datatracker.ietf.org/doc/html/rfc6762) -based peer discovery, which is a key feature in other libp2p implementations like [go-libp2p](https://github.com/libp2p/go-libp2p) and [js-libp2p](https://github.com/libp2p/js-libp2p) mDNS enables automatic peer discovery on local networks without the need for manually configured bootstrap nodes or DHT lookups.

This feature is especially useful in LAN environments, local-first applications, peer-to-peer IoT deployments, and offline-friendly use cases. For example, developers working on WebRTC-based peer connections (e.g., [libp2p/js-libp2p-example-webrtc-private-to-private](https://github.com/libp2p/js-libp2p-example-webrtc-private-to-private)) often rely on mDNS to bootstrap initial peer connections.

Implementing mDNS discovery in `py-libp2p` will:

* Bring feature parity with other language implementations.
* Allow `py-libp2p` nodes to be discovered automatically on local networks.
* Improve interoperability in mixed-language libp2p applications.

This issue proposes the development of an `mdns-discovery` module within `py-libp2p`, aligned with the libp2p [discovery spec](https://github.com/libp2p/specs/blob/master/discovery/mdns.md), and integrated with the Host or Swarm behavior. Ideally, it should emit discovered peers via events or callback hooks for seamless integration into higher-level applications.

### Motivation

Peer discovery is a fundamental part of libp2p’s modular networking stack. In many environments—particularly those involving local networks (e.g. home/office LANs, peer-to-peer demos, mesh networks, IoT setups, or classroom settings)—manual bootstrapping or use of a public DHT is impractical or impossible due to:

* Lack of external connectivity (offline-first or air-gapped environments),
* Privacy concerns (avoiding public exposure via bootstrapping or DHT),
* Need for frictionless setup (developer experience, demo readiness).

**Multicast DNS (mDNS)** enables libp2p nodes to automatically discover each other on the same local network, without relying on any external infrastructure. This functionality is already provided in both `go-libp2p` and `js-libp2p` through mature mDNS modules.

However, the **lack of mDNS in `py-libp2p` creates a serious gap**:

* Python-based libp2p nodes **cannot auto-discover peers** on a local network.
[...]

PR #649, opened 2025-06-01
feat: Implement Multicast DNS

sumanjeet0012:
## What was wrong?

Py-libp2p does not have mDNS support.

Issue #633 

## How was it fixed?

The implementation of Multicast DNS is still in progress.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/libp2p/__init__.py
+++ b/libp2p/__init__.py
@@ -71,6 +71,9 @@ from libp2p.transport.upgrader import (
 from libp2p.utils.logging import (
     setup_logging,
 )
+from libp2p.discovery.mdns.mdns import (
+    MDNSDiscovery
+)
 
 # Initialize logging configuration
 setup_logging()
@@ -243,6 +246,7 @@ def new_host(
     disc_opt: Optional[IPeerRouting] = None,
     muxer_preference: Optional[Literal["YAMUX", "MPLEX"]] = None,
     listen_addrs: Sequence[multiaddr.Multiaddr] = None,
+    enable_mDNS: bool = False,
 ) -> IHost:
     """
     Create a new libp2p host based on the given parameters.
@@ -266,8 +270,15 @@ def new_host(
     )
 
     if disc_opt is not None:
-        return RoutedHost(swarm, disc_opt)
-    return BasicHost(swarm)
+        host = RoutedHost(swarm, disc_opt)
+    else:
+        host = BasicHost(swarm)
 
+    if enable_mDNS:
+        mdns = MDNSDiscovery(swarm)
+        mdns.start()
+        host._mdns = mdns 
+        
+    return host
 
 __version__ = __version("libp2p")
--- a/examples/discovery/mDNS/mDNS.py
+++ b/examples/discovery/mDNS/mDNS.py
@@ -1,0 +1,36 @@
+import secrets
+import multiaddr
+import trio
+
+from libp2p import new_host
+from libp2p.crypto.secp256k1 import create_new_key_pair
+
+async def main():
+    # Generate a key pair for the host
+    secret = secrets.token_bytes(32)
+    key_pair = create_new_key_pair(secret)
+
+    # Listen on a random TCP port
+    listen_addr = multiaddr.Multiaddr("/ip4/0.0.0.0/tcp/0")
+
+    # Enable mDNS discovery
+    host = new_host(key_pair=key_pair, enable_mDNS=True)
+
+    async with host.run(listen_addrs=[listen_addr]):
+        print("Host started!")
+        print("Peer ID:", host.get_id())
+        print("Listening on:", [str(addr) for addr in host.get_addrs()])
+
+        # Print discovered peers via mDNS
+        print("Waiting for mDNS peer discovery events (Ctrl+C to exit)...")
+        try:
+            while True:
+                # Print all known peers every 5 seconds
+                peers = host.get_peerstore().peer_ids()
+                print("Known peers:", [str(p) for p in peers if p != host.get_id()])
+                await trio.sleep(5)
+        except KeyboardInterrupt:
+            print("Exiting...")
+
+if __name__ == "__main__":
+    trio.run(main)
--- a/libp2p/discovery/mdns/mdns.py
+++ b/libp2p/discovery/mdns/mdns.py
@@ -1,0 +1,85 @@
+"""
+mDNS-based peer discovery for py-libp2p.
+Conforms to https://github.com/libp2p/specs/blob/master/discovery/mdns.md
+Uses zeroconf for mDNS broadcast/listen. Async operations use trio.
+"""
+import trio
+from typing import Callable, Optional
+from zeroconf import ServiceInfo, Zeroconf, ServiceBrowser, ServiceStateChange
+from .utils import (
+    stringGen
+)
+from libp2p.abc import (
+    INetworkService
+)
+
+SERVICE_TYPE = "_p2p._udp.local."
+MCAST_PORT = 5353
+MCAST_ADDR = "224.0.0.251"
+
+class MDNSDiscovery:
+    def __init__(self, swarm: INetworkService):
+        self.peer_id = swarm.get_peer_id()
+        self.port = 8000  # Default port, can be overridden
+        # self.broadcast = init.get('broadcast', True) is not False
+        # self.on_peer = on_peer  # Callback: async def on_peer(peer_info)
+        # self.service_name = service_name or f"{peer_id}.{SERVICE_TYPE}"
+        # self.zeroconf = Zeroconf()
+        # self._service_info = ServiceInfo(
+        #     SERVICE_TYPE,
+        #     self.service_name,
+        #     addresses=[],  # Will be set on register
+        #     port=self.port,
+        #     properties={b'id': peer_id.encode()},
+        # )
+        # self._browser = None
+        # self._running = False
+
+    def main(self) -> None:
+        """
+        Main entry point for the mDNS discovery service.
+        This method is intended to be run in an event loop.
+        """
+        trio.run(self.start)
+
+    async def start(self):
+        await trio.sleep(10)
+        # self._running = True
+        # await trio.to_thread.run_sync(self.zeroconf.register_service, self._service_info)
+        # self._browser = ServiceBrowser(self.zeroconf, SERVICE_TYPE, handlers=[self._on_service_state_change])
+        print(f"Starting mDNS discovery for peer {self.peer_id} on port {self.port}")
+
+    async def stop(self):
+        # self._running = False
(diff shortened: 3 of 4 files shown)


seetadev:
@sumanjeet0012 : Thank you for resolving the merge conflicts. Did leave some notes at https://github.com/libp2p/py-libp2p/discussions/634#discussioncomment-13480675 . We will discuss soon.

sumanjeet0012:
@seetadev The pull request is ready for review. I would especially appreciate your feedback on the logic used to extract the local IP address [(mDNS.broadcaster.py)](libp2p/discovery/mdns/broadcaster.py) and construct the PeerInfo object [mdns.listener.py](libp2p/discovery/mdns/listener.py) In the meantime, I am working on designing appropriate test cases for this PR.

libp2p/discovery/mdns/listener.py
```diff
@@ -0,0 +1,90 @@
...
+    ServiceListener,
+    Zeroconf,
+)
+
+from libp2p.abc import IPeerStore, Multiaddr
+from libp2p.discovery.events.peerDiscovery import peerDiscovery
+from libp2p.peer.id import ID
+from libp2p.peer.peerinfo import PeerInfo
+
+logger = logging.getLogger("libp2p.discovery.mdns.listner")
```
guha-rahul: it should be `listener` not `listner`

libp2p/discovery/mdns/listener.py
```diff
@@ -0,0 +1,90 @@
...
+            return
+        peer_info = self._extract_peer_info(info)
+        if peer_info:
+            self.peerstore.add_addrs(peer_info.peer_id, peer_info.addrs, 10)
+            logger.debug("Discovered:", peer_info.peer_id)
+
+    def _extract_peer_info(self, info: ServiceInfo) -> PeerInfo | None:
+        try:
+            addrs = [
+                Multiaddr(f"/ip4/{socket.inet_ntoa(addr)}/udp/{info.port}")
```
guha-rahul: a smol doubt: Will this not be `tcp` which  since it is in the multiaddr the actual transport protocol that libp2p peers use to connect to each other?

sumanjeet0012: @guha-rahul I also believe this should be TCP. Could you please confirm whether the IP address used to create the multiaddress is correct?

libp2p/discovery/mdns/listener.py
sumanjeet0012: fixed it.

libp2p/discovery/mdns/listener.py
guha-rahul: its seems correct

guha-rahul: also regarding that, we also want ipv6 support right? so we have to handle that accordingly.

seetadev:
@sumanjeet0012 : Great work. Thank you for the initiative. Appreciate your wonderful efforts.

@guha-rahul : Thank you for sharing the feedback and issue. Appreciate it.

CCing @acul71 and @lla-dane , who will be using mdns discovery method in py-peer for the workshop and coming projects. Please review and share feedback. This is a very important PR indeed :)

libp2p/__init__.py
```diff
@@ -265,9 +269,12 @@ def new_host(
         listen_addrs=listen_addrs,
     )
 
+    if enable_mDNS:
+        mdns = MDNSDiscovery(swarm)
+        mdns.start()
+
```
lla-dane: @sumanjeet0012 : Here in the `new_host` function, mDNS service is started. If there is an mDNS object accessible in `IHost` or `INetworkInterface`, users could interact directly with the mDNS object in their application code.

And with this, if `enable_mDNS` is true while creating the `host = new_host()`, they could also stop it after a few blocks of code using something like: `host.mdns.stop()` or `host.swarm.mdns.stop()`. 
Is this kind of feature available to the users ?

libp2p/discovery/mdns/listener.py
```diff
@@ -0,0 +1,90 @@
...
+            return
+        info = zc.get_service_info(type_, name, timeout=5000)
+        if not info:
+            return
+        peer_info = self._extract_peer_info(info)
+        if peer_info:
+            self.discovered_services[name] = peer_info.peer_id
+            self.peerstore.add_addrs(peer_info.peer_id, peer_info.addrs, 10)
+            peerDiscovery.emit_peer_discovered(peer_info)
+            logger.debug("Discovered Peer:", peer_info.peer_id)
```
lla-dane: @sumanjeet0012 : Just for curiosity, how exactly does the `listener.add_service()` gets called when there is a new_node in the LAN. Like where is the trigger and how does it work?

lla-dane:
Hey @sumanjeet0012: great work on this PR, looks solid. Experimented a little bit with the mDNS demo. Added a few inline suggestions and questions.

libp2p/discovery/mdns/listener.py
```diff
@@ -0,0 +1,90 @@
...
+        info = zc.get_service_info(type_, name, timeout=5000)
+        if not info:
+            return
+        peer_info = self._extract_peer_info(info)
+        if peer_info:
+            self.peerstore.clear_addrs(peer_info.peer_id)
+            self.peerstore.add_addrs(peer_info.peer_id, peer_info.addrs, 10)
+            logger.debug("Updated Peer", peer_info.peer_id)
+
+    def _process_discovered_service(
```
kaneki003: Hey @sumanjeet0012 ,this functions is doing almost same work as `add_service` function .Is there any special use-case for this?

libp2p/discovery/mdns/broadcaster.py
```diff
@@ -0,0 +1,64 @@
...
+            # This doesn't actually send data
+            with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as s:
+                s.connect(("8.8.8.8", 80))
+                local_ip = s.getsockname()[0]
+            return local_ip
+        except Exception:
+            # Fallback to localhost if we can't determine the IP
+            return "127.0.0.1"
+
+    def register(self) -> None:
```
kaneki003, line 59: Just an enhancement,adding try-catch hook to log `EventLoopBlocked` exception here would be nice.

libp2p/discovery/mdns/broadcaster.py
```diff
@@ -0,0 +1,64 @@
...
+        """Register the peer's mDNS service on the network."""
+        self.zeroconf.register_service(self.service_info)
+
+    def unregister(self) -> None:
```
kaneki003, line 75: Here too.

libp2p/discovery/mdns/listener.py
sumanjeet0012: I forgot to remove it when I was refactoring the codes.

libp2p/discovery/mdns/broadcaster.py
sumanjeet0012: Added try catch block.

libp2p/discovery/mdns/broadcaster.py
sumanjeet0012: Added try catch block, is everything good now ?

kaneki003: Yes!
Great PR on mDNS  🎉🎉

libp2p/discovery/mdns/listener.py
sumanjeet0012: @lla-dane That's an interesting question.

What happens is, in the `listener.py` file, a `ServiceBrowser` instance is created with `self` passed as the listener. This means the class in which the `ServiceBrowser` is instantiated becomes the listener for the service.

The listener receives a `state_change` argument of type `ServiceStateChange`, which can have values like `Added`, `Removed`, or `Updated`.

Any listener class that extends the `ServiceListener` interface must implement the methods `add_service`, `update_service`, and `remove_service`. Based on the value of `state_change`, the corresponding method is automatically invoked by the underlying Zeroconf mechanism.

Alternatively, we can create a custom listener without extending the `ServiceListener` class. In that case, we need to manually inspect the `state_change` value and call the appropriate method ourselves.

libp2p/discovery/mdns/listener.py
sumanjeet0012: Yes I agree, I'll add it soon.

libp2p/__init__.py
sumanjeet0012: @lla-dane I’m also interested in implementing something similar. I was attempting to attach mDNS to the host lifecycle but couldn't figure out the correct approach to do so.

Additionally, I was considering how we might allow users to provide configuration options (such as the port number). To achieve this, we may need to initialize mDNS in a similar manner to how it's done in JS libp2p:

```ts
export function mdns (init: MulticastDNSInit = {}): (components: MulticastDNSComponents) => PeerDiscovery {
  return (components: MulticastDNSComponents) => new MulticastDNS(components, init)
}
```

What’s your opinion on this approach?
Should I continue exploring this direction, or would it be better to leave it for a future release of mDNS?

lla-dane:
```py
class BasicHost(IHost):
    _network: INetworkService
    peerstore: IPeerStore
    mdns: MDNSDiscovery # Added MDNSDiscovery object

    multiselect: Multiselect
    multiselect_client: MultiselectClient

    def __init__(
        self,
        network: INetworkService,
        default_protocols: Optional["OrderedDict[TProtocol, StreamHandlerFn]"] = None,
    ) -> None:
        self._network = network
        self._network.set_stream_handler(self._swarm_stream_handler)
        self.peerstore = self._network.peerstore
        self.mdns = MDNSDiscovery(self._network) #Initiliaze MDNS object
        # Protocol muxing
        default_protocols = default_protocols or get_default_protocols(self)
        self.multiselect = Multiselect(dict(default_protocols.items()))
        self.multiselect_client = MultiselectClient()
```

Would just adding MDNS Discovery object in this way won't work ?  @sumanjeet0012 
Ans then starting with, like: `host.mdns.start()`

seetadev:
@sumanjeet0012 : Wonderful work, Sumanjeet. Please share developer docs + screencast (or demo). We are almost ready with the PR merge. Appreciate your efforts.

CCing @lla-dane , @kaneki003 and @guha-rahul for their feedback points and thoughts on the new changes. Thank you everyone for your great contribution in this PR :) Very important py-libp2p module indeed.

sumanjeet0012:
@seetadev Ready for review.

seetadev:
@sumanjeet0012 : Thank you so much for thoughtfully addressing all the feedback points and making the necessary improvements. The latest changes look great — clear, well-structured, and aligned with the direction we discussed.

Really appreciate your responsiveness and the care you've taken in refining this contribution. Everything looks good on our end, and it's ready to be merged!

Looking forward to more great contributions from you.

Issue #641, opened 2025-05-31
feat: Add blacklisting of peers


guha-rahul:
### Description

Adding blacklisting methods for malicious peers in `libp2p/pubsub`
https://github.com/libp2p/py-libp2p/blob/67ab6e27d8bd54004c49e8828b69b4b0db264c5b/libp2p/pubsub/pubsub.py#L612-L614

### Motivation

There is no way of blacklisting malicios peers 
ref - implementation in go-libp2p (https://github.com/libp2p/go-libp2p-pubsub/blob/3f89e4331c981a6b60206b762a10d015c04875a8/blacklist.go#L12)

### Current Implementation

Currently its not implemented in pubsub

### Are you planning to do it yourself in a pull request ?

Yes

mystical-prog:
Hey @guha-rahul 

As I have worked really closely with Gossipsub and Pubsub for the last few months, I would love to discuss the implementation and ideas you have in mind for this, please refer to some of the advance implementations like go-libp2p or rs-libp2p.

guha-rahul:
@mystical-prog I have referred some of the implementations and I have implemented the feature, will put a pr by EOD

mystical-prog:
great work @guha-rahul 
I pointed out an area of improvement on the PR

mystical-prog:
we can mark this issue closed @seetadev

seetadev:
@mystical-prog : Sure, thank you for sharing, Vraj. Thank you for sharing feedback and review on the PR.

Nice effort by @guha-rahul indeed.

PR #651, opened 2025-06-02
feat: Add blacklisting of peers

guha-rahul:
## What was wrong?

Issue #641 

## How was it fixed?

- Added some functions for blacklisting in `libp2p/pubsub.py` and added tests for those in `tests/core/pubsub/test_pubsub.py`
- added a check in `_handle_new_peer` to reject blacklisted peers before starting stream 
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -329,6 +329,12 @@ class Pubsub(Service, IPubsub):
     def add_to_blacklist(self, peer_id: ID) -> None:
         """
         Add a peer to the blacklist.
+        When a peer is blacklisted:
+        - Any existing connection to that peer is immediately closed and removed
+        - The peer is removed from all topic subscription mappings
+        - Future connection attempts from this peer will be rejected
+        - Messages forwarded by or originating from this peer will be dropped
+        - The peer will not be able to participate in pubsub communication
 
         :param peer_id: the peer ID to blacklist
         """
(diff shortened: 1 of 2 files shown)


libp2p/pubsub/pubsub.py
```diff
@@ -346,6 +409,10 @@ async def wait_until_ready(self) -> None:
         await self.event_handle_dead_peer_queue_started.wait()
 
     async def _handle_new_peer(self, peer_id: ID) -> None:
+        if self.is_peer_blacklisted(peer_id):
```
guha-rahul, line 431: Should this be handled before initiating a connection or should his be placed after sending hello packet.

libp2p/pubsub/pubsub.py
```diff
@@ -122,6 +122,9 @@ class Pubsub(Service, IPubsub):
     strict_signing: bool
     sign_key: PrivateKey
 
+    # Set of blacklisted peer IDs
+    blacklisted_peers: set[ID]
```
mystical-prog, line 126: just a minor performance improvement suggestion, I guess @seetadev and @pacrob would be a better judge of it - but having a dict instead of set would reduce blacklist check time to O(1)

```suggestion
    blacklisted_peers: dict[ID, bool]
```

libp2p/pubsub/pubsub.py
mystical-prog: I guess handling this before sending hello packet is correct

libp2p/pubsub/pubsub.py
guha-rahul: Since both set and dict both use hash tables under the hood and we have only membership checks, it should be O(1) nevertheless , right? . These are my naive thoughts.

pacrob:
`set` is fine to use here, since we're just storing the single piece of information, the `ID`. I think only if we needed more info about the ID would a `dict` be warranted.

This looks great, @guha-rahul! Please add a newsfragment. Could you also expand on the documentation a bit? When looking at the functions in the docs, all we we see is "Add a peer to the blacklist", "Remove a peer from the blacklist.", etc. A bit more information about what happens when you add/remove an id from the blacklist (e.g. it removes them from `peers` and `subscriptions`, etc.)  would be helpful for new users with less context.

PR #655, opened 2025-06-04
Add ttl for peer data expiration

### LINKED ISSUE #650: Add ttl to identified peers
### Description

libp2p should be able to remove all the identified peers with expired TTL. Only the valid ones should remain in the Peerstore. 

### Motivation

The peerstore will not advertise its multiaddr and other properties, such as identify/push messages to expired peers, reducing network congestion.

### Current Implementation

Currently, we pass the TTL value when creating an identified peer entry in Peerstore, but we do not clean it up after the TTL expires. 

Yes

sukhman-sukh:
Currently, we are not invalidating peer data based on the TTL and Last Identified. 

Issue https://github.com/libp2p/py-libp2p/issues/650
Closes #650 

## How was it fixed?
Created 2 fields under the PeerData class
-  `last_identified`: which gets updated whenever any host adds and address to any peer
-  `ttl`: the specified time to live (in seconds).

In the current implementation, I have kept TTL as optional with a default value = 0, which means the data will never expire.
Additionally, I have modified some previous tests where the TTL passed was 10 to 600 seconds (10 minutes).

### To-Do

- Add changelog once PR gets approved.

#### Cute Animal Picture

### DIFF
--- a/libp2p/abc.py
+++ b/libp2p/abc.py
@@ -1450,11 +1450,12 @@ class IPeerData(ABC):
     def get_last_identified(self) -> int:
         """
         Fetch the last identified timestamp
-        
+
         Returns
         -------
         last_identified_timestamp
             The lastIdentified time of peer.
+
         """
 
     @abstractmethod
@@ -1485,11 +1486,12 @@ class IPeerData(ABC):
     def is_expired(self) -> bool:
         """
         Check if the peer is expired based on last_identified and ttl
-        
+
         Returns
         -------
         bool
             True, if last_identified + ttl > current_time
+
         """
 
 
--- a/libp2p/abc.py
+++ b/libp2p/abc.py
@@ -1439,7 +1439,53 @@ class IPeerData(ABC):
             If the private key is not found.
 
         """
+        
+    @abstractmethod
+    def update_last_identified(self) -> None:
+        """
+        Updates timestamp to current time.
+        """
 
+    @abstractmethod
+    def get_last_identified(self) -> int:
+        """
+        Fetch the last identified timestamp
+        Returns
+        -------
+        last_identified_timestamp
+            The lastIdentified time of peer.
+        """
+    
+    def get_ttl(self) -> int:
+        """
+        Get ttl value for the peer for validity check
+
+        Parameters
+        ----------
+        ttl : int
+            The ttl for the peer.
+
+        """
+    
+    def set_ttl(self, ttl: int) -> None:
+        """
+        Set ttl value for the peer for validity check
+
+        Returns
+        -------
+        TTL
+            The ttl of the peer.
+
+        """
+    
+    def is_expired(self) -> bool:
+        """
+        Check if the peer is expired based on last_identified and ttl
+        Returns
+        -------
+        bool
+            True, if last_identified + ttl > current_time 
+        """
 
 # ------------------ multiselect_communicator interface.py ------------------
 
--- a/libp2p/abc.py
+++ b/libp2p/abc.py
@@ -1439,7 +1439,7 @@ class IPeerData(ABC):
             If the private key is not found.
 
         """
-        
+
     @abstractmethod
     def update_last_identified(self) -> None:
         """
@@ -1455,7 +1455,8 @@ class IPeerData(ABC):
         last_identified_timestamp
             The lastIdentified time of peer.
         """
-    
+
+    @abstractmethod
     def get_ttl(self) -> int:
         """
         Get ttl value for the peer for validity check
@@ -1466,7 +1467,8 @@ class IPeerData(ABC):
             The ttl for the peer.
 
         """
-    
+
+    @abstractmethod
     def set_ttl(self, ttl: int) -> None:
         """
         Set ttl value for the peer for validity check
@@ -1477,15 +1479,17 @@ class IPeerData(ABC):
             The ttl of the peer.
 
         """
-    
+
+    @abstractmethod
     def is_expired(self) -> bool:
         """
         Check if the peer is expired based on last_identified and ttl
         Returns
         -------
         bool
-            True, if last_identified + ttl > current_time 
+            True, if last_identified + ttl > current_time
         """
+
 
 # ------------------ multiselect_communicator interface.py ------------------
 
--- a/libp2p/abc.py
+++ b/libp2p/abc.py
@@ -1450,6 +1450,7 @@ class IPeerData(ABC):
     def get_last_identified(self) -> int:
         """
         Fetch the last identified timestamp
+        
         Returns
         -------
         last_identified_timestamp
@@ -1484,6 +1485,7 @@ class IPeerData(ABC):
     def is_expired(self) -> bool:
         """
         Check if the peer is expired based on last_identified and ttl
+        
         Returns
         -------
         bool
--- a/libp2p/abc.py
+++ b/libp2p/abc.py
@@ -1463,10 +1463,10 @@ class IPeerData(ABC):
         """
         Get ttl value for the peer for validity check
 
-        Parameters
-        ----------
-        ttl : int
-            The ttl for the peer.
+        Returns
+        -------
+        int
+            The ttl of the peer.
 
         """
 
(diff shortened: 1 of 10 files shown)


sukhman-sukh:
Hey @seetadev @mystical-prog, can you please reveiw this PR.

libp2p/host/basic_host.py
```diff
@@ -229,7 +229,7 @@ async def connect(self, peer_info: PeerInfo) -> None:
         :param peer_info: peer_info of the peer we want to connect to
         :type peer_info: peer.peerinfo.PeerInfo
         """
-        self.peerstore.add_addrs(peer_info.peer_id, peer_info.addrs, 10)
+        self.peerstore.add_addrs(peer_info.peer_id, peer_info.addrs, 600)
```
mystical-prog: with respect to go implementation, in basic host we keep the ttl for addrs as 120 (2 minutes), we should make sure that we are in sync with it so that interop issues can be avoided

libp2p/host/routed_host.py
mystical-prog, line 1: as mentioned in basic host, the value for ttl is 120 in go implementation

libp2p/peer/peerdata.py
```diff
@@ -24,13 +25,17 @@ class PeerData(IPeerData):
     metadata: dict[Any, Any]
     protocols: list[str]
     addrs: list[Multiaddr]
+    last_identified: int
+    ttl: int  # Keep ttl=0 by default for always valid
```
mystical-prog, line 29: since we are considering a peer permanent if ttl is 0 then we should update the const PERMANENT_ADDR_TTL in `peerstore.py` as well

libp2p/peer/peerstore.py
```diff
@@ -108,32 +124,38 @@ def put(self, peer_id: ID, key: str, val: Any) -> None:
         peer_data = self.peer_data_map[peer_id]
         peer_data.put_metadata(key, val)
 
-    def add_addr(self, peer_id: ID, addr: Multiaddr, ttl: int) -> None:
+    # pass without ttl and pass with 2 sec ttl and time.sleep(2)
```
mystical-prog: I guess this comment is unintentional ?

libp2p/peer/peerstore.py
```diff
@@ -49,8 +49,10 @@ def peer_info(self, peer_id: ID) -> PeerInfo:
         """
         if peer_id in self.peer_data_map:
             peer_data = self.peer_data_map[peer_id]
+            if peer_data.is_expired():
+                peer_data.clear_addrs()
             return PeerInfo(peer_id, peer_data.get_addrs())
-        raise PeerStoreError("peer ID not found")
+        raise PeerStoreError("peer ID not found or expired")
```
mystical-prog: we should have a separate exception type for peer not found and peer expired ?
we can use this differentiation for advance implementations

libp2p/peer/peerstore.py
sukhman-sukh: Oh yes, even I don't remember now why it was there. Removing it.

libp2p/peer/peerdata.py
sukhman-sukh: Done

libp2p/host/routed_host.py
sukhman-sukh: Done

libp2p/host/basic_host.py
sukhman-sukh: Done

libp2p/peer/peerstore.py
```diff
@@ -84,6 +86,20 @@ def peer_ids(self) -> list[ID]:
...
+        :return: all of the valid peer IDs stored in peer store
+        """
+        valid_peer_ids: list[ID] = []
+        peer_ids = list(self.peer_data_map.keys())
+        for peer_id in peer_ids:
+            peer_data = self.peer_data_map[peer_id]
+            if not peer_data.is_expired():
+                valid_peer_ids.append(peer_id)
+            else:
+                peer_data.clear_addrs()
```
mystical-prog:
```suggestion
        for peer_id, peer_data in self.peer_data_map.items():
            if not peer_data.is_expired():
                valid_peer_ids.append(peer_id)
            else:
                peer_data.clear_addrs()
```

I think we can reduce this block of code for better readability

libp2p/peer/peerstore.py
sukhman-sukh: Here, I am removing the expired part.  For expired, we are removing the addresses but keeping the other information. So, the implementation is to clear all the addresses on expiry and return the rest of the info on peer_info() call, so no need for raising expired error.

mystical-prog: in that case, the error message can be simply just `peer ID not found` cause the first thought that came to me when I read `or expired` was I can somehow recover the peer addrs (they seem to exist but with a tag of expired)

tests/core/peer/test_addrbook.py
mystical-prog, line 1: are the tests failing when you don't change the ttl to 600 ? if no then there is no need to change them

tests/core/peer/test_peerstore.py
```diff
@@ -16,9 +18,31 @@ def test_peer_info_empty():
 
 def test_peer_info_basic():
```
mystical-prog, line 21: if we can reduce the ttl for `foo` and `foo2` to 1 and 2, this would result in a faster testing, we are experiencing long CI duration due to unnecessarily testing with a sleep of 2-3 seconds, as we keep on adding more features and tests - we need to be vigilant of making the tests as fast as possible

tests/core/protocol_muxer/test_protocol_muxer.py
mystical-prog, line 1: are the tests failing when you don't change the ttl to 600 ? 
if no then there is no need to change them
if yes then set ttl as small as possible - just for sanity

tests/core/test_libp2p/test_libp2p.py
mystical-prog, line 1: again,

are the tests failing when you don't change the ttl to 600 ?
if no then there is no need to change them
if yes then set ttl as small as possible - just for sanity

sukhman-sukh:
@mystical-prog, all above suggested changes done.

mystical-prog:
kudos @sukhman-sukh 
great work, the PR is ready for review

cc : @seetadev @pacrob

libp2p/abc.py
```diff
@@ -1434,6 +1434,60 @@ def get_privkey(self) -> PrivateKey:
...
+
+    @abstractmethod
+    def set_ttl(self, ttl: int) -> None:
+        """
+        Set ttl value for the peer for validity check
+
+        Returns
+        -------
+        TTL
+            The ttl of the peer.
```
pacrob: Mismatch between actual return type and docstring

libp2p/abc.py
```diff
@@ -1434,6 +1434,60 @@ def get_privkey(self) -> PrivateKey:
...
+
+    @abstractmethod
+    def get_ttl(self) -> int:
+        """
+        Get ttl value for the peer for validity check
+
+        Parameters
+        ----------
+        ttl : int
+            The ttl for the peer.
```
pacrob: docstring mismatch

libp2p/peer/peerdata.py
```diff
@@ -115,6 +120,40 @@ def get_privkey(self) -> PrivateKey:
             raise PeerDataError("private key not found")
         return self.privkey
 
+    def update_last_identified(self) -> None:
+        """
+        :param timestamp: timestamp to set
```
pacrob: docstring mismatch

pacrob:
The code looks fine, please just clean up the docstrings so that they properly describe params and returns, and that they match, more or less, across `abc.py` and `peerdata.py`.

sukhman-sukh:
@pacrob, I have fixed the docstrings.

libp2p/peer/peerdata.py
```diff
@@ -115,6 +120,37 @@ def get_privkey(self) -> PrivateKey:
...
+        """
+        self.ttl = ttl
+
+    def is_expired(self) -> bool:
+        """
+        :return: true, if last_identified+ttl > current_time
+        """
+        # for ttl = 0; peer_data is always valid
+        if self.ttl > 0 and self.last_identified + self.ttl < int(time.time()):
+            print("reached true")
```
pacrob: Remove `print` statement. Use logging if you think it would be helpful for debugging.

PR #661, opened 2025-06-08
Updated examples to automatically use random port

### LINKED ISSUE #659: Update examples to automatically use random port
### Description

Currently in the examples, a default port of `8000` is usually used for servers, and in most examples, running client requires to manually provide a different port. For example in echo server:

Image

`py-libp2p` inherently do not require to specify a available port, we can simply run client or server using port `0` and a random open port is automatically assigned.

### Motivation

The aim is to show in examples that one can simply connect to a address without worrying if the port is available. 
Eg: `echo-demo -d /ip4/127.0.0.1/tcp/8000/p2p/16Uiu2HAmAsbxRR1HiGJRNVPQLNMeNsBCsXT3rDjoYBQzgzNpM5mJ`

### Current Implementation

Currently one needs to provide an available port for libp2p to start.

No, this could be picked up by someone looking for good first issue.

lla-dane:
Aims to resolve #659 

Updated the `ping, echo, chat` example to use random port when `-p` flag is not specified. 

`Ping-demo with random port`:

image


examples/chat/chat.py
```diff
@@ -87,25 +89,38 @@ def main() -> None:
...
+        except OSError:
+            return False
+
+
+def get_random_available_port(start: int = 10000, end: int = 20000) -> int:
+    for _ in range(50):  # try up to 50 random ports
+        port = random.randint(start, end)
+        if is_port_available(port):
+            return port
+    raise RuntimeError("Could not find an available port in the given range.")
```
AkMo3: @lla-dane  An easy to solve this to simply use default port as 0 while opening a listener/server, and then the application will automatically bind to a available port.

AkMo3 (changes requested):

examples/chat/chat.py
lla-dane: Yeah I tried that at first, but it wasn't working in py-libp2p. It was happening like this, when I passed the port as 0:
image

lla-dane: It was going to port 0, and not to some random port. And then ping from some random port was failing.

AkMo3: The address displayed here is incorrect, as it is generated from the following string:
`f"-d /ip4/{localhost_ip}/tcp/{port}/p2p/{host.get_id().pretty()}\n"`

If you replace this with something like
`f"-d {host.get_addrs()[0]}\n"`
it would print the actual address and the port it is bind to

lla-dane: Yeah, missed it, my bad. Will do this for all the examples.

lla-dane:
@seetadev @AkMo3 : Updated the examples as requested, please take a look.

examples/identify_push/identify_push_listener_dialer.py
```diff
@@ -241,42 +238,32 @@ def main() -> None:
...
-    Without arguments, it runs as a listener on port 8888.
-    With -d parameter, it runs as a dialer on port 8889.
+    Without arguments, it runs as a listener on random port.
+    With -d parameter, it runs as a dialer on random port.
     """
 
     example = (
-        f"/ip4/127.0.0.1/tcp/{DEFAULT_PORT}/p2p/"
-        "QmQn4SwGkDZkUEpBRBvTmheQycxAHJUNmVEnjA2v1qe8Q"
+        "/ip4/127.0.0.1/tcp/8000/p2p/" "QmQn4SwGkDZKkUEpBRBvTmheQycxAHJUNmVEnjA2v1qe8Q"
```
AkMo3: This address looks to be invalid, for a safer side, lets make it consistent with address used in other examples.

examples/chat/chat.py
```diff
@@ -87,19 +86,16 @@ def main() -> None:
...
         "--destination",
         type=str,
         help=f"destination multiaddr string, e.g. {example_maddr}",
     )
     args = parser.parse_args()
-
-    if not args.port:
-        raise RuntimeError("was not able to determine a local port")
+    if args.port is None:
+        raise RuntimeError("failed to determine local port")
```
AkMo3: Since port has a default value, the `args.port` cannot be None, we can skip this check. This similar check is also present in multiple other places, we can remove them as well.

AkMo3 (commented):
LGTM, a few minor suggestions:

examples/identify_push/identify_push_listener_dialer.py
lla-dane: Done!

examples/chat/chat.py
lla-dane: Done!

newsfragments/661.feature.rst
AkMo3, line 1: One final request, please put this change under `docs`, I don't think this qualifies as a feature.

lla-dane: Done!

AkMo3 (approved):

AkMo3:
@seetadev All good from my side, we can merge this 🚀

pacrob (approved):

Issue #652, opened 2025-06-02
refactor: fanout_heartbeat and gossip_heartbeat


guha-rahul:
### Description

The two functions contain similar logic thatt can be extracted into a single function. 
https://github.com/libp2p/py-libp2p/blob/67ab6e27d8bd54004c49e8828b69b4b0db264c5b/libp2p/pubsub/gossipsub.py#L517
and 
https://github.com/libp2p/py-libp2p/blob/67ab6e27d8bd54004c49e8828b69b4b0db264c5b/libp2p/pubsub/gossipsub.py#L566

### Motivation

This removes the duplicate code for collection and processing of message ids from message cache in mesh and fanout.

### Current Implementation

The logic is written in two separate functions.

### Are you planning to do it yourself in a pull request ?

Yes

PR #678, opened 2025-06-13
refactor(gossipsub.py): Add helper function to fanout and gossipsub  

guha-rahul:
## What was wrong?

Issue #652 

## How was it fixed?
Added a helper function from the common logic of the two functions 

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/libp2p/pubsub/gossipsub.py
+++ b/libp2p/pubsub/gossipsub.py
@@ -550,12 +550,6 @@ class GossipSub(IPubsubRouter, Service):
 
         # Skip if no peers have subscribed to the topic
         if topic not in self.pubsub.peer_topics:
-            if is_fanout:
-                # For fanout topics, check if we should remove the topic
-                if self.time_since_last_publish.get(topic, 0) + self.time_to_live < int(
-                    time.time()
-                ):
-                    return set(), True
             return current_peers, False
 
         # For fanout topics, check if we should remove the topic


seetadev:
@guha-rahul : Great, thank you for sharing the PR. Please add a test file supporting the PR and newsfragment.

Reviewing the PR. Will share feedback soon.

CCing @mystical-prog and @pacrob for their feedback too.

guha-rahul:
@seetadev This pr actually merges a common logic into a single function , so no new extra logic is being introduced and all the existing tests are running

mystical-prog:
thank you @seetadev for mentioning me on this PR, I would like to point out that this change is a very nice catch in terms of maintainability of code - nice eye for clean up for sure @guha-rahul :clap: 

as far as performance is concerned - it remains as it as and no tests should fail for this change, was considering to introduce this myself with gossipsub upgrade :upside_down_face: but kudos to Rahul :partying_face: !

seetadev:
> thank you @seetadev for mentioning me on this PR, I would like to point out that this change is a very nice catch in terms of maintainability of code - nice eye for clean up for sure @guha-rahul 👏
> 
> as far as performance is concerned - it remains as it as and no tests should fail for this change, was considering to introduce this myself with gossipsub upgrade 🙃 but kudos to Rahul 🥳 !


@mystical-prog : Hi Vraj.

Thanks so much for taking a look and sharing your thoughts! 🙏 Appreciate it.

Really glad you agree on the value of this cleanup — I felt the same and thought it was worth bringing into focus early. Appreciate you highlighting the maintainability aspect and confirming the performance implications as well — that gives us confidence moving forward.

And yes, full credit to @guha-rahul for the sharp catch here 👏 — great to see things like this landing even ahead of the Gossipsub upgrade! Looking forward to building on this momentum with all our team members at Py-libp2p 💪

libp2p/pubsub/gossipsub.py
```diff
@@ -528,82 +528,105 @@ def mesh_heartbeat(
...
+                ):
+                    return set(), True
+            return current_peers, False
+
+        # For fanout topics, check if we should remove the topic
+        if is_fanout:
+            if self.time_since_last_publish.get(topic, 0) + self.time_to_live < int(
+                time.time()
+            ):
+                return set(), True
```
pacrob: It looks like the fanout check is duplicated here. Would this not achieve the same result?

```suggestion
        # For fanout topics, check if we should remove the topic
        if is_fanout:
            if self.time_since_last_publish.get(topic, 0) + self.time_to_live < int(
                time.time()
            ):
                return set(), True
                
        # Skip if no peers have subscribed to the topic
        if topic not in self.pubsub.peer_topics:
            return current_peers, False
```

pacrob:
Looks great, @guha-rahul! One question on some repeated code, good to merge once that's resolved. Thanks!

libp2p/pubsub/gossipsub.py
guha-rahul: Ah yes it seems like the fanout check is duplicated. Also quick question, wouldnt it be a better choice if we first check `if topic not in self.pubsub.peer_topics` ?

pacrob: > wouldnt it be a better choice if we first check `if topic not in self.pubsub.peer_topics` ?

Yes, that makes sense.

guha-rahul: made the changes, thank you for the review!

PR #680, opened 2025-06-14
feat: add support for sparse connect 

### LINKED ISSUE #679: feat: Add support for Sparse Connect
### Description

Add the implementation for sparse connect and its related test cases.

### Motivation

Dense connect becomes inefficient in bigger networks , thus in bigger networks sparse connect scales better consuming less resources .

### Current Implementation

The current implementation only accepts dense connect.

Yes

guha-rahul:
## What was wrong?

Issue #679 
Closes #679 

## How was it fixed?

Added the sparse connect implementation in utils.py
added tests for it in test_gossipsub.py

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

Cute amimal

### DIFF
--- a/newsfragments/679.feature.rst
+++ b/newsfragments/679.feature.rst
@@ -1,0 +1,1 @@
+Added sparse connect to pubsub, automatically switching between dense and sparse topologies based on network size.
--- a/newsfragments/679.feature.rst
+++ b/newsfragments/679.feature.rst
@@ -1,1 +1,1 @@
-Added sparse connect to pubsub, automatically switching between dense and sparse topologies based on network size.
+Added sparse connect utility function to pubsub test utilities for creating test networks with configurable connectivity.
--- a/tests/core/pubsub/test_gossipsub.py
+++ b/tests/core/pubsub/test_gossipsub.py
@@ -510,9 +510,8 @@ async def test_gossip_heartbeat(initial_peer_count, monkeypatch):
 
 
 @pytest.mark.trio
-async def test_sparse_connect():
-    """Test sparse connect functionality including transition from dense to sparse."""
-    # Test with small network (should use dense connect)
+async def test_dense_connect_fallback():
+    """Test that sparse_connect falls back to dense connect for small networks."""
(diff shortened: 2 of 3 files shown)


seetadev:
@guha-rahul : Nice work on the PR. Reviewing it.

CCing @pacrob and @mystical-prog (Vraj) for their feedback too.

newsfragments/679.feature.rst
mystical-prog, line 1: I think we need to change the newsfragment text, as we are not adding new sparse_connect method to pubsub implementation but are adding it to pubsub test util, also there is no automatic switching when we will run pubsub implementation after these changes - the automatic switching remains inside a function which is in tests/utils and not the core implementation - so we should drop that

tests/core/pubsub/test_gossipsub.py
```diff
@@ -506,3 +507,68 @@ def window(topic):
...
+        # Verify message was received
+        received_messages = []
+        for queue in queues:
+            try:
+                msg = await queue.get()
+                received_messages.append(msg.data)
+            except Exception:
+                continue
+
+        assert len(received_messages) > 0, "No messages were received"
```
mystical-prog: in order to efficiently test the msg propagation, we should ensure that all of the nodes received the msg - thus ensuring scalability in context of sparse connections, if not all then at least more than half

tests/core/pubsub/test_gossipsub.py
pacrob, line 1: It would be better to break `test_sparse_connect` up into 2 tests, one for the dense connect path and one for the sparse connect path.

pacrob:
Thanks, @guha-rahul! I agree with comments from @mystical-prog, and left my own.

Another more general question: now that we have this in addition to `dense_connect`, are there places in our testing that would benefit from using `sparse_connect` instead of, or in addition to, `dense_connect`? Fine if not, just wondering :thinking:

guha-rahul:
I have added the changes like you suggested @pacrob @mystical-prog .
`test_fanout` and `test_fanout_maintenance` are some places which might use sparse connect instead of dense connect

pacrob (approved):

PR #681, opened 2025-06-15
fix: removed dummy ID(b) from upgrade_security for inbound connections

lla-dane:
### Resolved: FIXME: This dummy `ID(b"")` for the remote peer is useless.
Updated peer_id from `ID` to `Optional[ID] = None` in `transport/upgrader.py:upgrade_security()`.

@seetadev

### DIFF
--- a/libp2p/network/swarm.py
+++ b/libp2p/network/swarm.py
@@ -257,7 +257,6 @@ class Swarm(Service, INetworkService):
                 # Per, https://discuss.libp2p.io/t/multistream-security/130, we first
                 # secure the conn and then mux the conn
                 try:
-                    # FIXME: This dummy `ID(b"")` for the remote peer is useless.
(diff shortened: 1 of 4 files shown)


seetadev:
@lla-dane : Nice work on the PR. Looks ready for merge. Reviewing it once again.

CCing @pacrob and @sumanjeet0012 for their review too.

pacrob:
Hey, @lla-dane, this does fix the problem as described, but the more important thing is that the order of arguments to a public function has changed. This would be considered a breaking change, so the newsfragment should be of type `breaking` and the note should focus on reordering args for the function and providing a default for `peer_id`.

Because we're still in beta, a breaking change will not affect the release schedule, but good to keep track of such things. 

Thanks for your help! Once the newsfragment is updated, we can merge.

lla-dane:
Updated the newsfragment file to breaking. Please take a look @pacrob.

PR #696, opened 2025-06-20
fix: added negotiate timeout to MuxerMultistream

lla-dane:
Aims to resolve 
`FIXME: add negotiate timeout to MuxerMultistream`

There were 3 function which were reading from a partially established p2p connection: 
* `handshake`: Local peer writes `MULTISELECT_PROTOCOL_ID` to remote peer, and then waits for response in:
```py
response = await communicator.read()
```
Added a negotiate timeout there while reading like this:
```py
try:
            with trio.fail_after(DEFAULT_NEGOTIATE_TIMEOUT):  # Timeout after 5 seconds
                response = await communicator.read()
        except trio.TooSlowError:
            raise MultiselectClientError("protocol selection response timed out")
        except MultiselectCommunicatorError as error:
            raise MultiselectClientError() from error
```

* Did similar thing to `query_multistream_command`: Example 'ls' command to know the supported protocols.
* Similarly in `try_select`: Negotiate protocol after multistream_select handshake.

### DIFF
--- a/tests/core/protocol_muxer/test_negotiate_timeout.py
+++ b/tests/core/protocol_muxer/test_negotiate_timeout.py
@@ -1,0 +1,59 @@
+import pytest
+import trio
+
+from libp2p.abc import (
+    IMultiselectCommunicator,
+)
+from libp2p.custom_types import TProtocol
+from libp2p.protocol_muxer.exceptions import (
+    MultiselectClientError,
+    MultiselectError,
+)
+from libp2p.protocol_muxer.multiselect import Multiselect
+from libp2p.protocol_muxer.multiselect_client import MultiselectClient
+
+
+class DummyMultiselectCommunicator(IMultiselectCommunicator):
+    """
+    Dummy MultiSelectCommunicator to test out negotiate timmeout.
+    """
+
+    def __init__(self) -> None:
+        return
+
+    async def write(self, msg_str: str) -> None:
+        """Goes into infinite loop when .write is called"""
+        await trio.sleep_forever()
+
+    async def read(self) -> str:
+        """Returns a dummy read"""
+        return "dummy_read"
+
+
+@pytest.mark.trio
+async def test_select_one_of_timeout():
+    ECHO = TProtocol("/echo/1.0.0")
+    communicator = DummyMultiselectCommunicator()
+
+    client = MultiselectClient()
+
+    with pytest.raises(MultiselectClientError, match="response timed out"):
+        await client.select_one_of([ECHO], communicator, 2)
+
+
+@pytest.mark.trio
+async def test_query_multistream_command_timeout():
+    communicator = DummyMultiselectCommunicator()
+    client = MultiselectClient()
+
+    with pytest.raises(MultiselectClientError, match="response timed out"):
+        await client.query_multistream_command(communicator, "ls", 2)
+
+
+@pytest.mark.trio
+async def test_negotiate_timeout():
+    communicator = DummyMultiselectCommunicator()
+    server = Multiselect()
+
+    with pytest.raises(MultiselectError, match="handshake read timeout"):
+        await server.negotiate(communicator, 2)


lla-dane:
Please take a look @pacrob @seetadev, and see if this fixes it.

seetadev:
@lla-dane : Great work. Appreciate the initiative. 

Wish if you could summarize your approach for developing the feature enhancement: "added negotiate timeout to MuxerMultistream". Also, kindly help us understand the functionality that this PR brings in MuxerMultistream project (from a software architecture perspective) by sharing a note on the discussion page.

Appreciate your efforts and wonderful contribution.

seetadev:
@lla-dane : Please also add a test file and a newsfragment.

lla-dane:
So I am guessing this `FIXME` meant: adding a negotiation timeout whenever there's a handshake — i.e., when our node is on the receiving end and waiting for the remote peer to respond.

For example, during the `handshake` function shown here:

```py
 async def handshake(self, communicator: IMultiselectCommunicator) -> None:
        """
        Ensure that the client and multiselect are both using the same
        multiselect protocol.

        :param communicator: communicator to use to communicate with counterparty
        :raise MultiselectClientError: raised when handshake failed
        """
        try:
            await communicator.write(MULTISELECT_PROTOCOL_ID)
        except MultiselectCommunicatorError as error:
            raise MultiselectClientError() from error

        try:
            handshake_contents = await communicator.read()
        except MultiselectCommunicatorError as error:
            raise MultiselectClientError() from error

        if not is_valid_handshake(handshake_contents):
            raise MultiselectClientError("multiselect protocol ID mismatch")
```

In the scenario where our node has sent the multiselect protocol ID via the `.write(...)` call, and is now awaiting a response from the peer via `.read()`, the coroutine can hang indefinitely if the remote peer:
* never responds,
* or stalls without closing the stream.

Currently, there's no timeout in place to guard against this. So the `await communicator.read()` can suspend forever, which would cause the handshake — and therefore the entire connection negotiation — to stall.

To prevent this, I’ve wrapped the `read()` call in `trio.fail_after(DEFAULT_NEGOTIATE_TIMEOUT)`, so that if the peer doesn’t respond within a reasonable time window, the handshake fails fast with a MultiselectClientError.

This ensures that misbehaving or dead peers don't block the event loop or delay the dialer pipeline, especially in cases like peer discovery or during high churn.

@seetadev

kaneki003:
> Currently, there's no timeout in place to guard against this. So the `await communicator.read()` can suspend forever, which would cause the handshake — and therefore the entire connection negotiation — to stall.
> 
> To prevent this, I’ve wrapped the `read()` call in `trio.fail_after(DEFAULT_NEGOTIATE_TIMEOUT)`, so that if the peer doesn’t respond within a reasonable time window, the handshake fails fast with a MultiselectClientError.

Hey @lla-dane ,your approach is correct and the implementation is also looking fine.
Wanted to highlight that there is also a `handshake()` function in `multiselect.py`,which is doing the same work. I think the above changes should be done there as well.

lla-dane:
> Hey @lla-dane ,your approach is correct and the implementation is also looking fine.
Wanted to highlight that there is also a handshake() function in multiselect.py,which is doing the same work. I think the above changes should be done there as well.

Yeah thanks for the heads up @kaneki003, fixed there too.

lla-dane:
Also in the [negotiate](https://github.com/libp2p/py-libp2p/blob/4e2be87c73a82fd4bae3ea8524f9e12729d951ea/libp2p/protocol_muxer/multiselect.py#L59C1-L66C1) function here, there is one more read operation happening, but this is after the handshake, so do you think we should put similar kind of timeout here also @kaneki003 .

sumanjeet0012:
@lla-dane The read method `command = await communicator.read()` in libp2p/protocol_muxer/multiselect.py at line 63 currently lacks a timeout.

libp2p/protocol_muxer/multiselect.py
```diff
@@ -102,7 +105,10 @@ async def handshake(self, communicator: IMultiselectCommunicator) -> None:
             raise MultiselectError() from error
 
         try:
-            handshake_contents = await communicator.read()
+            with trio.fail_after(DEFAULT_NEGOTIATE_TIMEOUT):  # Timeout after 5 seconds
```
sumanjeet0012: The DEFAULT_NEGOTIATE_TIMEOUT is 60 seconds but the comment says timeout after 5 seconds.

libp2p/protocol_muxer/multiselect_client.py
```diff
@@ -93,8 +99,11 @@ async def query_multistream_command(
             raise ValueError("Command not supported")
 
         try:
-            response = await communicator.read()
+            with trio.fail_after(DEFAULT_NEGOTIATE_TIMEOUT):  # Timeout after 5 seconds
```
sumanjeet0012: Same here, DEFAULT_NEGOTIATE_TIMEOUT and comment are not matching.

libp2p/protocol_muxer/multiselect.py
lla-dane: Done! Changed `DEFAULT_NEGOTIATE_TIMEOUT` to 5 seconds.

libp2p/protocol_muxer/multiselect_client.py
lla-dane: Done!

lla-dane:
> @lla-dane The read method command = await communicator.read() in libp2p/protocol_muxer/multiselect.py at line 63 currently lacks a timeout.

Set the timeout condition there also.

sumanjeet0012:
@lla-dane I was thinking the main selection of protocol and creating new connections happen in stream_muxer/Muxer_multistream.py file, 
so instead of using timeout on every read and write operation of multiselect and multiselect_client, we can use timeout where these methods are used, in the select_transport and new_conn method in stream_muxer/muxer_multistream.py

```
async def select_transport(self, conn: IRawConnection) -> TMuxerClass:
 
    protocol: TProtocol
    communicator = MultiselectCommunicator(conn)
    async with trio.fail_after(DEFAULT_NEGOTIATE_TIMEOUT):
        if conn.is_initiator:
            protocol = await self.multiselect_client.select_one_of(
                tuple(self.transports.keys()), communicator
            )
        else:
            protocol, _ = await self.multiselect.negotiate(communicator)
    return self.transports[protocol]
```

What you think of this approach @seetadev, @acul71

lla-dane:
Yeah the negotiate timeout issue is relevant with `negotiate` and `select_one_of` functions. And both of these functions are used together at two places: 
1. `libp2p/security/security_multistream.py: select_transport()`
2. `libp2p/stream_muxer/muxer_multistream.py: select_transport()`

So either
1. we put timeout wrapper in `select_transport` functions at these two places or
2. in the `negotiate` and `select_one_of` functions which leads to `read/write` 

Doing this would be cleaner. 
So which should we go  with @sumanjeet0012 @seetadev

seetadev:
@lla-dane: Hi Abhinav. Thanks for laying this out so clearly — really appreciate the structured breakdown! Also, thanks to @sumanjeet0012 for his feedback and thoughts on this PR. Appreciate the review.

Yes, completely agree — the timeout concern is directly tied to the `negotiate` and `select_one_of` functions, since these are where the actual read/write happens during the multistream-select handshake. And you're also right that both `libp2p/security/security_multistream.py:select_transport()` and `libp2p/stream_muxer/muxer_multistream.py:select_transport()` invoke that logic.

Between the two approaches you outlined:

1. **Adding timeout wrappers directly in the two `select_transport()` functions**
2. **Adding timeout handling inside `negotiate` and `select_one_of` themselves**

I’d lean towards **option 2** — implementing the timeout logic within `negotiate` and `select_one_of`. It keeps the behavior consistent no matter where these functions are used, and avoids duplication. It also gives us a cleaner interface: callers of `select_transport()` won’t need to worry about wrapping the logic themselves — the timeout behavior is guaranteed and encapsulated.

That said, we should ensure this doesn’t conflict with any ongoing changes in `protocol_muxer/multiselect.py` that @lla-dane is working on. If their update already introduces timeout handling at the multistream protocol level, we can just build on that or integrate directly.

Wish to hear your thoughts and @sumanjeet0012 too - happy to sync on a direction that keeps things clean, safe at this juncture.

lla-dane:
> I’d lean towards option 2 — implementing the timeout logic within negotiate and select_one_of. It keeps the behavior consistent no matter where these functions are used, and avoids duplication. It also gives us a cleaner interface: callers of select_transport() won’t need to worry about wrapping the logic themselves — the timeout behavior is guaranteed and encapsulated.

@seetadev : Updated the timeout wrappers.
1. multiselect.py: `negotiate` function
2. multiselect_client.py: `select_one_of` , `query_multistream_command` functions

All these functions were leading to read/write operations. Please take a look @sumanjeet0012, if I missed something.

sumanjeet0012:
@lla-dane All changes are nicely done and I don't find anything missing,

sumanjeet0012:
@seetadev Thanks for the detailed explanation of why 2nd approach is much cleaner.
I also agree with your thoughts.
Now I don't find anything missing in this PR now,  and I believe we can move forward with merging it.

acul71:
@lla-dane 
Good work.
An option to make Timeout ( DEFAULT_NEGOTIATE_TIMEOUT ) configurable with a parameter would be appreciated (in view of a Connection Panel Control)

lla-dane:
> @lla-dane
Good work.
An option to make Timeout ( DEFAULT_NEGOTIATE_TIMEOUT ) configurable with a parameter would be appreciated (in view of a Connection Panel Control)

Made the DEFAULT_NEGOTIATE_TIMEOUT configurable. @acul71

pacrob:
Thanks, @lla-dane, please just add a test file showing that timeout works as expected. After that, I think it's good to merge.

lla-dane:
Hey @pacrob @seetadev , I am a little confused on how to write the tests for this. Like I tested it out by running nodes in` two separate codebases`. Without the timeout bound, if I create an infinite loop in the `write part of the handshake function` in one codebase like this:
```python
async def handshake(self, communicator: IMultiselectCommunicator) -> None:
        """
        Ensure that the client and multiselect are both using the same
        multiselect protocol.

        :param communicator: communicator to use to communicate with counterparty
        :raise MultiselectClientError: raised when handshake failed
        """
        try:
            await trio.sleep_forever()  ## Infinite loop
            await communicator.write(MULTISELECT_PROTOCOL_ID)
        except MultiselectCommunicatorError as error:
            raise MultiselectClientError() from error

        try:
            handshake_contents = await communicator.read()

        except MultiselectCommunicatorError as error:
            raise MultiselectClientError() from error

        if not is_valid_handshake(handshake_contents):
            raise MultiselectClientError("multiselect protocol ID mismatch")
```

, the connection goes into `deadlock`, but the same returns this error 
```
      | libp2p.exceptions.MultiError: Error 1: failed to upgrade security for peer QmNQgFYCn7Tey5shSNEpX96cbU982JjfYJvHAw4aZVnyKC
      | 
      | The above exception was the direct cause of the following exception:
      | 
      | Traceback (most recent call last):
      |   File "/home/shelby/Desktop/libp2p/py-libp2p/examples/ping/ping.py", line 75, in run
      |     await host.connect(info)
      |   File "/home/shelby/Desktop/libp2p/py-libp2p/libp2p/host/basic_host.py", line 243, in connect
      |     await self._network.dial_peer(peer_info.peer_id)
      |   File "/home/shelby/Desktop/libp2p/py-libp2p/libp2p/network/swarm.py", line 161, in dial_peer
      |     raise SwarmException(
      |     ...<2 lines>...
      |     ) from MultiError(exceptions)
      | libp2p.network.exceptions.SwarmException: unable to connect to QmNQgFYCn7Tey5shSNEpX96cbU982JjfYJvHAw4aZVnyKC, no addresses established a successful connection (with exceptions)
```
after `5 seconds which is the default negotiate timeout` with the timeout bound. So how should I approach this test, in a single `codebase` and `test_function`.

seetadev:
@lla-dane : Hi Abhinav.

Thanks for bringing this up and for sharing such a clear explanation — this is a great edge case to surface, and I really appreciate the thoughtful experimentation you've done with separate codebases. The example with the infinite loop in `handshake()` is super helpful for understanding how the timeout behavior kicks in under failure conditions.

Your observation about the `negotiate_timeout` kicking in after 5 seconds is spot-on, and it's a useful fail-safe that we definitely want to test for. I agree that writing this kind of test in a single codebase can feel a bit tricky, especially when simulating misbehavior from a peer.

Here’s one way you could approach it:

* You can simulate the “stuck peer” behavior by creating a mock or a dummy `IMultiselectCommunicator` implementation that delays indefinitely (or for a long time) before sending data.
* Then, call the actual handshake function using this mock communicator and assert that it raises a `MultiselectClientError` (or `SwarmException`, depending on the context), within the expected timeout window.
* That way, you're still covering the behavior in a single codebase and keeping it unit-testable, without spinning up actual peers or subprocesses.

Alternatively, if you want to simulate two real peers and test across a full connection attempt, you could use two `Host` instances within the same test using Trio nursery contexts. One could intentionally stall or misbehave in the handshake, and you’d assert that the other side fails gracefully within the timeout.

Both options are valid — one gives you a tighter unit test, and the other gives you broader integration-level assurance. Happy to help sketch either of these approaches out if you want!

Thanks again for diving deep into this — really valuable work that strengthens protocol robustness.

seetadev:
@lla-dane : This PR is very much close to final review + merge. Once you respond to my feedback, will review your changes once more and merge. Great work on the PR.

In the meantime, wish if you could add developer notes on how you approached the issue, the steps you took and how Py-libp2p developers can learn from your PR in a discussion page. Looking forward to more such contributions.

tests/core/protocol_muxer/test_negotiate_timeout.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self) -> None:
+        return
+
+    async def write(self, msg_str: str) -> None:
+        """Goes into infinite loop when .write is called"""
+        await trio.sleep_forever()
+
+    async def read(self) -> str:
+        """Returns a dummy read"""
+        return "dummy_read"
```
lla-dane, line 30: @seetadev : Thanks for the great suggestion. Created a `dummy multiselect communicator`, and set a infinte loop in the `write` function. Created the tests for all 3 functions that were changed across `multiselect_client.py` and `multiselect.py`.

lla-dane:
> In the meantime, wish if you could add developer notes on how you approached the issue, the steps you took and how Py-libp2p developers can learn from your PR in a discussion page. Looking forward to more such contributions.

For the developer notes, shall I create a new discussion page for this? and write the details of the approach in it? @seetadev

seetadev:
> > In the meantime, wish if you could add developer notes on how you approached the issue, the steps you took and how Py-libp2p developers can learn from your PR in a discussion page. Looking forward to more such contributions.
> 
> For the developer notes, shall I create a new discussion page for this? and write the details of the approach in it? @seetadev

@lla-dane :  Absolutely, that would be great, Abhinav 🙌 Yes, please go ahead and create a new discussion page specifically for the developer notes. It’ll be really helpful to document the approach, design rationale, and any implementation details or decisions you make along the way. This not only keeps everything organized and transparent, but also makes it easier for others to follow, contribute, or review later on.

Feel free to structure it however works best for you—high-level overview, breakdown of key components, trade-offs, anything that gives context. Looking forward to reading through it once it's up! Let me know if you want a second pair of eyes on anything too.

lla-dane:
@seetadev: Posted the implementation details in the form of developer notes in #739 [here](https://github.com/libp2p/py-libp2p/discussions/739#discussioncomment-13643517)

tests/core/protocol_muxer/test_negotiate_timeout.py
seetadev: @lla-dane : Awesome, great work Abhinav :) Appreciate your efforts.

seetadev:
@lla-dane : Nice work, Abhinav. Appreciate your efforts. Thank you for adding the requisite test cases and sharing the documentation.

Issue #701, opened 2025-06-24
fix: Yamux align stream creation logic with yamux specification


acul71:
### Summary

- The Python code only accepted new streams for `DATA+SYN` frames:
  ```python
  if typ == TYPE_DATA and flags & FLAG_SYN:
      # ... create stream ...
  ```
- The JavaScript client (and spec) can initiate streams with `WINDOW_UPDATE+SYN` frames as well.
- As a result, the Python host ignored valid stream creation attempts

### Expected behavior

Yamux should open a new stream even for  `WINDOW_UPDATE+SYN` frames as well

### Actual behavior

- The Python code only accepted new streams for `DATA+SYN` frames:

## The Fix

- The Python code is to be updated to accept new streams for both `DATA` and `WINDOW_UPDATE` frames with the `SYN` flag:
  ```python
  if (typ == TYPE_DATA or typ == TYPE_WINDOW_UPDATE) and flags & FLAG_SYN:
      # ... create stream ...
  ```
- After this change, the Python host will correctly accepted streams from the JS client

### Would you like to work on fixing this bug ?

Yes

PR #702, opened 2025-06-24
fix: accept new streams for both DATA and WINDOW_UPDATE frames with the SYN flag

acul71:
## What was wrong?
- The Python code only accepted new streams for `DATA+SYN` frames:
  ```python
  if typ == TYPE_DATA and flags & FLAG_SYN:
      # ... create stream ...
  ```
Issue https://github.com/libp2p/py-libp2p/issues/701

## How was it fixed?
- The Python code was updated to accept new streams for both `DATA` and `WINDOW_UPDATE` frames with the `SYN` flag:
  ```python
  if (typ == TYPE_DATA or typ == TYPE_WINDOW_UPDATE) and flags & FLAG_SYN:
      # ... create stream ...
  ```
  ## Relevant Specification

- **Yamux Spec:**
  - [yamux spec - Stream Creation](https://github.com/hashicorp/yamux/blob/master/spec.md#flag-field)
  - _"0x1 SYN - Signals the start of a new stream. May be sent with a data or window update message. Also sent with a ping to indicate outbound."_


acul71:
@seetadev @pacrob 
Ready for review.
Fix an issue related to py-js interop https://github.com/libp2p/py-libp2p/discussions/584#discussioncomment-13556913

pacrob (approved):
great!

mvds00:
I might create a PR one day of my current fix for this, but another bug in that area is that the yamux demultiplexer ignores data sent with FLAG_SYN (and also with FLAG_ACK), (maybe) in violation of the (vaguely worded) [spec](https://github.com/hashicorp/yamux/blob/master/spec.md) where it says:

> 0x1 SYN - Signals the start of a new stream. May be sent with a data or window update message.
> (etc)

The spec isn't clear about whether data may or may not be sent in DATA frames with flags set, but it seems safe to assume it is allowed, because it is not explicitly disallowed and it is technically possible (and some implementations indeed do it).

The bug is that you have this huge [if/elif/elif/elif chain](https://github.com/libp2p/py-libp2p/blob/74f4aaf136a022b5a8786bd7e57974b8f6033e7f/libp2p/stream_muxer/yamux/yamux.py#L548-L633) which effectively only checks DATA frames for data if no flags are set.

Issue #692, opened 2025-06-18
Feat: Implement Message Piggybacking in Gossipsub


sukhman-sukh:
### Description

`TODO: Go use 'sendRPC', which possibly piggybacks gossip/control messages.`
Add support for piggybacking control and gossip messages in Gossipsub by combining them into a single packet before sending, similar to the Go implementation’s sendRPC function.

### Motivation

Piggybacking reduces network overhead and improves efficiency by sending control (PRUNE, GRAFT) and gossip messages together, aligning the Python implementation with the Go reference and enhancing protocol performance.

### Current Implementation

- Gossip messages (publish messages) and control messages (PRUNE, GRAFT, etc.) are sent separately.
- There is no dedicated function to combine (piggyback) these messages into a single packet.

### Are you planning to do it yourself in a pull request ?

Yes

sukhman-sukh:
Please assign this issue to me.

seetadev:
@sukhman-sukh : assigned :) CCing @mystical-prog and @Khwahish29.

mystical-prog:
Thank you @seetadev 
for mentioning me on this one,

I will work closely with @sukhman-sukh and make sure that we ship quality code as this is a very core performance improvement issue.

seetadev:
Very welcome @mystical-prog  — appreciate your initiative as always 🙌 

Delighted to see you and @sukhman-sukh collaborating on this PR. It’s definitely an important performance area.

sukhman-sukh:
@seetadev, please close this issue.

seetadev:
@sukhman-sukh : Thank you for sharing :)

PR #704, opened 2025-06-26
Remove piggyback TODO from gossipsub

sukhman-sukh:
## What was wrong?
This issue was intended to fix the `TODO` in the gossipsub module to implement piggybacking. But later found out the code is already there.  

Issue #https://github.com/libp2p/py-libp2p/issues/692

## How was it fixed?
Removed the `TODO` tag from the module and added a comment pointing to piggyback implementation.

Summary of approach.

#### Cute Animal Picture

### DIFF
--- a/libp2p/pubsub/gossipsub.py
+++ b/libp2p/pubsub/gossipsub.py
@@ -251,7 +251,6 @@ class GossipSub(IPubsubRouter, Service):
             stream = self.pubsub.peers[peer_id]
             # FIXME: We should add a `WriteMsg` similar to write delimited messages.
             #   Ref: https://github.com/libp2p/go-libp2p-pubsub/blob/master/comm.go#L107
-            # TODO: Go use `sendRPC`, which possibly piggybacks gossip/control messages.
             try:
                 await stream.write(encode_varint_prefixed(rpc_msg.SerializeToString()))
             except StreamClosed:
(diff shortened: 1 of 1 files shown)


sukhman-sukh:
After analyzing the current code and Go implementation and discussing with @mystical-prog, we found that our implementation already combines all gossip and control RPCs into one, producing a piggyback effect. So, no need to change anything in the implementation.

mystical-prog:
Yep, we can remove the TODO comment as it is misleading and outdated 🙃

seetadev:
@sukhman-sukh , @mystical-prog : Thanks for the deep dive and clarification! 🙌  Appreciate your efforts.

Indeed, GossipSub protocol in py-libp2p already correctly piggybacks control messages (IHAVE, IWANT, GRAFT, PRUNE) along with data messages in a single RPC envelope. This aligns with the spec and ensures efficient propagation without additional overhead.

@sukhman-sukh : Appreciate you catching and cleaning up the outdated TODO as well. Merged! 💯

seetadev:
@sukhman-sukh , @mystical-prog : On the same note, let’s keep iterating — feel free to open issues or PRs if you come across any inconsistencies with the spec or missing gossipsub features in Py-libp2p. 

Also, go-libp2p is a good reference for us to follow from a gossipsub interop perspective. Working on Gossipsub PRs with interop in mind will be helpful in the near future.

PR #746, opened 2025-07-07
fixme/correct-type

Minimega12121:
## What was wrong?

Refers to "FIXME:  Replace with correct return type" in libp2p/abc.py

Issue #
The main issue was that the return type of `get_mux()` was not correctly specified — it returned `Any`, which made type checking and linting difficult. After updating the return type to `"Multiselect"`, related code using `get_mux().get_protocols()` in `identify.py` and `discovery.py` started showing linting issues due to potential `None` values in the returned tuple.

## How was it fixed?

- Updated `get_mux()` return type from `Any` to `"Multiselect"`.
- Implemented the `get_protocols()` method inside `Multiselect` to return a properly typed tuple of registered protocols.
- Updated `identify.py` to filter out `None` values from `get_protocols()` when converting protocols to strings.
- Updated `discovery.py` to:
  - Safely filter `None` values from `get_protocols()`
  - Convert string protocol values back to `TProtocol` instances for comparison.

These changes ensure better type safety and fix linter warnings/errors that arose due to stricter type expectations.
Please look into this @seetadev

### DIFF
--- a/libp2p/abc.py
+++ b/libp2p/abc.py
@@ -50,6 +50,11 @@ if TYPE_CHECKING:
         Pubsub,
     )
 
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from libp2p.protocol_muxer.multiselect import Multiselect
+
 from libp2p.pubsub.pb import (
     rpc_pb2,
 )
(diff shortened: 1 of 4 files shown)


pacrob:
Thanks, @Minimega12121! Please add a newsfragment to summarize the changes and a test to demonstrate that `get_protocols` is functioning as expected.

PR #710, opened 2025-06-27
Implement validation throttler for message validation in Pubsub

varun-r-mallya:
## What was wrong?
Async validators in the Pubsub module (libp2p.pubsub.pubsub.py) were executed without any form of throttling or concurrency control. This could result in excessive concurrent validator tasks being spawned under high message volume, potentially exhausting system resources or causing instability.

This was originally mentioned as a TODO in the codebase, and further discussed in [Discussion #647](https://github.com/libp2p/py-libp2p/discussions/647).

Issue #709 

## How was it fixed?
A new throttling system was introduced via a ValidationThrottler component:

Global Throttle: Limits the number of async validations running concurrently across all topics (default: 8192).

Per-Topic Throttle: Configurable semaphore-based limits on validators for individual topics (default: 1024).

Validation Queue: Bounded channel to queue incoming validation requests to prevent spikes (default: 32).

Worker Pool: Dedicated workers (default: 4) process messages and apply throttled validation logic.

Validators are now registered via Pubsub.set_topic_validator(...), and internally wrapped with throttling semantics.
This aligns py-libp2p more closely with go-libp2p-pubsub's validation model.

### To-Do

- [x] Clean up commit history
- [ ] Add or update documentation related to these changes
- [ ] Add entry to the [release notes](https://github.com/libp2p/py-libp2p/blob/main/newsfragments/README.md)

Issue #709, opened 2025-06-27
Implement throttle on async validators


varun-r-mallya:
### Description

Currently, the Pubsub validate message method lacks proper throttling control over concurrently running asynchronous validators. This can lead to unbounded concurrency during message validation in high-throughput environments, potentially causing memory pressure, degraded performance, or even system instability.

This issue proposes to introduce structured throttling mechanisms—global, per-topic, and queue-based—similar to those implemented in the Go-libp2p validation.go module.

A discussion on this can be found in https://github.com/libp2p/py-libp2p/discussions/647

### Motivation

Resource Safety: Without throttling, every incoming message can trigger an async validation coroutine, which, under high load, can lead to thousands of concurrent coroutines. This can saturate memory and CPU resources.

Fairness and Isolation: A single high-volume topic should not monopolize validator bandwidth. Per-topic throttling ensures fairer resource distribution.

Spike Protection: A bounded validation queue helps absorb traffic bursts and gives time for backpressure mechanisms to take effect.

Observability and Tunability: Aligning with Go-libp2p’s configuration model enables fine-tuning validator performance.

### Current Implementation

Current Implementation: 
Runs multiple validators concurrently using trio.gather, but without any throttling.

There is no global cap or queue on the number of concurrent validators.

No mechanism exists to prioritize or reject messages when the system is under load.

### Are you planning to do it yourself in a pull request ?

Yes

PR #755, opened 2025-07-10
TODO: throttle on async validators

lla-dane:
Fixed a `TODO: Implement throttle on async validators` in `libp2p/pubsub/pubsub.py::validate_msg()`.

Used semaphores to limit concurrency while running concurrent async_validators:

```py
semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS)

async def run_async_validator(func: AsyncValidatorFn) -> None:
    async with semaphore:
        result = await func(msg_forwarder, msg)
        results.append(result)

    for async_validator in async_topic_validators:
        nursery.start_soon(run_async_validator, async_validator)
```

### DIFF
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -701,10 +701,27 @@ class Pubsub(Service, IPubsub):
 
             async with trio.open_nursery() as nursery:
                 for async_validator in async_topic_validators:
-                    nursery.start_soon(run_async_validator, async_validator)
+                    nursery.start_soon(
+                        self._run_async_validator,
+                        async_validator,
+                        msg_forwarder,
+                        msg,
+                        results,
+                    )
 
             if not all(results):
                 raise ValidationError(f"Validation failed for msg={msg}")
+
+    async def _run_async_validator(
+        self,
+        func: AsyncValidatorFn,
+        msg_forwarder: ID,
+        msg: rpc_pb2.Message,
+        results: list[bool],
+    ) -> None:
+        async with self._validator_semaphore:
+            result = await func(msg_forwarder, msg)
+            results.append(result)
 
     async def push_msg(self, msg_forwarder: ID, msg: rpc_pb2.Message) -> None:
         """
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -692,12 +692,7 @@ class Pubsub(Service, IPubsub):
 
         if len(async_topic_validators) > 0:
             # Appends to lists are thread safe in CPython
-            results = []
-
-            async def run_async_validator(func: AsyncValidatorFn) -> None:
-                async with self._validator_semaphore:
-                    result = await func(msg_forwarder, msg)
-                    results.append(result)
+            results: list[bool] = []
 
             async with trio.open_nursery() as nursery:
                 for async_validator in async_topic_validators:
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -112,6 +112,7 @@ class Pubsub(Service, IPubsub):
 
     peer_receive_channel: trio.MemoryReceiveChannel[ID]
     dead_peer_receive_channel: trio.MemoryReceiveChannel[ID]
+    _validator_semaphore: trio.Semaphore
 
     seen_messages: LastSeenCache
 
@@ -146,6 +147,7 @@ class Pubsub(Service, IPubsub):
         msg_id_constructor: Callable[
             [rpc_pb2.Message], bytes
         ] = get_peer_and_seqno_msg_id,
+        max_concurrent_validator_count: int = MAX_CONCURRENT_VALIDATORS,
     ) -> None:
         """
         Construct a new Pubsub object, which is responsible for handling all
@@ -171,6 +173,7 @@ class Pubsub(Service, IPubsub):
         # Therefore, we can only close from the receive side.
         self.peer_receive_channel = peer_receive
         self.dead_peer_receive_channel = dead_peer_receive
+        self._validator_semaphore = trio.Semaphore(max_concurrent_validator_count)
         # Register a notifee
         self.host.get_network().register_notifee(
             PubsubNotifee(peer_send, dead_peer_send)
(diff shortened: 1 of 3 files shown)


lla-dane:
Will add the tests for this and make the concurrency limit configurable. By the mean time, please do check if the solution is correct @seetadev @pacrob

lla-dane:
Added the tests to check that the concurrency limit is respected. Please check i f everything is alright, and the default concurrency limit is correct. @pacrob @seetadev

paschal533:
Hi @lla-dane, This looks like a solid fix for the async validator throttling TODO, but there's a critical issue where the default parameter `limit: trio.Semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS)` creates a new semaphore on every call instead of sharing one globally, you should move the semaphore to be an instance variable in the `__init__` method like `self._validator_semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS)` and use that directly in the `run_async_validator` function. Also consider making the concurrency limit configurable via a constructor parameter, and the test could be simplified to focus more on the actual semaphore behavior rather than mocking the entire `validate_msg` method. Overall it's a good implementation that properly addresses resource exhaustion concerns, just needs that parameter fix to actually work correctly.

lla-dane:
Hey @paschal533
> Hi @lla-dane, This looks like a solid fix for the async validator throttling TODO, but there's a critical issue where the default parameter limit: trio.Semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS) creates a new semaphore on every call instead of sharing one globally, you should move the semaphore to be an instance variable in the __init__ method like self._validator_semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS) and use that directly in the run_async_validator function. Also consider making the concurrency limit configurable via a constructor parameter.

Did this, added a `self._validator_semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS)` in the `PubSub` constructor, and is also configurable via the constructor parameter.

>  the test could be simplified to focus more on the actual semaphore behavior rather than mocking the entire validate_msg method.

For this, added the concurrency checker part in the original `validate_msg` test. so now the `throttle and validate_msg` both are getting tested in the same test. 

Does this work @paschal533 ?

seetadev:
@lla-dane : Hi Abhinav. Thank you for submitting the PR. Appreciate your great efforts and initiative.

Wish to ask whether you got a chance to review Varun's efforts at https://github.com/libp2p/py-libp2p/discussions/647 and https://github.com/libp2p/py-libp2p/pull/710 (reference issue: https://github.com/libp2p/py-libp2p/issues/709 )

pacrob:
@lla-dane - Looking good! I'm concerned about copying so much code directly from `Pubsub` though. What do you think about extracting `run_async_validator`  out to a class method `_run_async_validator` and then just mocking that? You'd need to adjust arguments, but way less copied code in test.

paschal533:
> > Hi @lla-dane, This looks like a solid fix for the async validator throttling TODO, but there's a critical issue where the default parameter limit: trio.Semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS) creates a new semaphore on every call instead of sharing one globally, you should move the semaphore to be an instance variable in the **init** method like self._validator_semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS) and use that directly in the run_async_validator function. Also consider making the concurrency limit configurable via a constructor parameter.
> 
> Did this, added a `self._validator_semaphore = trio.Semaphore(MAX_CONCURRENT_VALIDATORS)` in the `PubSub` constructor, and is also configurable via the constructor parameter.
> 
> > the test could be simplified to focus more on the actual semaphore behavior rather than mocking the entire validate_msg method.
> 
> For this, added the concurrency checker part in the original `validate_msg` test. so now the `throttle and validate_msg` both are getting tested in the same test.
> 
> Does this work @paschal533 ?

This is a great approach. this definitely fixes the main issue, and integrating the concurrency checking into the existing `validate_msg` test is actually a cleaner solution than having separate tests. The combined test approach makes sense since you're testing that the throttling works correctly within the actual validation flow rather than in isolation. This should work well now. The semaphore will now properly limit concurrency across all calls to `validate_msg` on the same Pubsub instance, which is exactly what we want for resource management. Great job and Well done @lla-dane 👏

libp2p/pubsub/pubsub.py
```diff
@@ -680,23 +690,39 @@ async def validate_msg(self, msg_forwarder: ID, msg: rpc_pb2.Message) -> None:
...
+    async def _run_async_validator(
+        self,
+        func: AsyncValidatorFn,
+        msg_forwarder: ID,
+        msg: rpc_pb2.Message,
+        results: list[bool],
+    ) -> None:
+        async with self._validator_semaphore:
+            result = await func(msg_forwarder, msg)
+            results.append(result)
```
lla-dane, line 719: @pacrob: separated the `run_async_validator` method to a separate class method as you suggested. Now there is only this much duplicated code from the `pubsub.py` in the tests:

```py
async def mock_run_async_validator(
        self,
        func: AsyncValidatorFn,
        msg_forwarder: ID,
        msg: rpc_pb2.Message,
        results: list[bool],
    ) -> None:
        async with self._validator_semaphore:
            async with lock:
                state["concurrency_counter"] += 1
                if state["concurrency_counter"] > state["max_observed"]:
                    state["max_observed"] = state["concurrency_counter"]

            try:
                result = await func(msg_forwarder, msg)
                results.append(result)
            finally:
                async with lock:
                    state["concurrency_counter"] -= 1
```

lla-dane:
> @lla-dane - Looking good! I'm concerned about copying so much code directly from Pubsub though. What do you think about extracting run_async_validator out to a class method _run_async_validator and then just mocking that? You'd need to adjust arguments, but way less copied code in test.

@pacrob : Did a few changes as you suggested here. Please see if everything is alright.

libp2p/pubsub/pubsub.py
```diff
@@ -680,23 +690,39 @@ async def validate_msg(self, msg_forwarder: ID, msg: rpc_pb2.Message) -> None:
...
         if len(async_topic_validators) > 0:
             # Appends to lists are thread safe in CPython
             results = []
 
             async def run_async_validator(func: AsyncValidatorFn) -> None:
-                result = await func(msg_forwarder, msg)
-                results.append(result)
+                async with self._validator_semaphore:
+                    result = await func(msg_forwarder, msg)
+                    results.append(result)
```
pacrob: Since you extracted the logic out, this code can be deleted, right?

lla-dane: My bad, forgot to remove this. Now removed!!

pacrob (approved):

PR #758, opened 2025-07-13
feat: implement upgrade_listener and add Noise Pipes (IK handshake) support

sujalsalekar:
## What was changed?

This PR introduces two major improvements to the py-libp2p codebase:

### 1. `TransportUpgrader.upgrade_listener` Implementation
- The `upgrade_listener` method is now fully implemented.
- It wraps the base transport listener’s handler so that all incoming connections are automatically upgraded: first secured, then multiplexed, before being passed to the application handler.
- This ensures that all inbound connections are handled consistently and securely, matching the behavior of outbound connections.

### 2. Noise Pipes (IK Handshake) and Static Key Cache Support
- Added support for Noise Pipes, enabling the use of the IK handshake pattern when a remote peer’s static Noise key is known.
- Introduced a static key cache in the Noise transport to store remote peers’ static keys after a successful XX handshake.
- The transport will now use the XX handshake if no static key is cached for a peer, and switch to IK for subsequent connections.
- Added a new `PatternIK` class for the IK handshake, and updated both handshake patterns to expose the remote static key for caching.

## Why is this needed?

- **Upgrade Listener:** Ensures all inbound connections are properly secured and multiplexed, improving security and consistency.
- **Noise Pipes:** Improves connection performance and security by enabling the more efficient IK handshake when possible, and aligns with the libp2p spec for Noise Pipes.

### DIFF
--- a/libp2p/transport/upgrader.py
+++ b/libp2p/transport/upgrader.py
@@ -43,9 +43,37 @@ class TransportUpgrader:
         self.security_multistream = SecurityMultistream(secure_transports_by_protocol)
         self.muxer_multistream = MuxerMultistream(muxer_transports_by_protocol)
 
-    def upgrade_listener(self, transport: ITransport, listeners: IListener) -> None:
-        """Upgrade multiaddr listeners to libp2p-transport listeners."""
-        # TODO: Figure out what to do with this function.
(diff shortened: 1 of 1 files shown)


acul71:
@sujalsalekar Sorry `IK Patterns` were deprecated in specs, sorry for letting you know so late

PR #772, opened 2025-07-15
removal: private attribute for cached_property

LVivona:
## What was wrong?

We can clean up the ``libp2p.peer.ID`` class by using functools.cached_property. Since we are presuming the id is immutable, we don't need to cache it via an attribute, when we may use the functools ``cached_property`` decorator.

## How was it fixed?

```diff
+ import functools
import hashlib

import base58
import multihash

from libp2p.crypto.keys import (
    PublicKey,
)

# ...

class ID:
    _bytes: bytes
-   _xor_id: int | None = None
-   _b58_str: str | None = None

    def __init__(self, peer_id_bytes: bytes) -> None:
        self._bytes = peer_id_bytes

+  @functools.cached_property
-   @property
    def xor_id(self) -> int:
        return int(sha256_digest(self._bytes).hex(), 16)

+  @functools.cached_property
+    def base58(self) -> str:
+        return base58.b58encode(self._bytes).decode()

    def to_bytes(self) -> bytes:
        return self._bytes

    def to_base58(self) -> str:
-       if not self._b58_str:
-            self._b58_str = base58.b58encode(self._bytes).decode()
-      return self._b58_str
        return self.base58
```

### Summary of approach.

- removed private attributes for cached_property.
- added base58 property.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

cute_cat


LVivona:
I'm not sure if I would classify that workflow as an error of my making, due to it not being associated with the PeerID tests, rather the process being too slow, specifically on the windows machine.

seetadev:
@LVivona : Thanks for the great cleanup and thoughtful use of `functools.cached_property` in the `libp2p.peer.ID` class — this is a really nice refactor! 

You're absolutely right that since `peer_id` is immutable, `cached_property` is a clean and Pythonic way to avoid the boilerplate of manually managing and invalidating cached attributes like `_b58_str` or `_xor_id`. The changes improve clarity and reduce internal state that we previously had to track explicitly — definitely a solid move toward better maintainability.

I also appreciate how clearly you summarized:

* what was wrong (redundant internal caching),
* how it was fixed (replacing with `cached_property`), and
* your next steps (commit cleanup, docs, and release notes).

About the test failure in `test_trio_service_manager_run_task_waits_for_task_completion` — it looks unrelated to your changes. The traceback points to a timeout and cancellation (`trio.TooSlowError` and `trio.Cancelled`) inside the async service manager context. This specific test seems sensitive to event loop scheduling or platform-specific timing quirks (especially on Windows), so it might be worth re-running the job or marking it flaky for now. If we continue seeing it, we could isolate and investigate it separately to avoid blocking unrelated PRs like yours.

Please go ahead with the final polish — once the commit history is cleaned and the newsfragment is added, we’ll be good to merge 

Thanks again for modernizing this part of the codebase! Your continued efforts on these improvements are much appreciated.

CCing @acul71, @kaneki003 and @sumanjeet0012 for their thoughts and feedback.

seetadev:
@LVivona : Thank you Luca for considering my feedback and making requisite additions. Appreciate your efforts and initiative.

This PR is indeed ready for merge.

Issue #842, opened 2025-08-13
Kademlia Routing Table does not split buckets on overflow, limiting network scalability


sumanjeet0012:
### Summary

The Kademlia DHT RoutingTable implementation in libp2p/py-libp2p does not correctly split k-buckets when they overflow. Currently, all peers are stored in a single KBucket covering the full keyspace, and when the bucket is full, peers are only evicted if unresponsive. This prevents the RoutingTable from scaling efficiently or maintaining logarithmic bucket structure as described in the Kademlia paper.

### Expected behavior

When the RoutingTable grows and a bucket reaches its maximum BUCKET_SIZE, the bucket should be split into two, redistributing peers according to their key ranges. This allows the RoutingTable to maintain multiple buckets with smaller key ranges, supporting efficient lookups and scalability for large networks.

### Actual behavior

Currently, when a bucket is full, the implementation does not split the bucket. Instead, it attempts to evict the least-recently seen peer if unresponsive, otherwise the new peer is rejected. All peers remain in a single bucket covering the entire key range, so the RoutingTable never grows beyond one bucket. This breaks the expected logarithmic search efficiency and limits the network to only BUCKET_SIZE peers.

### Relevant log output

```shell

```

### Possible Solution

Refactor the RoutingTable and KBucket logic to support dynamic bucket splitting. When a bucket overflows and contains the local node's ID, split it into two buckets and redistribute peers. Update add_peer, find_bucket, and peer lookup methods to work with multiple buckets. This will ensure proper scaling and alignment with the Kademlia protocol.

### Environment

```shell

```

### Would you like to work on fixing this bug ?

Yes

PR #846, opened 2025-08-16
Fix: kbucket splitting in routing table.

sumanjeet0012:
## What was wrong?

Issue #842

The routing table was only maintaining a single kbucket, and the kbucket was not splitting when it became full. This prevented the routing table from scaling and properly organizing peers as specified by the Kademlia DHT protocol.

## How was it fixed?

- Updated the routing table logic to ensure that kbuckets split correctly when full, allowing the routing table to maintain multiple kbuckets and properly distribute peers.
- Added and updated tests to verify that bucket splitting occurs when more than the bucket size (20) peers are added.

### To-Do

- [x] Clean up commit history


seetadev:
@sumanjeet0012 : Great, thank you so much for submitting the PR. This is indeed ready for final review + merge. Appreciate the efforts, initiative and great support.

Issue #891, opened 2025-09-01
Add queuing to Gossipsub


sukhman-sukh:
### Description

As discussed in https://github.com/libp2p/py-libp2p/discussions/719 Issue:1, we need to have `Urgent`, queuing, message splitting and a function equivalent to `sendRPC` of go-libp2p in py-libp2p's gossipsub.

### Motivation

It will stabilize current gossipsub implementation by message splitting and queuing for retries and other important fallback on message failure. 

### Current Implementation

Currently, control messages are piggybacked and send to peers but queuing and message splitting is not present

### Are you planning to do it yourself in a pull request ?

Yes

PR #913, opened 2025-09-07
Improve gossipsub 

sukhman-sukh:
## What was wrong? 
Current version of gossipsub did not have message queuing, priority message queuing, and also, message splitting based on maxMessageSize as discussed for interoperability with go-libp2p.
Issue # https://github.com/libp2p/py-libp2p/issues/891

## How was it fixed?
 - Added a rpc_queue which handles message sharing and separates it from pubsub. 
 - Added message size limit
 -  RPC message transfer similar to go-libp2p

#### Cute Animal Picture


sukhman-sukh:
Hey @seetadev, 
As per the discussion thread #719, I have shifted our gossipsub write_msg to a message queuing and splitting architecture. Please review this, and is there anything else to add to this PR.
(I am unable to tag luca).

seetadev:
@sukhman-sukh : This is wonderful. Great work. Reviewing this in detail. Thank you so much for the contribution.

CCing @sumanjeet0012 , @lla-dane and @Winter-Soren , who are also working on different Gossipsub branches. Asking them to review this important PR with me. 

@acul71 (Luca) had some github admin issues. They are now resolved and you can definitely tag Luca again :)

libp2p/pubsub/gossipsub.py
```diff
@@ -852,11 +851,9 @@ async def handle_iwant(
                 sender_peer_id,
             )
             return
-        peer_stream = self.pubsub.peers[sender_peer_id]
 
         # 4) And write the packet to the stream
-        await self.pubsub.write_msg(peer_stream, packet)
-
+        await self.send_rpc(to_peer=sender_peer_id, rpc=packet)
```
lla-dane: The `urgent` bool parameter has to be mentioned here also right ?

lla-dane:
@sukhman-sukh: I had one question about the splitting logic, if an outgoing RPC is larger than `maxMessageSize`, it gets split into multiple smaller RPCs. On the receiving side, do we need to handle these fragments as parts of a larger message (i.e reassemble them), or is each split RPC a complete standalone message that can be processed independently?

sukhman-sukh:
Hey @lla-dane, according to me, every RPC is standalone. We were first piggybacking multiple messages into a single RPC which might exceed it sometime, so now it is taking that big RPC and breaking it into a standalone list of RPCs.

lla-dane:
Yeah I understand that. But suppose we break a big chunk RPC into individual RPCs, so isn't there a chamce that the content of those 2 smaller RPCs would be dependent, and since our ultimate goal was to somehow deliver the complete content of the big RPC on the receiver end, so the content of the 2 smaller RPCs need to agregated or something on the receiver end. 
Is it not the case ?
@sukhman-sukh

seetadev:
@sukhman-sukh : Hi Sukhman. Appreciate your efforts. Wish if you could resolve the CI/CD issues.

Wish if you could also respond to @lla-dane on his question/comment, whenever you get a chance.

CCing @Winter-Soren and @sumanjeet0012 for their thoughts and feedback too.

sukhman-sukh:
> Yeah I understand that. But suppose we break a big chunk RPC into individual RPCs, so isn't there a chamce that the content of those 2 smaller RPCs would be dependent, and since our ultimate goal was to somehow deliver the complete content of the big RPC on the receiver end, so the content of the 2 smaller RPCs need to agregated or something on the receiver end. Is it not the case ? @sukhman-sukh

Have a look here at go-libp2p impl (https://github.com/libp2p/go-libp2p-pubsub/blob/ed53c170f9a784c45f81d88dd6051218fbd5ece7/pubsub.go#L268). 
Basically, acc, to my understanding. Let's say we have a queue of RPCs -> <1,2,3,4> now we first piggyback them into <1234>  single RPC and then if size exceeds max then it will split it again into <12,34>. so the order is preserved for the same type of message. And also, afaik publish and then control message always works. If you have any edge cases in mind, please share so that we can improve the algorithm for greater correctness.

lla-dane:
Oh I see, I was thinking it was splitting a single RPC message into parts. That's good then, now I understood.

So would it possible for a single RPC message to exceed max size ?

sukhman-sukh:
> Oh I see, I was thinking it was splitting a single RPC message into parts. That's good then, now I understood.
> 
> So would it possible for a single RPC message to exceed max size ?

Yes, it is possible. In that case: 
```
for rpc in rpc_list:
                    if rpc.ByteSize() > max_message_size:
                        self.drop_rpc(rpc)
                        continue
                    await self.do_send_rpc(rpc, to_peer, urgent)
```
will be executed and that RPC will be dropped for now.

sukhman-sukh:
@seetadev, please re run the ci

seetadev:
@sukhman-sukh : We had some CI/CD performance issues over the last 7 hours.

Re-ran the CI/CD pipeline. We should now get test results soon.

IronJam11:
There are a few potential bugs that I encounterd @sukhman-sukh 

1. Infinite loop on oversized single items (all sections)

If a single item exceeds the limit, the splitting loop never advances.

Example from the publish section:

```
while i < n:
    new_rpc = base_rpc()
    size = 0
    j = i
    while j < n and size + incr_sizes[j] <= limit:
        size += incr_sizes[j]
        j += 1
    if j > i:
        new_rpc.publish.extend(publish_msgs[i:j])
        result.append(new_rpc)
    i = j
```

If incr_sizes[i] > limit:

* The inner loop never advances (j == i)
* Nothing is appended
* i = j leaves i unchanged
* The outer loop repeats forever

The same structural issue exists in graft, prune, iwant, and ihave sections where break exits the inner loop without advancing the index. Expected behavior (consistent with go-libp2p) is that oversized items are yielded as standalone RPCs and later dropped by the caller if necessary.

Fix pattern:

```
if j == i:
    new_rpc.publish.extend([publish_msgs[i]])
    result.append(new_rpc)
    i += 1
```

Equivalent logic should be applied to all control sections.

2. Double-append bug in IWant splitting

In the iwant section:

```
if size + incremental_size > limit:
    if len(new_iwant.messageIDs) > 1:
        new_iwant.messageIDs.pop()
        new_rpc.control.iwant.extend([new_iwant])
        result.append(new_rpc)
    break

if new_iwant.messageIDs:
    new_rpc.control.iwant.extend([new_iwant])
    result.append(new_rpc)
```

When the limit is exceeded and len > 1:

* new_iwant is appended inside the break path
* After break, the post-loop block still runs
* The same iwant is appended again
* The same RPC is appended to result twice

This results in duplicate RPCs in result and duplicated iwant entries inside the RPC.

Fix: introduce a flag or continue so the post-loop append is skipped when the break path already handled the append.

3. Missing final append in IHave splitting

In the ihave section, when the loop completes normally (not via break), the final batch is never appended:

```
while current_index < len(msg_ids):
    ...
    if size + incremental_size > limit:
        ...
        break
    ...
k = current_index
# No final append
```

If all remaining messageIDs fit within the limit:

* new_ihave contains valid IDs
* new_rpc is never appended
* The last batch is silently lost

Fix:

```
if new_ihave.messageIDs:
    new_rpc.control.ihave.extend([new_ihave])
    result.append(new_rpc)
```

4. Incorrect size tracking in control sections

In graft, prune, iwant, and ihave, size tracking is incorrect:

```
new_rpc.control.graft.extend([graft])
incremental_size = new_rpc.ByteSize()
if size + incremental_size > limit:
    ...
size += incremental_size
```

new_rpc.ByteSize() returns the total RPC size, not the incremental delta. This leads to double counting.

Example:

* First graft → RPC size = 120 → size = 120
* Second graft → RPC size = 220 → size = 120 + 220 = 340
* Actual size is only 220

This causes premature fragmentation and incorrect splitting.

Fix: remove the size accumulator and directly check total size:

```
if new_rpc.ByteSize() > limit:
```

This matches go-libp2p’s approach.

5. UnboundLocalError in send_rpc when pubsub is None

In send_rpc:

```
if self.pubsub:
    max_message_size = self.pubsub.maxMessageSize
if msg_size < max_message_size:
```

If self.pubsub is None, max_message_size is never assigned, causing UnboundLocalError.

Fix:

```
if self.pubsub is None:
    raise NoPubsubAttached

max_message_size = self.pubsub.maxMessageSize
```

CCing @seetadev @Winter-Soren

IronJam11:
@sukhman-sukh also can we switch to using deques for making it more efficient  ( O(n) -> O(1) )? @seetadev @Winter-Soren

IronJam11:
Right now, this task used "trio.lowlevel.spawn_system_task(self.handle_sending_message, peer_id, stream)" that is actually not inside any nursery i.e. Exceptions in handle_sending_message are not propagated to the parent. Consider using self.manager.run_daemon_task or a dedicated nursery.

acul71:
Closing as superseded by #1263 (rpc_queue / gossipsub improvements). This branch should not be revived as-is. Thanks for the early work.

PR #928, opened 2025-09-14
Fix multiaddr dep to use specific commit hash to resolve install issue

acul71:
Changed `multiaddr` dep to last py-multiaddr master commit
as previous commit was in branch `new_install` that wasn't available anymore


acul71:
@pacrob @seetadev 
Please merge ASAP

PR #1054, opened 2025-11-21
kad-dht example key fix

hayotensor:
## What was wrong?

The Kademlia DHT example uses:

`val_key = create_key_from_binary(b"py-libp2p kademlia example value")`

## How was it fixed?

`create_key_from_binary()` returns a multihash-encoded value, which is arbitrary binary data.

This returns a multihash-encoded hash, which is binary data (not valid UTF-8) and generates the error: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd7 in position 0: invalid continuation byte.

This causes the example to fail because the resulting key is not valid UTF-8, leading to:

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd7 in position 0: invalid continuation byte`

This prevents both the KAD server from storing the DHT record.

**Further:**

Added `examples/kademlia/server_node_addr.txt` to the `.gitignore`. This generates a server node address and stores it there.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


hayotensor:
Closing because https://github.com/libp2p/py-libp2p/pull/1060

PR #1055, opened 2025-11-22
Fix transport layer

codemaestro64:
## What was wrong?

Closes #726

Passing a `trio.Nursery` into `listen()` was messy. It forced the `Swarm` to micro-manage the lifecycle of every transport listener, which leaked implementation details and made the API awkward to use. Simply put, the `Swarm` was doing work that the Listener should have been doing itself.

## How was it fixed?

This PR moves the nursery management *inside* the listeners.

Instead of asking for a nursery, `TCPListener`, `WebsocketListener`, and `QUICListener` now spin up their own internal nurseries using system tasks. This keeps the concurrency logic encapsulated where it belongs.

**Changes:**
* **Cleaner API:** The `nursery` argument was dropped from `IListener.listen(maddr)`.
* **Self-contained Transports:** Listeners now handle their own background loops. Crucially, this PR ensures `close()` properly cancels these internal tasks so we don't leave orphans hanging around.


yashksaini-coder:
Hi @codemaestro64 — I've adopted the work from this PR in #1308 to get Issue 1 of discussion #726 moving again.

Every refactor commit carries `Co-authored-by: Michael Eze <code.maestro64@gmail.com>` so your authorship of the approach is preserved in history. The system-task + internal-nursery pattern is yours; I've rebased on current main (which had drifted — `listen()` now returns `None` and raises `OpenConnectionError` rather than returning `bool`), captured startup exceptions properly so they surface to the caller, added the missing newsfragment, and made `close()` await the background task's completion.

Full list of changes and rationale in the PR description. Would appreciate your review when you have time.

acul71:
Closing as adopted by #1308 (`fix(transport): remove nursery param from IListener.listen`). Thank you — your work was credited there.

PR #1060, opened 2025-11-24
Migrate DHT API to accept string keys.

### LINKED ISSUE #1059: Kademlia DHT example not working.
### Summary

The current Kademlia DHT API requires `bytes` for keys, which causes UnicodeDecodeError when using binary multihash keys. The kademlia example (`examples/kademlia/kademlia.py`) is currently broken due to this issue. This differs from Go libp2p's design and makes the API less user-friendly.

### Expected behavior

The kademlia DHT example should work without any issue.

### Actual behavior

The kademlia example (`examples/kademlia/kademlia.py`) crashes with:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 1: invalid start byte
```

This occurs because create_key_from_binary() generates a 34-byte multihash containing non-UTF-8 bytes, which the DHT tries to decode.

### Relevant log output

```shell

```

### Possible Solution

Remove the unnecessary UTF-8 decoding step. Instead, use the string directly as the key, allowing the system to handle the internal string-to-binary conversion for storing values in the value store.

### Environment

```shell

```

### Would you like to work on fixing this bug ?

Yes

sumanjeet0012:
## What was wrong?

Fixes #1059

The Kademlia DHT example was broken due to UnicodeDecodeError when using `create_key_from_binary()` with multihash keys. The DHT API required `bytes` keys but internally called `.decode("utf-8")` on them, which failed for binary multihash values (34 bytes containing non-UTF-8 data).

## How was it fixed?

This PR migrates the Kademlia DHT API from `bytes` to `str` keys, following the Go libp2p design pattern. The API now accepts string keys (e.g., `put_value(key: str, value: bytes)`) which are encoded to bytes internally using UTF-8, avoiding the problematic decode operation. This maintains binary transmission over the wire while providing a developer-friendly string interface.

#### 1. **Core API Changes** (`libp2p/kad_dht/kad_dht.py`)
- ✅ Changed `put_value(key: bytes, ...)` → `put_value(key: str, ...)`
- ✅ Changed `get_value(key: bytes)` → `get_value(key: str)`
- ✅ Changed `provide(key: bytes)` → `provide(key: str)`
- ✅ Changed `find_providers(key: bytes, ...)` → `find_providers(key: str, ...)`
- ✅ All methods convert `key.encode("utf-8")` internally for storage
- ✅ Simplified validator checks: `key.startswith("/")` instead of `key.decode(...).startswith("/")`

#### 2. **Internal Storage** (`libp2p/kad_dht/value_store.py`)
- ✅ `ValueStore` continues to use `bytes` keys internally
- ✅ Fixed `make_put_record()` import to avoid circular dependency

#### 3. **Record Utilities** (`libp2p/records/record.py`)
- ✅ Changed `make_put_record(key: str, ...)` → `make_put_record(key: bytes, ...)`
- ✅ Removed unnecessary `.encode()` call since internal APIs use bytes

#### 4. **Example Updates** (`examples/kademlia/kademlia.py`)
- ✅ Simplified to use plain string keys: `key = "my-example-key"`
- ✅ Removed confusing `create_key_from_binary()` usage from basic example
- ✅ Updated to use `dht.provide()` and `dht.find_providers()` convenience methods
- ✅ Demonstrates both key-value storage and provider advertisement

#### 5. **Test Updates** (`tests/core/kad_dht/`)
- ✅ Updated all tests to use string keys: `key = "random_key"`
- ✅ Fixed validator tests to pass strings directly
- ✅ Updated unit tests for `make_put_record()` signature change
- ✅ All 119 DHT tests passing

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/libp2p/pubsub/pubsub.py
+++ b/libp2p/pubsub/pubsub.py
@@ -481,6 +481,17 @@ class Pubsub(Service, IPubsub):
 
         logger.debug("added new peer %s", peer_id)
 
+    async def _handle_new_peer_safe(self, peer_id: ID) -> None:
+        """
+        Safely handle new peer with exception handling.
+        This wrapper ensures that any exceptions during peer negotiation
+        don't crash the entire pubsub service.
+        """
+        try:
+            await self._handle_new_peer(peer_id)
+        except Exception as error:
+            logger.info(f"Protocol negotiation failed for peer {peer_id}: {error}")
+
     def _handle_dead_peer(self, peer_id: ID) -> None:
         if peer_id not in self.peers:
             return
@@ -503,8 +514,8 @@ class Pubsub(Service, IPubsub):
         async with self.peer_receive_channel:
             self.event_handle_peer_queue_started.set()
             async for peer_id in self.peer_receive_channel:
-                # Add Peer
-                self.manager.run_task(self._handle_new_peer, peer_id)
+                # Add Peer - wrap in exception handler to prevent service crash
+                self.manager.run_task(self._handle_new_peer_safe, peer_id)
 
     async def handle_dead_peer_queue(self) -> None:
         """
--- a/newsfragments/910.bugfix.rst
+++ b/newsfragments/910.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed pubsub service crashes when protocol negotiation fails by adding proper exception handling.
--- a/tests/core/pubsub/test_pubsub.py
+++ b/tests/core/pubsub/test_pubsub.py
@@ -1277,3 +1277,29 @@ async def test_get_message_id():
             expected_custom_id = msg.data + msg.from_id
 
             assert custom_msg_id == expected_custom_id
+
+
+@pytest.mark.trio
+async def test_handle_peer_queue_exception_handling():
+    """Test that _handle_new_peer_safe gracefully handles exceptions."""
+    async with PubsubFactory.create_batch_with_floodsub(1) as pubsubs_fsub:
+        pubsub = pubsubs_fsub[0]
+
+        original_handle_new_peer = pubsub._handle_new_peer
+
+        async def mock_handle_new_peer(peer_id):
+            raise Exception("Protocol negotiation failed")
+
+        pubsub._handle_new_peer = mock_handle_new_peer
+
+        test_peer = IDFactory()
+
+        # Directly call the safe wrapper that's used by handle_peer_queue
+        await pubsub._handle_new_peer_safe(test_peer)
+
+        # The key test: service should still be running despite the exception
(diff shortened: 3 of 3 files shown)


seetadev:
@sumanjeet0012 : Great work, Sumanjeet. Appreciate your contribution.

Kindly let me know once the issue is fixed. Will do a final review + merge.

seetadev:
@sumanjeet0012 : Great work, Sumanjeet. Wish if you could arrive at just 1 CI/CD issue that is failing. Should be a quick fix. We had similar issues in Gossipsub. Soham made a fix that helped us arrive at a good conclusion on similar issues.

=========================== short test summary info ============================
FAILED tests/core/kad_dht/test_kad_dht.py::test_reissue_when_listen_addrs_change
====== 1 failed, 1611 passed, 8 skipped, 5 warnings in 673.52s (0:11:13) =======
py312-core: exit 1 (675.18 seconds) /home/runner/work/py-libp2p/py-libp2p> pytest -n auto --timeout=1200 tests/core pid=2390
  py312-core: FAIL code 1 (702.88=setup[27.69]+cmd[675.18] seconds)
  evaluation failed :( (703.62 seconds)
Error: Process completed with exit code 1.

seetadev:
Fantastic work on this, @sumanjeet0012 — this is a clean and much-needed fix for the long-standing key-handling issue in the Kademlia DHT. 🚀

The underlying bug (#1059) stemmed from calling .decode("utf-8") on binary multihash keys, which inevitably caused UnicodeDecodeError for non-UTF-8 bytes. Your migration of the DHT API from bytes → str keys, aligned with the Go libp2p design, cleanly resolves the problem without altering the binary protocol or wire format.

Really appreciate the thoroughness of the update:

- API Modernization

Moving all DHT-facing methods to accept str keys is the right long-term design.

Clean internal conversion (key.encode("utf-8")) keeps everything safe, explicit, and predictable.

Validator cleanup and simplified checks improve readability.

- Internals & Records

Maintaining bytes internally in ValueStore is correct.

Fixing the import for make_put_record() avoids circular dependencies — nice catch.

Adjusting record utilities to accept bytes ensures the API layers stay clean.

- Examples & UX

Updating the example to stop using create_key_from_binary() is a huge improvement for developer onboarding.

Using plain string keys matches expectations from Go, JS, and Rust examples.

- Tests

All DHT tests updated cleanly to string keys.

119 tests passing is excellent, and the small signature adjustment was handled gracefully.

The CI/CD issue you tackled was the only blocker, and now that everything is green, this PR is in great shape to merge.

Merging this — thanks again for the careful implementation and for pushing this to the finish line! 🙌

PR #1068, opened 2025-11-30
Implement fallback mechanism in Kademlia DHT for peer lookup.

sumanjeet0012:
## What was wrong?

Issue #905

This PR fixes issue #905 by adding a fallback mechanism in the Kademlia DHT implementation to use connected peers and peerstore peers when the routing table has insufficient peers.

The current Kademlia DHT implementation has a bootstrap problem that prevents proper peer discovery during the initial network join phase:

1. Bootstrap connects to peers → Peers added to peerstore
2. DHT queries attempt to start → `find_local_closest_peers()` returns empty/insufficient list (routing table is empty or has few peers)
3. Random walks and peer discovery fail → No peers available for DHT queries
4. Network isolation → Node cannot discover new peers beyond bootstrap connections

## How was it fixed?

Added a fallback mechanism in `find_closest_peers_network()` method that uses connected peers and peerstore peers when the routing table has fewer peers than `MIN_PEERS_THRESHOLD` (default: 5):

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,11 @@ coverage.xml
 *.cover
 .pytest_cache/
 
+# Test result files
+tests/**/results/*.json
+tests/**/results/*.xml
+tests/**/results/*.html
+
 # Translations
 *.mo
 *.pot
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -310,7 +310,7 @@ autodoc_mock_imports = [
     "tests.factories",
     # Mocked ONLY for Sphinx/autodoc: this module does not exist in the codebase
     # but some doc tools may try to import it. No real code references this import.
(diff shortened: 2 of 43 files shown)


seetadev:
@sumanjeet0012 : Merging this PR — great work! 🚀

Thanks for the clear fix for #905 and for strengthening the Kademlia DHT bootstrap flow. The fallback to connected + peerstore peers is exactly what was needed to prevent early-stage isolation and enable reliable discovery when the routing table is still warming up. Reviewed the changes, and PR looks solid.

Appreciate the contribution and the thoughtful explanation. Looking forward to the follow-up items on docs + release notes.

Issue #910, opened 2025-09-06
Pubsub service crashes on protocol negotiation failures


sumanjeet0012:
### Summary

The pubsub service currently crashes when protocol negotiation fails during peer handling. In the `_handle_new_peer_queue method`, unhandled exceptions from `_handle_new_peer` cause the entire pubsub service to stop processing new peers.

### Expected behavior

The exception should be handled gracefully when protocol negotiation fails. In such cases, a basic TCP raw connection with the peer should be maintained, even if the PubSub protocol negotiation is unsuccessful.

### Actual behavior

```
async for peer_id in self.peer_receive_channel:
    self.manager.run_task(self._handle_new_peer, peer_id)  # Can throw unhandled exceptions
```

When _handle_new_peer fails (due to protocol negotiation failures), the exception propagates up and crashes the pubsub service, making it unable to handle any future peer connections.

### Relevant log output

```shell

```

### Possible Solution

Simple try expect block can also be used to handle the exception.

```
async for peer_id in self.peer_receive_channel:
    try:
        self.manager.run_task(self._handle_new_peer, peer_id)
    except Exception as e:
        logger.info(f"Protocol negotiation failed: {e}")
```

### Environment

```shell

```

### Would you like to work on fixing this bug ?

Maybe

PR #1069, opened 2025-11-30
Add exception handling to prevent pubsub service crashes.

sumanjeet0012:
## What was wrong?

Issue #910 

## How was it fixed?

1. **Minimal Invasive Fix:**
   - Added only a try-except block around the problematic line
   - No changes to the existing API or method signatures
   - Preserved all existing functionality

2. **Service Continuity:**
   - The pubsub service now continues running even when individual peer negotiations fail
   - TCP connections are maintained even if PubSub protocol negotiation fails
   - Future peer connections can still be processed
   
### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,11 @@ coverage.xml
 *.cover
 .pytest_cache/
 
+# Test result files
+tests/**/results/*.json
+tests/**/results/*.xml
+tests/**/results/*.html
+
 # Translations
 *.mo
 *.pot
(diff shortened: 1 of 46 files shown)


seetadev:
Thanks for the fix, @sumanjeet0012 — this is a clean and well-scoped solution. 🚀

The issue in #910 was causing unnecessary crashes in the pubsub service during failed protocol negotiation, and your approach addresses it exactly as needed: a minimal, non-invasive try/except that preserves the existing API and keeps the service stable under partial failures.

Really appreciate that the fix:

- avoids any behavior regressions

- keeps TCP connections alive

- ensures the pubsub loop continues serving new peers

- isolates the failure to just the problematic negotiation path

This is the right lightweight patch for now, and it unblocks downstream components depending on pubsub stability.

We can take the remaining items (docs + release notes + cleanup) in a follow-up.

Merging this — great work!

PR #1088, opened 2025-12-11
Fix Dockerfile: install py-multihash before libp2p to resolve dependency issues

acul71:
This PR fixes a dependency resolution issue in the Dockerfile by ensuring py-multihash is installed from git before installing libp2p. This prevents pip from failing to resolve git dependencies correctly when installing from a local file.


seetadev:
@acul71 : Thanks a lot for this fix — it’s a clean and much-needed improvement. Installing py-multihash ahead of libp2p in the Dockerfile is exactly the right call, and it finally resolves the dependency resolution issue we kept running into with pip when mixing local paths and git installs.

The updates look solid to me, and the commits are clear and well-scoped. If everything looks good on your side as well, let’s go ahead and merge this in so we can unblock the rest of the pipeline and keep the transport work moving forward.

Great work — really appreciate the quick turnaround on this!

acul71:
> @acul71 : Thanks a lot for this fix — it’s a clean and much-needed improvement. Installing py-multihash ahead of libp2p in the Dockerfile is exactly the right call, and it finally resolves the dependency resolution issue we kept running into with pip when mixing local paths and git installs.
> 
> The updates look solid to me, and the commits are clear and well-scoped. If everything looks good on your side as well, let’s go ahead and merge this in so we can unblock the rest of the pipeline and keep the transport work moving forward.
> 
> Great work — really appreciate the quick turnaround on this!

@seetadev This was a quic fix overdue, to make the interop test-plans transport test go, is should not be needed .
I'll close this when It will be sure, in a bit

seetadev:
@acul71 : Thank you so much Luca. Sure, please take your time. Appreciate your efforts.

acul71:
Closing this as fixed with new `py-multihash` release

PR #1018, opened 2025-11-04
Feat/advance connection management

yashksaini-coder:
## What was wrong?

Python libp2p was missing advanced connection management features present in JavaScript libp2p, including:
- Priority-based dial queue management
- Automatic reconnection for KEEP_ALIVE peers
- Global connection limits and intelligent pruning
- Rate limiting for incoming connections
- IP allow/deny list filtering
- Address sorting and DNS resolution
- Comprehensive connection metrics
- Enhanced connection state tracking

## How was it fixed?

Implemented a complete connection management system matching JavaScript libp2p's capabilities:

### Core Components Added:
- **DialQueue**: Priority-based scheduling with concurrency limits and queue size management
- **ReconnectQueue**: Automatic reconnection with exponential backoff for KEEP_ALIVE tagged peers
- **ConnectionPruner**: Intelligent connection pruning based on peer tags, stream count, direction, and age
- **ConnectionRateLimiter**: Per-host rate limiting using sliding window algorithm
- **ConnectionGate**: IP allow/deny list filtering with CIDR support
- **AddressManager**: Address sorting and filtering (transport priority, public/private, circuit relay)
- **DNSResolver**: Async DNS resolution for `/dns4/`, `/dns6/`, and `/dnsaddr/` multiaddrs with caching
- **ConnectionState**: Enhanced connection status tracking (PENDING, OPEN, CLOSING, CLOSED) with timeline

### Configuration Enhancements:
Extended `ConnectionConfig` with 17+ new options matching JS libp2p defaults:
- Global connection limits (`max_connections`, `max_parallel_dials`, etc.)
- Comprehensive timeout configuration (`dial_timeout`, `inbound_upgrade_timeout`, etc.)
- Reconnection parameters (`reconnect_retries`, `reconnect_backoff_factor`, etc.)
- Security lists (`allow_list`, `deny_list` with CIDR support)
- Custom address sorting

### Integration:
- Fully integrated all components into `Swarm` class
- Maintains backward compatibility with existing APIs
- All components follow trio async patterns
- Comprehensive error handling and logging

### Testing:
- 24 unit tests covering all new components
- All tests passing ✅
- Type checking passing ✅
- Linting passing ✅

### Summary of approach:
1. Analyzed JS libp2p connection manager implementation
2. Implemented components in phases following dependency order
3. Integrated components into Swarm with proper lifecycle management
4. Added comprehensive tests and fixed all type/lint issues

---

@seetadev @sumanjeet0012 Raised a base PR to help add more advance connection management features.

PR #1092, opened 2025-12-13
feat: advance connection management checkout

### LINKED ISSUE #553: Implement Advanced Connection Management Features from JS libp2p
### Description

# Missing Features in Python's Connection Management Compared to JavaScript

After analyzing the JavaScript implementation https://github.com/libp2p/js-libp2p/blob/d53ef170cb171f5301758d5b2fc9e782950b4204/packages/libp2p/src/connection-manager/index.ts and comparing it with the Python codebase, here are the key features that are missing in the Python implementation:

## Advanced Connection Management Features
- **Multiple connections per peer support** (JavaScript uses `PeerMap<Connection[]>` while Python uses a simple `dict[ID, INetConn]`)
- **Connection limits and resource management** with configurable thresholds
- **Connection pruning mechanism** to manage resource usage
- **Rate limiting** for incoming connections

## Connection Queue Management
- Sophisticated **dial queue** with priority-based scheduling
- **Reconnection queue** with configurable retry policies
- **Maximum queue length limits**
- **Parallel dial limits**

## Connection Configuration Options
Extensive configuration options for timeouts:
- `dialTimeout`
- `inboundUpgradeTimeout`
- `outboundUpgradeTimeout`
- `protocolNegotiationTimeout`
- `outboundStreamProtocolNegotiationTimeout`
- `inboundStreamProtocolNegotiationTimeout`

Connection limits:
- `maxConnections`
- `maxParallelDials`
- `maxDialQueueLength`
- `maxPeerAddrsToDial`
- `maxIncomingPendingConnections`

## Security and Access Control
- **Allow/deny lists** for IP addresses
- **Connection gating** based on IP networks
- **Rate limiting** for incoming connections per host

## Metrics and Monitoring
Detailed metrics tracking for:
- **Connection counts** (inbound/outbound)
- **Pending connections**
- **Protocol streams per connection**
- **90th percentile metrics for streams**

## Reconnection Management
- **Automatic reconnection** for peers tagged with `KEEP_ALIVE`
- **Configurable retry policies** with backoff
- **Maximum parallel reconnects limit**

## Connection State Management
- More **granular connection state tracking**
- Better handling of **connection lifecycle events**
- **Improved error handling** and connection cleanup

## Address Management
- **Address sorting** before dialing
- **DNS resolver support**
- **Thin waist address handling**

## Event System
- **More comprehensive event system** for connection state changes
- Better event handling for **connection lifecycle events**

## Resource Management
- Better handling of **pending connections**
- More sophisticated **connection pruning**
- **Resource limits enforcement**

### Motivation

The Python implementation is more basic and focuses on core functionality, while the JavaScript version provides a more complete and feature-rich connection management system. The JavaScript version also has better **resource management, monitoring capabilities, and configuration options**.

### Current Implementation

This Python implementation of libp2p has similar connection management functionality to the JavaScript implementation

# Swarm Class in py-libp2p: Key Components and Similarities

## Core Connection Management
The `Swarm` class in `libp2p/network/swarm.py` serves as the main connection manager, similar to the JavaScript `connection-manager.ts`. It handles:
- **Opening connections** (`dial_peer`, `dial_addr`)
- **Closing connections** (`close_peer`)
- **Managing active connections** (stored in `self.connections` dictionary)
- **Connection lifecycle management**

## Connection Querying
The `Swarm` class provides methods to:
- Get active connections through `self.connections`
- Check connection status
- Filter connections by peer ID

## Connection Management
The `Swarm` class handles:
- Opening new connections via `dial_peer` and `dial_addr`
- Closing connections via `close_peer`
- Managing both direct peer ID and multiaddr-based connections

## Connection Limits and Resources
The implementation includes:
- Connection tracking through the `connections` dictionary
- Connection cleanup and resource management
- Connection state management through events

## Additional Features
The `SwarmConn` class in `libp2p/network/connection/swarm_connection.py` provides:
- **Stream management**
- **Connection state tracking**
- **Event-based notifications** for connection state changes

## Integration Points
The `Swarm` class works with:
- `MultiaddrConnection` for low-level connection handling
- `PeerId` for peer identification
- `PeerStore` for peer information management
- Transport layer for actual connection establishment

The connection management is more tightly integrated with the `Swarm` class rather than being a separate connection manager
The Python implementation uses a simpler connection model where each peer ID maps to a single connection, while the JavaScript version allows multiple connections per peer ID

Maybe

yashksaini-coder:
## How was it fixed?

Implemented a complete connection management system matching JavaScript libp2p's capabilities:

### Core Components Added:
- **DialQueue**: Priority-based scheduling with concurrency limits and queue size management
- **ReconnectQueue**: Automatic reconnection with exponential backoff for KEEP_ALIVE tagged peers
- **ConnectionPruner**: Intelligent connection pruning based on peer tags, stream count, direction, and age
- **ConnectionRateLimiter**: Per-host rate limiting using sliding window algorithm
- **ConnectionGate**: IP allow/deny list filtering with CIDR support
- **AddressManager**: Address sorting and filtering (transport priority, public/private, circuit relay)
- **DNSResolver**: Async DNS resolution for `/dns4/`, `/dns6/`, and `/dnsaddr/` multiaddrs with caching
- **ConnectionState**: Enhanced connection status tracking (PENDING, OPEN, CLOSING, CLOSED) with timeline

### Configuration Enhancements:
Extended `ConnectionConfig` with 17+ new options matching JS libp2p defaults:
- Global connection limits (`max_connections`, `max_parallel_dials`, etc.)
- Comprehensive timeout configuration (`dial_timeout`, `inbound_upgrade_timeout`, etc.)
- Reconnection parameters (`reconnect_retries`, `reconnect_backoff_factor`, etc.)
- Security lists (`allow_list`, `deny_list` with CIDR support)
- Custom address sorting

### Integration:
- Fully integrated all components into `Swarm` class
- Maintains backward compatibility with existing APIs
- All components follow trio async patterns
- Comprehensive error handling and logging

### Summary of approach:
1. Analyzed JS libp2p connection manager implementation
2. Implemented components in phases following dependency order
3. Integrated components into Swarm with proper lifecycle management
4. Added comprehensive tests and fixed all type/lint issues


yashksaini-coder:
### @seetadev @sumanjeet0012 @acul71 This PR is backed up from the original #1018 which got corrupt after resolving the commits. Raised this after the latest main branch, and the only difference in this branch is the git commit history. 

> [!NOTE]
> The failing interop tests are due to `requests` library not installed. There are no changes in the `pyproject.toml`

sumanjeet0012:
@seetadev @acul71 Ready for review.

seetadev:
@sumanjeet0012 and @yashksaini-coder : This is a **substantial and very well-executed piece of work** — huge thanks for the persistence, depth, and care that’s gone into this PR.

You’ve effectively delivered a **full-fledged advanced connection management stack** for py-libp2p that closely mirrors the JavaScript implementation while still respecting Python and Trio idioms. 

The breakdown into clear, composable components (DialQueue, ReconnectQueue, ConnectionPruner, RateLimiter, Gate, AddressManager, DNS resolver, and explicit connection state tracking) is especially strong, and the way everything is cleanly integrated into `Swarm` with lifecycle awareness shows real architectural maturity. 

Appreciate the discipline around backward compatibility, structured configuration, comprehensive defaults, and the amount of test coverage added along the way — that’s not easy at this scale.


@pacrob : Wish to recommend this PR for your review. Would love to have your feedback, particularly on the pruning heuristics, reconnection strategy, and long-lived swarm behavior. 

With that final pass, I think this will be a major milestone for py-libp2p and unblock a whole class of real-world deployments.

@sumanjeet0012 and @yashksaini-coder: Excellent work by both of you — this is exactly the kind of contribution that moves the ecosystem forward.

pacrob:
The example could use a page in `docs` to help users find it.

libp2p/network/swarm.py
```diff
@@ -695,37 +1155,24 @@ async def upgrade_inbound_raw_conn(
         :raises SwarmException: raised when security or muxer upgrade fails
         :return: network connection with security and multiplexing established
         """
```
pacrob, line 1193: Do we want to check allow/deny lists when handling a new inbound connection?

libp2p/network/swarm.py
```diff
@@ -140,6 +192,7 @@ async def run(self) -> None:
         async with trio.open_nursery() as nursery:
             # Create a nursery for listener tasks.
             self.listener_nursery = nursery
+            self.event_listener_nursery_created.set()
```
pacrob: Moving this up here contradicts the comment directly below by setting the event before the background nursery.

libp2p/network/connection_gate.py
```diff
@@ -0,0 +1,302 @@
...
+        except ValueError:
+            return False
+
+        for allow_net in self.allow_list:
+            if ip in allow_net:
+                return True
+
+        return False
+
+    def is_allowed(self, remote_addr: Multiaddr) -> bool:
```
pacrob: This will reject any non-IP Multiaddrs, such as `/dns4`, `/dns6`, `/dnsaddr`, or `/p2p-circuit`. Is that the desired behavior?

libp2p/network/swarm.py
yashksaini-coder: Moved event_listener_nursery_created.set() to after the background nursery is set on the transport, so the event is only signaled once the nursery is configured.

```python
    async def run(self) -> None:
            # Create a nursery for listener tasks.
            self.listener_nursery = nursery

            # Set background nursery BEFORE setting the event
            # This ensures transports have the nursery when they check
            if isinstance(self.transport, QUICTransport):
                self.transport.set_background_nursery(nursery)
                self.transport.set_swarm(self)
            elif hasattr(self.transport, "set_background_nursery"):
                # WebSocket transport also needs background nursery
                # for connection management
                self.transport.set_background_nursery(nursery)  # type: ignore[attr-defined]

            # Set event after background nursery is configured
            # This ensures transports have the nursery when they check the event
            self.event_listener_nursery_created.set()
```

libp2p/network/swarm.py
yashksaini-coder: @pacrob This design and security check was addressed in the original issue raised by @acul71 So we based our main design approach on that. Here in #553

> ## Security and Access Control
> * **Allow/deny lists** for IP addresses
> * **Connection gating** based on IP networks
> * **Rate limiting** for incoming connections per host

libp2p/network/connection_gate.py
sumanjeet0012: Updated the extract_ip_from_multiaddr function to enable IP address extraction from /dns4, /dns6, /dnsaddr, and /p2p-circuit multiaddr formats.

sumanjeet0012:
@pacrob Ready for review.

libp2p/network/connection_gate.py
yashksaini-coder: Thanks sumanjeet, I was unable to provide a fix, grateful for your assistance sir 👍🏼

libp2p/network/connection_pruner.py
```diff
@@ -0,0 +1,433 @@
...
+                    continue
+
+            # Check allow list (connections in allow list are never pruned)
+            if is_connection_in_allow_list(connection, self.swarm):
+                continue
+
+            to_close.append(connection)
+
+            if len(to_close) >= to_prune:
+                break
```
pacrob, line 285: I'm not seeing that we check/respect the `GRACE_PERIOD` defined in `ConnectionConfig` here when deciding whether to prune or not. It it being checked somewhere I'm missing?

libp2p/network/swarm.py
pacrob: I see that we have allow/deny lists for IP addresses, but I don't see that they are consulted in this function (`upgrade_inbound_raw_conn`), meaning peers on the `deny` list can still connect inbound.

libp2p/network/swarm.py
pacrob: The type issue here should be resolved instead of ignored. Can places that call this function handle a `None properly?

yashksaini-coder: Replaced the `return None # type: ignore[return-value]` with a proper `SwarmException`

This also required adding a specific exception handler to re-raise `SwarmException`  separately from the generic exception handler (which implements fail-open behavior).

libp2p/network/connection_pruner.py
yashksaini-coder: Add graceful check, and update helper method to properly check connection grace period.

yashksaini-coder:
@sumanjeet0012 I have answered the review comments from Paul's. From your end the PR is ready ?

sumanjeet0012:
@yashksaini-coder yes the PR is ready for review.

pyproject.toml
```diff
@@ -107,6 +107,9 @@ release = [
     "twine",
     "wheel",
 ]
+interop = [
+    "requests>=2.25.0",
+]
```
pacrob: This should be added to main dependencies instead of separately down here. I do see that it is used within interop tests, but it is also used by a couple files within `autotls`. I'm not sure why it was not added to dependencies there.

pacrob: I'll make the fix, just a sec.

pacrob:
2 concerns:

Inbound allow/deny lists are not enforced. `ConnectionGate.is_allowed` is only applied on outbound dial, so inbound connections bypass the gate entirely. That doesn't seem like it is desirable behavior. I asked about this in near line 1163 of `swarm.py`, but didn't get a clear answer. If this is desireable, tell me why. If is not desireable, you can either fix here or file an issue for it to be handled later. If I'm missing something, please let me know!

The `inbound_upgrade_timeout` config is never used anywhere, meaning a stalled security upgrade could hang indefinitely even when a timeout is set. This doesn't seem like desireable behavior. Again, fine to fix here, or file an issue to be fixed later.

yashksaini-coder:
> 2 concerns:
> 
> Inbound allow/deny lists are not enforced. `ConnectionGate.is_allowed` is only applied on outbound dial, so inbound connections bypass the gate entirely. That doesn't seem like it is desirable behavior. I asked about this in near line 1163 of `swarm.py`, but didn't get a clear answer. If this is desireable, tell me why. If is not desireable, you can either fix here or file an issue for it to be handled later. If I'm missing something, please let me know!
> 
> The `inbound_upgrade_timeout` config is never used anywhere, meaning a stalled security upgrade could hang indefinitely even when a timeout is set. This doesn't seem like desireable behavior. Again, fine to fix here, or file an issue to be fixed later.

@sumanjeet0012 I tried resolving and adding security upgrade and add test to check inbound connection gate, but failed to do so, can you help me here ?

yashksaini-coder:
@pacrob Sumanjeet has replied back to your reviewed comment,

libp2p/network/swarm.py
```diff
@@ -635,11 +1085,23 @@ async def listen(self, *multiaddrs: Multiaddr) -> bool:
             async def conn_handler(
                 read_write_closer: ReadWriteCloser, maddr: Multiaddr = maddr
             ) -> None:
+                # Enforce connection gate on inbound connections
+                if not self.connection_gate.is_allowed(maddr):
```
pacrob: `is_allowed` is an async function, but is not awaited here, so this clause will never trigger.

pacrob: Also I'm not sure it's checking the right address. Isn't `maddr` here the address the node is listening on? Let me know if I'm mistaken here.

libp2p/network/swarm.py
```diff
@@ -773,7 +1292,21 @@ async def upgrade_inbound_raw_conn(
                     peer_id, endpoint_ip=ep
                 )
                 if conn_scope is None:
-                    await secured_conn.close()
+                    # Clean up connections
```
pacrob, line 1313: Do we need to clean up `muxed_conn` too?

libp2p/network/swarm.py
sumanjeet0012: I will use `is_allowed` with `await`.

Also, the `maddr` appears to be incorrect in this case. I am not sure whether I can obtain the remote node’s multiaddress using `read_write_closer.get_remote_address()`. The method signature is:

`def get_remote_address(self) -> tuple[str, int] | None`

Based on this, it seems it may only return an IP address and port, rather than a full multiaddress. I will verify this behavior and adjust the implementation accordingly.

libp2p/network/swarm.py
sumanjeet0012: Yes, I believe so as well. I will include it.

libp2p/network/swarm.py
sumanjeet0012: Used the correct multiaddress and used await with is_allowed.

libp2p/network/swarm.py
sumanjeet0012: added muxed_conn.close().

sumanjeet0012:
@pacrob I have done the requested changes.
Ready for review.

libp2p/__init__.py
```diff
@@ -478,6 +478,7 @@ def new_host(
     negotiate_timeout: int = DEFAULT_NEGOTIATE_TIMEOUT,
     enable_quic: bool = False,
     quic_transport_opt: QUICTransportConfig | None = None,
+    connection_config: ConnectionConfig | None = None,
```
pacrob: Does this need to be here, or can it go at the end? It does seem better ordered to keep it with the other configs, but it would be a breaking change. Please add an additional `553.breaking.rst` newsfragment noting the change.

pacrob (approved):
Needs additional newsfragment for breaking function signature change. Approved for merge once that is handled.

libp2p/__init__.py
sumanjeet0012: Placed the config at the end.

sumanjeet0012:
> Needs additional newsfragment for breaking function signature change. Approved for merge once that is handled.

Added the newsfragment file.

seetadev:
@yashksaini-coder : Great, thank you for resolving the merge conflicts. Please also add some more details at the discussion page. 

@sumanjeet0012 : Appreciate your great efforts on this PR. Also, wish to thank @yashksaini-coder for his contribution. Keep up the great progress, friends :)

@pacrob: Thank you so much Paul for your detailed review and feedback. Appreciate your great support and mentorship.

Issue #1111, opened 2025-12-26
ipv6 error: MultiAddr '/ip6/::1/tcp/0' does not contain protocol ip4


TimSC:
### Summary

When I try to create peers that communicate over ipv6 I get an error using libp2p 0.5.0:

```python
    |   File "/home/tim/dev/p2ptest/env/lib/python3.12/site-packages/libp2p/transport/tcp/tcp.py", line 91, in listen
    |     ip4_host_str = maddr.value_for_protocol("ip4")
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/home/tim/dev/p2ptest/env/lib/python3.12/site-packages/multiaddr/multiaddr.py", line 293, in value_for_protocol
    |     raise exceptions.ProtocolLookupError(proto, str(self))
    | multiaddr.exceptions.ProtocolLookupError: MultiAddr '/ip6/::1/tcp/0' does not contain protocol Protocol(code=4, name='ip4', codec='ip4')
```

My minimal example:

```python
import argparse

import trio
import multiaddr

from libp2p import (
    InsecureTransport,
    PLAINTEXT_PROTOCOL_ID,
    TProtocol,
    create_new_ed25519_key_pair,
    new_host,
)
from libp2p.peer.peerinfo import info_from_p2p_addr
from libp2p.peer.peerstore import PeerStore
from libp2p.tools.async_service import background_trio_service


def make_plaintext_host(listen_addrs=None):
    key_pair = create_new_ed25519_key_pair()
    peerstore = PeerStore()
    sec_opt = {
        TProtocol(PLAINTEXT_PROTOCOL_ID): InsecureTransport(
            key_pair, peerstore=peerstore
        )
    }
    return new_host(
        key_pair=key_pair,
        sec_opt=sec_opt,
        peerstore_opt=peerstore,
        listen_addrs=listen_addrs,
    )


async def main(use_ipv4: bool):
    if use_ipv4:
        listen_addr = multiaddr.Multiaddr("/ip4/127.0.0.1/tcp/0")
    else:
        listen_addr = multiaddr.Multiaddr("/ip6/::1/tcp/0")

    server = make_plaintext_host([listen_addr])
    client = make_plaintext_host()

    async with server.run([listen_addr]):
        async with background_trio_service(client.get_network()):
            await trio.sleep(0.1)
            server_addrs = server.get_addrs()
            print("server addrs:", server_addrs)

[...]

PR #1131, opened 2026-01-13
 Add IPv6 support to TCP transport layer

bomanaps:
## What was wrong?
https://github.com/libp2p/py-libp2p/discussions/1113 https://github.com/libp2p/py-libp2p/issues/1111 

The TCP transport hardcoded IPv4 protocol lookups (maddr.value_for_protocol("ip4")), causing ProtocolLookupError when using IPv6 addresses like /ip6/::1/tcp/0. The _multiaddr_from_socket() function also always created /ip4/ multiaddrs regardless of actual socket family. 

## How was it fixed?
Created shared helper functions in libp2p/utils/multiaddr_utils.py that extract IP addresses from multiaddrs by trying IPv4 first, then IPv6. Updated TCP transport's listen(), dial(), and _multiaddr_from_socket() to use these helpers and detect socket family (AF_INET6 vs AF_INET). Also enabled IPv6 enumeration in address validation utilities and updated constants to support IPv6 via the LIBP2P_BIND environment variable.
Summary of approach.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/transport/test_tcp.py
+++ b/tests/core/transport/test_tcp.py
@@ -38,7 +38,9 @@ async def test_tcp_listener(nursery):
     assert len(listener.get_addrs()) == 2
 
 
-@pytest.mark.skip(reason="IPv6 listener hangs - trio.serve_tcp may need IPv6-specific configuration")
+@pytest.mark.skip(
+    reason="IPv6 listener hangs - trio.serve_tcp may need IPv6-specific configuration"
+)
 @pytest.mark.trio
 async def test_tcp_listener_ipv6(nursery):
     """Test TCP listener with IPv6 address."""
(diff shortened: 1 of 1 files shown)


bomanaps:
Keeping this in draft as  I ran into two pre-commit issues that appear to be pre-existing and unrelated to my IPv6 changes @acul71 @seetadev

bomanaps:
Hello @seetadev @acul71 @sumanjeet0012 please can I get a review on this pr?

tests/core/transport/test_tcp.py
```diff
@@ -38,6 +38,28 @@ async def handler(tcp_stream):
     assert len(listener.get_addrs()) == 2
 
 
+@pytest.mark.skip(
```
pacrob, line 41: What needs to be done to not need to skip these?

If we're adding tests, they should pass. If there is more work to be done that is outside the scope of this PR, a new issue should be filed prior to merge with a note to remove the skip. Otherwise they'll be forgotten.

pacrob:
@seetadev This is the PR to be closed, correct?

Issue #920, opened 2025-09-12
Add Gossipsub 2.0 Support in py-libp2p


seetadev:
Description

We need to add Gossipsub 2.0 support to py-libp2p in order to bring the Python implementation up to parity with other libp2p language implementations.

Gossipsub 2.0 provides several improvements over Gossipsub 1.1, mainly focusing on security, robustness under adversarial conditions, and adaptive message propagation. Adding support in py-libp2p will make it possible for Python-based projects to participate in modern libp2p pubsub networks without being limited by outdated protocol versions.

Why This Is Important

Security: Peer scoring and behavioral penalties mitigate spam, Sybil, and Eclipse attacks.

Reliability: Adaptive gossip dissemination ensures better message propagation in dynamic networks.

Compatibility: Projects using Python can interoperate with other libp2p implementations already on Gossipsub 2.0.

Scope

This issue tracks the initial implementation and integration of Gossipsub 2.0 into py-libp2p. It includes:

Implementing peer scoring (delivery rates, behavior penalties, decay).

Adding message validation hooks to allow application-level validation.

Supporting adaptive gossip dissemination.

Ensuring interoperability with other libp2p implementations (Go, JS).

Maintaining fallback/compatibility with Gossipsub 1.1 for transition.

References

[Gossipsub v2 Spec](https://github.com/libp2p/specs/tree/master/pubsub/gossipsub)

Gossipsub v2 Research Paper

Existing implementations for reference:

[go-libp2p-pubsub](https://github.com/libp2p/go-libp2p-pubsub)

[js-libp2p-pubsub](https://github.com/libp2p/js-libp2p/tree/main/packages/pubsub)

Motivation

Currently, py-libp2p only supports Gossipsub 1.1, which is limited in its ability to handle adversarial network conditions such as spam, Sybil attacks, and poorly behaving peers.

With the growing adoption of Gossipsub 2.0 in Go and JS implementations, the Python ecosystem risks falling behind if it cannot interoperate with modern pubsub networks. Many decentralized applications, research projects, and data-sharing systems written in Python rely on py-libp2p for networking.

Without Gossipsub 2.0 support, they are exposed to:

Security risks from malicious peers (no peer scoring or penalties in 1.1).

Unreliable message propagation in high-churn networks.

Lack of interoperability with the latest libp2p implementations.

Adding Gossipsub 2.0 to py-libp2p will:

Bring feature parity with Go and JS implementations.

[...]

PR #1139, opened 2026-01-18
eg/1130-add-gossipsub-comparison-and-standalone-examples

### LINKED ISSUE #1130: Add Gossipsub Version Comparison Demo and Standalone Examples
## Overview

Create comprehensive examples showcasing the differences between Gossipsub 1.1, 1.2, and 2.0, along with a standalone demo highlighting the new v2.0 features including peer scoring, adaptive gossip, and validation behavior.

## Background

With the recent implementation of Gossipsub 2.0 support in py-libp2p (PR #1066), we now have full protocol version support spanning:
- **Gossipsub 1.0** (`/meshsub/1.0.0`) - Basic mesh-based pubsub
- **Gossipsub 1.1** (`/meshsub/1.1.0`) - Added peer scoring and behavioral penalties
- **Gossipsub 1.2** (`/meshsub/1.2.0`) - Added IDONTWANT message filtering
- **Gossipsub 2.0** (`/meshsub/2.0.0`) - Enhanced security, adaptive gossip, and advanced peer scoring

These examples will help users understand the evolution of the protocol and demonstrate the practical benefits of upgrading to newer versions.

### 1. Multi-Version Comparison Demo 
**Objective**: Side-by-side demonstration of how different Gossipsub versions handle the same network scenarios.

**Key Features**:
- **Network Simulation**: Create identical network topologies running different protocol versions
- **Scenario Testing**: Test various network conditions (high churn, malicious peers, network partitions)
- **Performance Metrics**: Compare message delivery rates, latency, and network overhead
- **Visual Output**: Real-time comparison charts and statistics
- **Interactive Mode**: Allow users to inject different types of network events

**Scenarios to Demonstrate**:
- Normal operation with honest peers
- Network with high peer churn
- Presence of spamming/flooding peers
- Network partition and recovery
- Eclipse attack simulation
- Message validation failures

### 2. Gossipsub 2.0 Feature Showcase
**Objective**: Interactive demonstration of Gossipsub 2.0's advanced features.

**Key Components**:

#### Peer Scoring Visualization
- **Real-time Score Display**: Show peer scores (P1-P7 parameters) updating in real-time
- **Score Component Breakdown**: Visualize individual scoring components
- **Behavioral Penalties**: Demonstrate how misbehavior affects peer scores
- **IP Colocation Penalties**: Show P7 penalties for peers from same IP ranges
- **Application Scoring**: Demonstrate P6 custom application-defined scoring

#### Adaptive Gossip Demonstration
- **Network Health Monitoring**: Display network health score calculation
- **Dynamic Parameter Adjustment**: Show how gossip parameters adapt to network conditions
- **Mesh Quality Maintenance**: Visualize mesh degree adjustments
- **Opportunistic Grafting**: Demonstrate score-based peer selection

#### Security Features
- **Spam Protection**: Show rate limiting in action
- **Eclipse Attack Protection**: Demonstrate IP diversity enforcement
- **Equivocation Detection**: Show detection and penalties for duplicate messages
- **Message Validation**: Demonstrate validation hooks and caching

## Technical Architecture

```mermaid
graph TB
    subgraph "Version Comparison Demo"
        A[Network Simulator] --> B[v1.1 Network]
        A --> C[v1.2 Network] 
        A --> D[v2.0 Network]
        B --> E[Metrics Collector]
        C --> E
        D --> E
        E --> F[Comparison Dashboard]
    end
    
    subgraph "v2.0 Features Demo"
        G[Feature Controller] --> H[Peer Scoring Monitor]
        G --> I[Adaptive Gossip Monitor]
        G --> J[Security Features Monitor]
        H --> K[Real-time Visualization]
        I --> K
        J --> K
    end
 
```

## Implements from

- Gossipsub 1.1 #871
- Gossipsub 1.2 #806
- Gossipsub 2.0 #920

Winter-Soren:
## What was wrong?

Fixes #1130

py-libp2p lacked comprehensive examples demonstrating differences between Gossipsub protocol versions and showcasing advanced features (scoring, adaptive gossip, security options).

## How was it fixed?

Added a versioned example suite under `examples/pubsub/gossipsub/`:

- Shared helpers in `_common.py` (host lifecycle, mesh wiring, trio deadlines)
- Per-version demos: `gossipsub_v1.0.py` … `gossipsub_v2.0.py`
- Multi-version comparison: `compare_versions.py` (normal / spam / churn)
- Sphinx page `docs/examples.gossipsub.rst` + newsfragment `1130.docs.rst`
- Short-duration smoke tests in `tests/examples/test_gossipsub_version_examples.py`

Companion run guide: Discussion #1297.

**Notes:** `/meshsub/1.4.0` and `/meshsub/2.0.0` are experimental py-libp2p profiles (not upstream-standardized); demos state this explicitly.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


IronJam11:
Missing: 
[v1.0] 
- fanout behaviour not shown in v1.0

[v1.1] 
- ⁠p6 and p7 (IP Colocation Factor, Behavioural Penalty)
⁠- enabling prune backoff and peer exchange
⁠- ”can” add application score function (in real applications, this could be based on staking/reputation in a blockchain, role in the network (validator, full node,...)

@Winter-Soren @seetadev

Winter-Soren:
> Missing: [v1.0]
> 
> * fanout behaviour not shown in v1.0
> 
> [v1.1]
> 
> * ⁠p6 and p7 (IP Colocation Factor, Behavioural Penalty)
>   ⁠- enabling prune backoff and peer exchange
>   ⁠- ”can” add application score function (in real applications, this could be based on staking/reputation in a blockchain, role in the network (validator, full node,...)
> 
> @Winter-Soren @seetadev


Hey @IronJam11,

Thank you for flagging this!  
I have incorporated the missing features into the **Gossipsub 1.0** and **Gossipsub 1.1** examples.

1. **v1.0 demo**: Shows `node_0` as a fanout-only publisher (no subscription, 3 messages sent, 0 received), with other nodes receiving via fanout. The summary explicitly lists the fanout behavior.

2. **v1.1 demo**: Runs with **P1–P7**, prune backoff, **PX**, and an optional app score, all described in the banner and final feature list. Honest vs. malicious behavior stats are printed as expected.

Fanout (**v1.0**) and **P6/P7 + PX + prune backoff + app score** (**v1.1**) are now covered.

seetadev:
@Winter-Soren , @IronJam11 : Hi friends. Appreciate the efforts. The PR does an excellent job of making the evolution of Gossipsub (v1.0 → v2.0) tangible through well-structured, runnable examples, something that has been missing in py-libp2p for a while. The multi-version comparison, scenario-based simulations (churn, spam, partitions), and especially the interactive v2.0 showcase with peer scoring and adaptive gossip provide real educational and debugging value for both new and experienced contributors.

It’s also great to see the feedback loop handled so well, fanout behavior in v1.0 and the inclusion of P6/P7, PX, prune backoff, and optional app scoring in v1.1 really strengthen the completeness and accuracy of the examples. The attention to concurrency (trio nursery management), error handling, and metrics output makes this not just illustrative but practically usable.

I’m aligned with the approach to scope this PR to all versions except v1.3 and take that up separately to avoid blocking the release. This is in a very good state to land once the final check clears, would be great to get this merged ahead of the upcoming release.

Winter-Soren:
Added Discussion page #1297, which explains the differences among all Gossipsub versions, along with instructions on how to run these examples.

Winter-Soren:
Hi @seetadev @pacrob @acul71 @sumanjeet0012 @yashksaini-coder 
I would love to have your review and feedback on the PR. A guide on running these examples is added in #1297.

examples/pubsub/gossipsub/gossipsub_v2.0.py
```diff
@@ -0,0 +1,393 @@
...
+
+# Configure logging
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s - %(name)s - %(levelname)s - %(message)s",
+)
+logger = logging.getLogger("gossipsub-v2.0")
+
+# Protocol version
+GOSSIPSUB_V20 = TProtocol("/meshsub/2.0.0")
```
yashksaini-coder, line 50: This protocol ID `/meshsub/2.0.0` doesn't exist in the libp2p spec — the official GossipSub spec only goes up to v1.2. py-libp2p has this constant internally but the example's docstring reads like v2.0 is a standardized thing with distinct features like "equivocation detection" and "eclipse attack protection via IP diversity", which are really just v1.4 features repackaged.

This will confuse users who look at the spec and can't find v2.0 anywhere. Can you either:
- Add a clear disclaimer at the top of the docstring: "Note: GossipSub 2.0 is an experimental/internal protocol version in py-libp2p, not part of the official libp2p specification"
- Or drop this file and end the progression at v1.4

What do you think?

examples/pubsub/gossipsub/gossipsub_v1.0.py
```diff
@@ -0,0 +1,303 @@
...
+        self.port = port
+        self.fanout_only = fanout_only  # If True, node only publishes (no subscribe)
+        self.host: IHost | None = None
+        self.pubsub: Pubsub | None = None
+        self.gossipsub: GossipSub | None = None
+        self.subscription: ISubscriptionAPI | None = None
+        self.messages_sent = 0
+        self.messages_received = 0
+
+    async def start(self):
```
yashksaini-coder, line 65: This entire method body (`start`, `publish_message`, `receive_messages`, `connect_to_peer`) is copy-pasted nearly identically across all 6 version files. Same for the Demo class (`setup_network`, `start_network`, `_connect_nodes`, `_print_statistics`, `main`). I diffed v1.0 against v1.1 and the method signatures are identical — only the GossipSub constructor args change.

That's roughly 200+ lines of boilerplate repeated 6 times. If anything in the demo infra needs fixing (say, the connection logic or nursery management), you'd have to patch all 6 files.

Would you consider extracting a shared `BaseGossipsubNode` + `BaseGossipsubDemo` that each version file subclasses? Each file would then only need to override the GossipSub config and the feature checklist printout — maybe 40-60 lines per version instead of 300-550. Happy to help sketch this out if useful.

examples/pubsub/gossipsub/gossipsub_v1.0.py
```diff
@@ -0,0 +1,303 @@
...
+            # No score_params - v1.0 doesn't have peer scoring
+            # No max_idontwant_messages - v1.0 doesn't support IDONTWANT
+            # No adaptive features - v1.0 doesn't have adaptive gossip
+            # No security features - v1.0 has basic security only
+        )
+
+        self.pubsub = Pubsub(self.host, self.gossipsub)
+
+        # Start services
+        import multiaddr
```
yashksaini-coder, line 91: Minor: `import multiaddr` is inside the method body here and in `connect_to_peer()`. In the v1.3 and v1.4 examples it's imported at module level. Should be consistent — move it to the top with the other imports.

examples/pubsub/gossipsub/gossipsub_v1.0.py
```diff
@@ -0,0 +1,303 @@
...
+                print(f"\n{'=' * 60}")
+                print("GOSSIPSUB 1.0 DEMO")
+                print(f"{'=' * 60}")
+                print(f"Running for {duration} seconds...")
+                print("Protocol: /meshsub/1.0.0")
+                print("Features: Basic mesh-based pubsub, simple flooding, fanout demo")
+                print("  (node_0 is fanout-only: publishes via fanout, not in mesh)")
+                print(f"{'=' * 60}\n")
+
+                while time.time() < end_time:
```
yashksaini-coder, line 206: `time.time()` in a trio async context can drift since trio has its own clock. Consider using `trio.current_time()` with relative deadline tracking, or just count iterations — it's more idiomatic for trio code.

examples/pubsub/gossipsub/gossipsub_v1.1.py
```diff
@@ -0,0 +1,361 @@
...
+                message_counter = 0
+
+                print(f"\n{'=' * 60}")
+                print("GOSSIPSUB 1.1 DEMO")
+                print(f"{'=' * 60}")
+                print(f"Running for {duration} seconds...")
+                print("Protocol: /meshsub/1.1.0")
+                print(
+                    "Features: Peer scoring (P1-P7), prune backoff, peer exchange (PX),"
+                    "optional app score"
```
yashksaini-coder, line 237: Looks like there's a missing space before "optional app score" — the string concatenation drops it:
```python
"Features: Peer scoring (P1-P7), prune backoff, peer exchange (PX),"
"optional app score"
```
Should be `"...PX), "` with a trailing space, or `", optional app score"`.

tests/core/pubsub/test_gossipsub.py
```diff
@@ -724,7 +724,9 @@ async def test_sparse_connect_degree_zero():
         degree = 0
 
         await sparse_connect(hosts, degree)
-        await trio.sleep(0.1)  # Allow connections to establish
+        # Match test_sparse_connect / test_dense_connect_fallback: pubsub streams need
```
yashksaini-coder, line 730: This test fix (sleep 0.1s → 2s for CI stability) makes sense but it's unrelated to the gossipsub examples. Would be cleaner as its own 1-line PR so it can land independently without waiting for the examples review to finish. Up to you though, not blocking on this.

yashksaini-coder (changes requested):
Thanks for putting this together @Winter-Soren — the educational progression from v1.0 through v2.0 is genuinely useful and this has been missing from py-libp2p for a while. The per-version feature checklists and the companion discussion (#1297) are nice touches.

That said, I've flagged a few things that need attention before this can be merged. The big ones are: missing newsfragment (project policy blocker), heavy code duplication across the 6 example files, and the v2.0 example implying a standardized protocol that doesn't exist in the spec. Details inline below.

Also the commit history still needs squashing (21 commits with 9 merge commits) — the PR TODO has this unchecked too.

examples/pubsub/gossipsub/gossipsub_v1.0.py
Winter-Soren: Thank you @yashksaini-coder for flagging this, I have moved the import to the top in this commit: [063efa3](https://github.com/libp2p/py-libp2p/pull/1139/commits/063efa397ec027296175cc886fdf249bf7132807)

examples/pubsub/gossipsub/gossipsub_v1.0.py
Winter-Soren: Thanks @yashksaini-coder, you’re right that there’s a lot of structural overlap across the six scripts, and a small shared module would make infra fixes easier to land in one place.

For this PR, I am intentionally keeping each example self-contained in a single file so newcomers can read or copy one script without following imports across the folder. That’s a deliberate tradeoff: we accept more parallel edits if these demos stay mostly static.

examples/pubsub/gossipsub/gossipsub_v2.0.py
Winter-Soren: Thank you @yashksaini-coder for identifying this, you are absolutely correct that Gossipsub 2.0 specification is absent from the official libp2p specification. You can find details of the Gossipsub 2.0 here: #920

examples/pubsub/gossipsub/gossipsub_v1.0.py
Winter-Soren: Thanks for the note @yashksaini-coder,  you’re right that wall-clock `time.time()` doesn’t line up with trio’s scheduling clock. I’ve updated the demo loop to use `trio.current_time()` with a relative deadline `(deadline = trio.current_time() + duration)`. I dropped import time and simplified the published payload to `msg_{counter}` since the counter is enough to identify messages in this run. You can find it here [419b90e](https://github.com/libp2p/py-libp2p/pull/1139/commits/419b90e92fbb7a6801e6baae35dc14102e368071)

examples/pubsub/gossipsub/gossipsub_v1.1.py
Winter-Soren: Fixed by adding a trailing space before `optional` as addition of space after (PX), resulting in the failing pre-commit hook, so the line prints correctly.

tests/core/pubsub/test_gossipsub.py
Winter-Soren: Right! to keep that change separate from the examples work. I reverted `test_sparse_connect_degree_zero` to the shorter `trio.sleep(0.1)` here so this PR stays scoped; if the longer settle time is still needed for CI under xdist, I’ll open a tiny follow-up PR just for that.

acul71:
Please address remaining review feedback (newsfragment, dedupe, squash) so we can merge. CI is green and branch is CLEAN.

acul71:
Addressed remaining review feedback and squashed onto current `main` (single commit `d8209bfe`):

- Extracted shared helpers (`examples/pubsub/gossipsub/_common.py`) and slimmed the six version scripts
- Explicitly disable adaptive gossip / spam / eclipse defaults on older demos
- Added experimental disclaimers for `/meshsub/1.4.0` and `/meshsub/2.0.0`
- Restored multi-version comparison (`compare_versions.py`) for issue #1130
- Real P6 app scores + score snapshots; v2.0 topic validator rejects `invalid_*` payloads
- Added `newsfragments/1130.docs.rst`, Sphinx page `docs/examples.gossipsub.rst`, and short smoke tests
- PR description now uses `Fixes #1130`; commit history cleaned up

@yashksaini-coder please re-review when you have a chance.

yashksaini-coder (changes requested):
Re-reviewed after the rework — thanks @Winter-Soren, this is close. My earlier blockers are all resolved: newsfragment added, the six scripts now share `_common.py` (nice extraction), v2.0 carries clear "experimental py-libp2p profile" disclaimers, and the history is squashed. ruff/format are clean and the example smoke tests pass locally (7/7).

**Requesting one change before approval — the `churn` scenario does not churn.**

`compare_versions.py` advertises it as:
> churn — disconnect/reconnect a subset mid-run

but the implementation (`compare_versions.py:185-190`) only logs and `await trio.sleep(1.0)` once — no node is paused, disconnected, or reconnected, and every honest node keeps publishing throughout. An educational example shouldn't teach a scenario it does not actually run. Please either:

- implement a minimal real churn (e.g. cancel a couple of nodes' tasks mid-run and restart them, or drop and re-establish their connections) so the metric reflects churn, **or**
- if real churn is out of scope for now, drop the `churn` choice / relabel it honestly (e.g. "publish pause") so the docstring matches the behavior.

**Minor (non-blocking):** `delivery_ratio` is total-received / total-sent across N subscribers, so it can exceed 1.0 (~N−1 receives per message). Fine as a relative metric, but a one-line note in the table header would save readers a double-take.

For the maintainers: the red **`tox (3.13, demos)`** check is unrelated to this PR — it fails on `test_dht_chat.py::test_dht_chat_find_peer_then_message` (from #1506), which passes locally and on every other matrix leg; it is flaky under parallel load on Linux 3.13. This PR's own gossipsub tests pass.

Winter-Soren:
Thanks @acul71 for handling the earlier feedback and cleanup. I’ve built on your rework and addressed the remaining churn and delivery-ratio feedback in 996fe173.

Winter-Soren:
Thanks @yashksaini-coder. Fixed in 996fe173: the churn scenario now disconnects half the nodes while the others continue publishing, then reconnects them and resumes their publishing. The regression test checks live connections before, during, and after churn. I also clarified in the table and docs that delivery_ratio aggregates receives across subscribers and can exceed 1.0. All eight targeted tests and pre-commit checks pass locally. Please take another look.

yashksaini-coder (approved):
Re-reviewed — thanks @Winter-Soren, this addresses everything from my earlier pass. The `churn` scenario now genuinely disconnects half the nodes mid-run and reconnects, and the new `test_compare_versions_churn_disconnects_and_reconnects` proves it end-to-end (disconnect → offline nodes stop publishing → reconnect). The `delivery_ratio may exceed 1.0` note is in as well. CI is green and the example smoke tests pass. LGTM 👍

One **non-blocking** follow-up: the churn test leaks sockets on teardown. Measured with `pytest <test> -W all::ResourceWarning -p no:cacheprovider -o addopts=""`: the `normal` scenario reports **0** unclosed sockets, `churn` reports **4** — the disconnect/reconnect path leaves established sockets open. It passes today only because the leak surfaces as an *unraisable* `ResourceWarning` (pytest converts it to `PytestUnraisableExceptionWarning`, which the `error::ResourceWarning` guard from #1533 does not escalate). That is the same fragility as #1541 and can become a hard failure on `windows (3.12, demos)`. Worth closing the churn teardown (or confirming whether it is a `host.disconnect()`/reconnect leak at the library level) in a follow-up — not a blocker for this PR.

Winter-Soren:
Thanks @yashksaini-coder — reproduced the four leaked sockets and fixed the teardown in 154fdaf9. The original connections closed during `host.disconnect()`; final nursery cancellation was interrupting cleanup of remaining connections. `DemoNode.start()` now closes its host in a shielded `finally` block.

The churn regression test now checks that every TCP socket is closed when the run returns. All eight example tests pass locally with both `ResourceWarning` and `pytest.PytestUnraisableExceptionWarning` treated as errors. An in-process normal-scenario check also leaves zero sockets open. Pre-commit passes; CI is running.

PR #1153, opened 2026-01-24
fix(swarm): prevent listener crash on inbound upgrade failure

### LINKED ISSUE #417: Failure to negotiate a peer should not take down the listener
It seems that an exception in this function:

https://github.com/libp2p/py-libp2p/blob/master/libp2p/network/swarm.py#L234

results in the swarm (?) throwing an exception which takes down the entire listener.

We should catch failures in peer negotiation appropriately so this doesn't happen.

imApoorva36:
This PR fixes an issue where failures during inbound peer negotiation (e.g. security or mux upgrade failures) could propagate exceptions that terminate the swarm listener. The listener now logs and closes failed connections without crashing, allowing it to continue accepting new peers.

Code changes:
- Catch and handle inbound upgrade failures inside `conn_handler`
- Ensure failed connections are closed without propagating exceptions
- Add defensive handling around `add_conn`
- Add a regression test verifying listener resilience after upgrade failures

Added Testing:
- Added `test_swarm_listener_resilience_on_upgrade_failure`
- Verified that a failed inbound negotiation does not crash the listener
- Verified that subsequent connection attempts succeed

Closes #417


seetadev:
@imApoorva36 : Thanks for the detailed fix and the clear write-up, Apoorva — this is a solid improvement 👍

Preventing listener crashes on inbound upgrade failures is exactly the right behavior, and the added regression test gives good confidence that the swarm remains resilient after negotiation failures. Nicely scoped changes overall.

Before we merge, could you please:

Take a look at the failing CI checks (tox failures on a few Python versions) and resolve those, and

Add a newsfragment describing this fix, since it’s a user-visible stability improvement for the swarm listener.

Once those are addressed,  will ask @acul71, @lla-dane and @sumanjeet0012 for their peer reviews. Thanks again for the thoughtful work and test coverage here 🙌

imApoorva36:
Hey @seetadev, I looked into the CI failures and it seems the jobs are failing during environment setup, before tests run itself

The logs show uv trying to fetch pip==25.3, which doesn’t appear to exist on PyPI, resulting in a 404. This looks like an infra or dependency resolution issue...

Would you suggest re-running the workflows once, or should we address this separately (e.g. pinning pip to a valid version) in a follow-up PR?

imApoorva36:
Hey @sumanjeet0012 , I ran the `make pr` command here as well and no issues showed up...

Issue #1118, opened 2026-01-03
Async topic validator crashes gossipsub service


hayotensor:
### Summary

I just recently received this issue that shuts down the pubsub service when using an async topic validator:

When using a sync topic validator with the same logic, this issue doesn't occur

### Expected behavior

MessageCache to successfully pop the topic history

### Actual behavior

The key is attempting to be popped but it apparently doesn't exist

### Relevant log output

```shell
2026-01-03 16:16:27,482 [ERROR] [async_service.Manager] <Manager[GossipSub] flags=SRcfe>: task heartbeat[daemon=True] exited with error: (b'\x00\x00\x00\x00iY\x86\xff', b'\x00$\x08\x01\x12 /X\x99\xbbp\xa2\xaa9\x93\xc2\xd6f\xa2\x99\x1e[vH"rac\x8e\x1e!\x8b\xff\x91\x0c)\xe7x')
Traceback (most recent call last):
  File "/home/bob/py-libp2p-subnet-gossip/.venv/lib/python3.10/site-packages/libp2p/tools/async_service/base.py", line 334, in _run_and_manage_task
    await task.run()
  File "/home/bob/py-libp2p-subnet-gossip/.venv/lib/python3.10/site-packages/libp2p/tools/async_service/trio_service.py", line 118, in run
    await self._async_fn(*self._async_fn_args)
  File "/home/bob/py-libp2p-subnet-gossip/.venv/lib/python3.10/site-packages/libp2p/pubsub/gossipsub.py", line 565, in heartbeat
    self.mcache.shift()
  File "/home/bob/py-libp2p-subnet-gossip/.venv/lib/python3.10/site-packages/libp2p/pubsub/mcache.py", line 102, in shift
    self.msgs.pop(entry.mid)
KeyError: (b'\x00\x00\x00\x00iY\x86\xff', b'\x00$\x08\x01\x12 /X\x99\xbbp\xa2\xaa9\x93\xc2\xd6f\xa2\x99\x1e[vH"rac\x8e\x1e!\x8b\xff\x91\x0c)\xe7x')
2026-01-03 16:16:27,496 [ERROR] [__main__] Fatal error: Exceptions from Trio nursery (1 sub-exception)
  + Exception Group Traceback (most recent call last):
  |   File "/home/bob/py-libp2p-subnet-gossip/subnet/cli/run_node.py", line 408, in main
  |     trio.run(server.run)
[...]

PR #1156, opened 2026-01-25
[FIX] Async topic validator issue

IronJam11:
## What was wrong?

Issue #1118

When using async topic validators in GossipSub, the pubsub service crashes with `KeyError` in `mcache.shift()`.

**Root cause:** Race condition in `MessageCache.put()`

1. Two async validators process the same message concurrently
2. Both call `mcache.put(msg)` after validation completes
3. `put()` has two operations:
   - `self.msgs[mid] = msg` - overwrites if exists (dict count stays 1)
   - `self.history[0].append(...)` - always appends (list count becomes 2)
4. Result: 1 entry in `msgs`, but 2 entries in `history` for the same message ID
5. When `shift()` rotates the history window, it tries to `pop()` the same mid twice
6. Second pop raises `KeyError` because message was already removed

## How was it fixed?

**1. Duplicate detection in `put()`**

Added early return if message ID already exists:

```python
def put(self, msg: rpc_pb2.Message) -> None:
    mid: tuple[bytes, bytes] = (msg.seqno, msg.from_id)
    if mid in self.msgs:
        return  # Already cached, don't create duplicate history entry
    self.msgs[mid] = msg
    self.history[0].append(CacheEntry(mid, msg.topicIDs))
```

**2. Defensive pop in `shift()`**

Changed from `pop(mid)` to `pop(mid, None)` for additional safety:

```python
self.msgs.pop(entry.mid, None)  # Won't raise KeyError if missing
```

**Why this works in Trio:** Since `put()` is synchronous (no `await` points), each call runs atomically. First call adds to both `msgs` and `history`, second call finds `mid in self.msgs` and returns early. No duplicate history entries can be created.

**Files changed:**
- `libp2p/pubsub/mcache.py` - Added duplicate check + defensive pop
- `tests/core/pubsub/test_mcache_race_condition.py` - Tests that fail before fix, pass after

**Impact:**
- ✅ Fixes production crashes with async validators
- ✅ Minimal performance impact (single dict lookup per `put()`)
- ✅ Backwards compatible (no API changes)
- ✅ No side effects (messages cached exactly once)

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

capybara

Before fix: 
Screenshot 2026-01-25 at 4 04 43 PM

After fix: 
Screenshot 2026-01-25 at 4 06 07 PM

### DIFF
--- a/docs/libp2p.utils.rst
+++ b/docs/libp2p.utils.rst
@@ -12,6 +12,14 @@ libp2p.utils.logging module
    :undoc-members:
    :show-inheritance:
 
+libp2p.utils.multiaddr_utils module
+-------------------------------------
+
+.. automodule:: libp2p.utils.multiaddr_utils
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
 libp2p.utils.varint module
 --------------------------
 
--- a/libp2p/bitswap/cid.py
+++ b/libp2p/bitswap/cid.py
@@ -7,7 +7,7 @@ Note: This is a simplified implementation for demonstration. In production,
 use a proper CID library like py-cid or multiformats.
 """
 
-import hashlib
+import multihash
 
 # Simplified CID version constants
 CID_V0 = 0
@@ -23,7 +23,7 @@ HASH_SHA256 = 0x12
 
 def compute_cid_v0(data: bytes) -> bytes:
     """
-    Compute a CIDv0 for data (simplified version).
+    Compute a CIDv0 for data using py-multihash v3 API.
 
     CIDv0 is just a base58-encoded multihash (SHA-256).
     For simplicity, we return the raw multihash bytes.
@@ -35,19 +35,13 @@ def compute_cid_v0(data: bytes) -> bytes:
         CIDv0 as bytes (multihash format)
 
     """
-    # Compute SHA-256 hash
-    digest = hashlib.sha256(data).digest()
-
-    # Multihash format: <hash-type><hash-length><hash-digest>
-    # 0x12 = SHA-256, 0x20 = 32 bytes
-    multihash = bytes([HASH_SHA256, len(digest)]) + digest
-
-    return multihash
+    mh = multihash.digest(data, multihash.Func.sha2_256)
+    return mh.encode()
 
 
 def compute_cid_v1(data: bytes, codec: int = CODEC_RAW) -> bytes:
     """
-    Compute a CIDv1 for data (simplified version).
+    Compute a CIDv1 for data using py-multihash v3 API.
 
     CIDv1 format: <version><codec><multihash>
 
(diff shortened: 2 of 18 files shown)


IronJam11:
Also please note: they might be pyrefly errors that are pre-existing in the repo

tests/core/pubsub/test_mcache_race_condition.py
```diff
@@ -0,0 +1,53 @@
...
+    msg = make_msg(b"\x01", b"peer1")
+
+    mcache.put(msg)
+    mcache.put(msg)
+
+    # Shift through entire history - must not raise KeyError
+    for _ in range(mcache.history_size):
+        mcache.shift()
+
+    assert len(mcache.msgs) == 0
```
yashksaini-coder, line 53: What will happen in a scenario when the same message arrives twice with different topics attached, which topics should the cache remember ? 

Your changes did not cover this scenario, I suggest adding this test case, without this test, the behavior is undefined. Someone could accidentally change the code to merge topics, overwrite with the latest, or do something else entirely. This test locks in the rule: first message wins, duplicates are ignored.

This matters because GossipSub uses these cached topics to decide which peers get notified about messages. If the topic list changed unpredictably during races, peers might miss messages or get inconsistent views of what's happening on the network.

It's a small test, but it prevents future headaches by making the "first wins" policy explicit and protected against accidental changes.

```python
def test_duplicate_put_keeps_first_topics():
    """When same mid is put with different topicIDs, first one wins."""
    mcache = MessageCache(window_size=3, history_size=5)

    # Create two messages with the same mid but different topics
    msg_first = rpc_pb2.Message(from_id=b"peer1", seqno=b"\x01", topicIDs=["topic-A"])
    msg_second = rpc_pb2.Message(from_id=b"peer1", seqno=b"\x01", topicIDs=["topic-A", "topic-B"])  # duplicate mid

    mcache.put(msg_first)
    mcache.put(msg_second)  # This will be ignored due to duplicate mid

    mid = (b"\x01", b"peer1")

    # First message's topics are preserved
    window_a = mcache.window("topic-A")
    assert mid in window_a

    window_b = mcache.window("topic-B")
    assert mid not in window_b  # Not present from the second (ignored) put
```

Image

seetadev:
@acul71 , @sumanjeet0012, @Winter-Soren : CI/CD failure in this issue is not related to this PR. Please visit the logs: =========================== short test summary info ============================
FAILED tests/core/pubsub/test_dummyaccount_demo.py::test_set_then_send_from_five_diff_nodes_five_nodes_ring_topography
====== 1 failed, 1828 passed, 11 skipped, 3 warnings in 618.21s (0:10:18) ======
py310-core: exit 1 (620.13 seconds) /home/runner/work/py-libp2p/py-libp2p> pytest -n auto --timeout=1200 tests/core pid=2640
  py310-core: FAIL code 1 (642.65=setup[19.32]+cmd[0.10,3.10,620.13] seconds)
  evaluation failed :( (644.47 seconds)
Error: Process completed with exit code 1

Wish if you could open a new PR for arriving at a good conclusion on this issue.

yashksaini-coder:
> @acul71 , @sumanjeet0012, @Winter-Soren : CI/CD failure in this issue is not related to this PR. Please visit the logs: =========================== short test summary info ============================
> FAILED tests/core/pubsub/test_dummyaccount_demo.py::test_set_then_send_from_five_diff_nodes_five_nodes_ring_topography
> ====== 1 failed, 1828 passed, 11 skipped, 3 warnings in 618.21s (0:10:18) ======
> py310-core: exit 1 (620.13 seconds) /home/runner/work/py-libp2p/py-libp2p> pytest -n auto --timeout=1200 tests/core pid=2640
>   py310-core: FAIL code 1 (642.65=setup[19.32]+cmd[0.10,3.10,620.13] seconds)
>   evaluation failed :( (644.47 seconds)
> Error: Process completed with exit code 1
> 
> Wish if you could open a new PR for arriving at a good conclusion on this issue.

https://github.com/libp2p/py-libp2p/issues/1164

Facing similar issue in one of my PRs, whoever is working on providing the fix please also verify the ci failure address in the issue #1164

seetadev:
@pacrob , @acul71: This looks ready for final review + merge too. Wish to have your feedback and thoughts.

PR #1160, opened 2026-01-25
Fix: Make TLS Extensions Non-Critical for Rust Interop

yashksaini-coder:
This PR changes the `critical` flag for `BasicConstraints` and `KeyUsage` X.509 extensions from `True` to `False` in `libp2p/security/tls/certificate.py`. This change is needed for TLS certificate interoperability with Rust libp2p nodes, which (correctly per spec) abort if they encounter critical extensions they do not recognize.

## Context

- See [libp2p/specs/tls.md](https://github.com/libp2p/specs/blob/master/tls/tls.md): Non-critical (critical=False) is required for unrecognized or optional extensions.
- Previously both extensions were marked `critical=True`. This caused Rust libp2p (`rust-libp2p-tls`) to abort TLS connections from Python, breaking interoperability.
- The fix is to set `critical=False` on both, so Rust (and any other implementations) will ignore rather than reject the extensions.

## Changes

- **libp2p/security/tls/certificate.py**
  - Set `critical=False` for `BasicConstraints`
  - Set `critical=False` for `KeyUsage`
- No changes to other extensions or WSS self-signed certificate routines.

## Rationale

- Aligns with the TLS interoperability spec (see quote below).
- Restores working cross-language communication (Python <-> Rust).
- Harmless for Python-only operation: all Python checks still apply.

> Implementations MUST ignore non-critical extensions with unknown OIDs. Endpoints MUST abort the connection attempt if the certificate contains critical extensions that the endpoint does not understand.

---

@acul71 @seetadev The fix is in place and should resolve the Rust interop failures.

### DIFF
--- a/interop/transport/ping_test.py
+++ b/interop/transport/ping_test.py
@@ -47,7 +47,6 @@ from libp2p.security.noise.transport import (
 )
 from libp2p.security.tls.transport import (
     PROTOCOL_ID as TLS_PROTOCOL_ID,
-    IdentityConfig,
     TLSTransport,
 )
 from libp2p.utils.address_validation import get_available_interfaces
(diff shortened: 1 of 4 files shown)


acul71:
Thanks @yashksaini-coder , I've chore some comments, added a test to enforce the flags, and changed `ping-demo.py`
Merging

PR #1179, opened 2026-01-31
feat: multibase integration

asmit27rai:
## Multibase Integration

Integration of [py-multibase](https://github.com/multiformats/py-multibase) into py-libp2p as discussed here https://github.com/libp2p/py-libp2p/discussions/1173.
Fix : #1209 

---

#### **Phase 1: Add Multibase Support (Completed)**
- Add `py-multibase` as a dependency.
- Extend Peer ID class with multibase encode/decode methods (backward compatible).
- Add multibase-aware utility functions for DHT keys.
- Update pubsub message ID encoding to use multibase.
- Preserve legacy base58/base64 support for compatibility.
- Add and update tests for multibase, compatibility, and error handling.

#### **Phase 2: Internal Migration (Completed)**
- Migrate internal code to use multibase by default.
- Update peer ID parsing, DHT, and pubsub internals.
- Update examples and documentation.

#### **Phase 3: External API & Config (Completed)**
- Consider making multibase the default for external APIs.
- Provide configuration for encoding preferences.

---

### Testing

- All new and existing tests pass for Phase 3.

### DIFF
--- a/tests/core/peer/test_encoding_config.py
+++ b/tests/core/peer/test_encoding_config.py
@@ -2,9 +2,9 @@
 
 import hashlib
 
+import pytest
 import base58 as b58
 import multibase
-import pytest
 
 from libp2p.encoding_config import (
     encoding_override,


seetadev:
@asmit27rai : Thanks for the detailed breakdown and the solid work on the multibase integration — the phased approach and backward compatibility are especially well thought out 👏

Inviting @lla-dane, @yashksaini-coder, @sumanjeet0012, @acul71 to review the changes, particularly around the internal migration, API defaults, and config surface.

Noting that a few CI/CD checks are currently failing (docs, lint across multiple Python versions, and Windows 3.13 core). Could you please take a pass at fixing these so we can get this into a mergeable state? Happy to help debug if needed.

Overall, this is a strong contribution and a meaningful step forward for py-libp2p. Looking forward to the reviews and CI being green ✅

yashksaini-coder:
# Multibase Integration (PR #1179)

## Issues that I Found

### Critical Blockers
- **Missing newsfragment file in `newsfragments/`.

### Major Issues

1. **Inconsistent Return Type for `multibase.decode()`:** In some versions, this returns `bytes`, in others `(encoding, payload)`. Handling exists, but not explained.
   
2. **Potential Race in `encoding_override`:** In the `encoding_config.py`, context manager, The global lock is released after context init, not held for whole override duration. Docstring/doc comment should clarify race condition possibility.  

3. **Unclear Error in `ID.from_string()`:**
   - In the main `id.py`, method `ID.from_string`,  When decoding fails, users cannot tell if it was invalid base58 or multibase without better error messages.

---

## Security Risks & Risk Mitigations

- **Overall Security Risk:** LOW  
- **Input Validation:** Uses py-multibase, validates encodings, and falls back to base58 for compatibility.
- **Thread Safety:** Protected by a lock, but `encoding_override()` not thread-safe under concurrency.

asmit27rai:
@yashksaini-coder I added newsfragments and explainations.

libp2p/kad_dht/utils.py
```diff
@@ -128,18 +130,46 @@ def xor_distance(key1: bytes, key2: bytes) -> int:
...
+    result = multibase.decode(multibase_str)
+    # py-multibase may return bytes or (encoding, bytes) depending on
+    # version — handle both.
+    return result[1] if isinstance(result, tuple) else result
+
+
+# Keep old function for backward compatibility
+def bytes_to_base58(data: bytes) -> str:
+    """Deprecated: Use bytes_to_multibase instead."""
+    return bytes_to_multibase(data, "base58btc")
```
pacrob: Thanks for adding this for backwards compatibility! It does change the output, adding the `z` prefix. For example:

```python
# before
>>> bytes_to_base58(b"hello")
'Cn8eVZg`
# after
>>> bytes_to_base58(b"hello")
'zCn8eVZg`
```

pacrob: Consider adding a test to validate this behavior too.

libp2p/peer/id.py
```diff
@@ -58,6 +60,104 @@ def to_bytes(self) -> bytes:
...
+            return cls(peer_id_bytes)
+        except (multibase.InvalidMultibaseStringError, multibase.DecodingError):
+            raise
+        except Exception as e:
+            raise multibase.DecodingError(
+                f"Failed to decode multibase peer ID: {e}"
+            ) from e
+
+    @classmethod
+    def from_string(cls, peer_id_str: str) -> "ID":
```
pacrob, line 117: If the `peer_id_str` is base58-encoded but also validly decodes as multibase, we will preferentially return the multibase decoding. Is that the desired behavior?

An example is "f123abc". `from_base58` returns `"f123abc"` but `from_multibase` returns `7891`.

asmit27rai: Updated `from_string` to match `go-libp2p's` `peer.Decode()` logic - it no longer blindly prefers multibase. Strings starting with Qm or 1 are always decoded as base58 (legacy SHA-256 and identity multihash peer IDs). Everything else tries multibase first, but the result is only accepted if it passes `multihash.decode()` validation; otherwise we fall back to base58.

asmit27rai:
@pacrob Addressed you comments

libp2p/pubsub/pubsub.py
```diff
@@ -96,8 +95,23 @@ def get_peer_and_seqno_msg_id(msg: rpc_pb2.Message) -> bytes:
...
+    :param encoding: Encoding to use. When *None* the process-wide default
+        from :mod:`libp2p.encoding_config` is used.
+    :return: Multibase-encoded message ID
+    """
+    from libp2p.encoding_config import get_default_encoding
+
+    if encoding is None:
+        encoding = get_default_encoding()
+    digest = hashlib.sha256(msg.data).digest()
+    return multibase.encode(encoding, digest)
```
pacrob, line 114: Adding a breakpoint within this function shows that it is not getting hit during testing. Please add tests to validate new functionality and prevent regression in the future.

pacrob:
One request for testing, otherwise looking good!

asmit27rai:
@pacrob I added the test you can look into it.
Thanks.

seetadev:
@pacrob : This looks ready for final review + merge. 

CCing @acul71. We can go ahead with the merge.

tests/core/peer/test_encoding_config.py
```diff
@@ -0,0 +1,209 @@
...
+
+        msg = rpc_pb2.Message(data=b"hello world")
+        result = get_content_addressed_msg_id(msg)
+        expected_digest = hashlib.sha256(b"hello world").digest()
+        expected = multibase.encode("base58btc", expected_digest)
+        assert result == expected
+
+    def test_uses_default_encoding(self) -> None:
+        from libp2p.pubsub.pb import rpc_pb2
+        from libp2p.pubsub.pubsub import get_content_addressed_msg_id
```
pacrob: Why are you reimporting rpc_pb2 and get_content_addressed_msg_id with every test? These can be safetly hoisted to the top of the file. Or am I missing something?

seetadev: @asmit27rai : Indeed, please make the requisite correction. Thanks @pacrob for the feedback.

asmit27rai: No you are not missing just my mistake while testing each fn.

asmit27rai: taking all at modular level.

seetadev:
@pacrob : Great, thank you so much for sharing feedback. We will land this PR in today's release.

asmit27rai:
@seetadev All CI is passing.

PR #893, opened 2025-09-02
fix: GossipSub peer propagation to include FloodSub peers

paschal533:
## What was wrong?

When a node publishes a message immediately after connecting to a peer, py-libp2p sometimes crashes with:

```
KeyError: <PeerID>
```

This happens in pubsub/gossipsub.py, where `self.peer_protocol[peer_id`] is accessed before the identify handshake has finished. The handshake is asynchronous, so peer_protocol may not yet contain the connected peer.

Issue #887

## How was it fixed?

- Guarded against missing entries in `peer_protocol`
- Optionally, queue publishes until identify completes.

Summary of approach.

### To-Do

- [x] Clean up commit history
- [x] Add or update documentation related to these changes
- [x] Add entry to the [release notes](https://github.com/libp2p/py-libp2p/blob/main/newsfragments/README.md)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #1201, opened 2026-02-10
fix(gossipsub): queue early publishes until identify completes

### LINKED ISSUE #887: KeyError in GossipSub when publishing before identify handshake completes
When a node publishes a message immediately after connecting to a peer, py-libp2p sometimes crashes with:

```

KeyError: <PeerID>

```
This happens in pubsub/gossipsub.py, where `self.peer_protocol[peer_id]` is accessed before the identify handshake has finished. The handshake is asynchronous, so peer_protocol may not yet contain the connected peer.

imApoorva36:
GossipSub could drop messages published immediately after connecting to a
peer, before the identify handshake completed. While #893 prevented a
crash, the behavior was still timing-dependent.

This PR makes publishing identify-aware by queueing messages for peers
whose protocol negotiation is still in progress and flushing them once
the peer stream is registered.

Includes regression tests for early publish scenarios.

Closes #887

### DIFF
--- a/libp2p/kad_dht/kad_dht.py
+++ b/libp2p/kad_dht/kad_dht.py
@@ -133,6 +133,7 @@ class KadDHT(Service):
         protocol_prefix: TProtocol = PROTOCOL_PREFIX,
         enable_providers: bool = True,
         enable_values: bool = True,
+        strict_validation: bool = False,
     ):
         """
         Initialize a new Kademlia DHT node.
@@ -140,6 +141,36 @@ class KadDHT(Service):
         :param host: The libp2p host.
         :param mode: The mode of host (Client or Server) - must be DHTMode enum
         :param enable_random_walk: Whether to enable automatic random walk
+        :param validator: Custom NamespacedValidator for DHT records
+        :param validator_changed: If True, indicates the validator was explicitly set
+            and defaults should not be used
+        :param protocol_prefix: Protocol prefix (default: /ipfs)
+        :param enable_providers: Enable provider record support
+        :param enable_values: Enable value record support
+        :param strict_validation: If True, enforce strict namespace validation for all
+            records. Only namespaced keys (e.g., /pk/, /ipns/, /myapp/) with registered
+            validators will be accepted. If False (default), non-namespaced keys will
+            be accepted without validation for backward compatibility.
+
+            Setting this to True aligns behavior with go-libp2p and rust-libp2p where:
+            - All DHT records MUST have a registered validator for their namespace
+            - Keys without a matching namespace validator are rejected
+            - This enforces permissioned keyspaces for security and correctness
+
+        Example with strict validation:
+            # Create validator with custom namespace
+            validator = NamespacedValidator({
+                "pk": PublicKeyValidator(),
+                "myapp": MyAppValidator(),
+            }, strict_validation=True)
+            dht = KadDHT(
+                host, DHTMode.SERVER, validator=validator, strict_validation=True
+            )
+
+            # Only namespaced keys are allowed:
+            await dht.put_value("/myapp/key", b"value")  # OK
+            await dht.put_value("/pk/...", pubkey)       # OK
+            await dht.put_value("plain-key", b"value")   # Raises InvalidRecordType
         """
         super().__init__()
 
@@ -159,10 +190,17 @@ class KadDHT(Service):
         self.protocol_prefix = protocol_prefix
         self.enable_providers = enable_providers
         self.enable_values = enable_values
+        self._strict_validation = strict_validation
         self.validator = validator
 
-        if validator is None:
-            self.validator = NamespacedValidator({"pk": PublicKeyValidator()})
+        if self.validator is None:
+            self.validator = NamespacedValidator(
+                {"pk": PublicKeyValidator()},
+                strict_validation=strict_validation,
+            )
+
+        # Keep strict_validation synchronized with the active validator.
+        self.strict_validation = strict_validation
 
         # If true implies that the validator has been changed and that
         # Defaults should not be used
@@ -193,6 +231,27 @@ class KadDHT(Service):
 
         # Set protocol handlers
         host.set_stream_handler(PROTOCOL_ID, self.handle_stream)
+
+    @property
+    def strict_validation(self) -> bool:
+        """
+        Return strict validation mode.
+
+        The validator is the source of truth when it supports
+        ``strict_validation`` at runtime.
+        """
+        validator = self.validator
+        if isinstance(validator, NamespacedValidator):
+            return validator.strict_validation
+        return self._strict_validation
+
+    @strict_validation.setter
+    def strict_validation(self, value: bool) -> None:
+        """Set strict validation mode and synchronize with validator."""
+        self._strict_validation = value
+        validator = self.validator
+        if isinstance(validator, NamespacedValidator):
+            validator.strict_validation = value
 
     def _create_query_function(self) -> Callable[[bytes], Awaitable[list[ID]]]:
         """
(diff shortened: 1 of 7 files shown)


imApoorva36:
@seetadev @sumanjeet0012 this PR is now ready for review, would love your thoughts

libp2p/pubsub/gossipsub.py
```diff
@@ -416,6 +546,56 @@ async def publish(self, msg_forwarder: ID, pubsub_msg: rpc_pb2.Message) -> None:
...
+                    if peer_id in self.direct_peers:
+                        continue
+                    if peer_id in queued_peers:
+                        continue
+                    # Queue optimistically if peer not in peer_topics
+                    if (
+                        topic not in pubsub.peer_topics
+                        or peer_id not in pubsub.peer_topics[topic]
+                    ):
+                        self._pending_messages[peer_id].append(rpc_msg)
```
pacrob: I'm concerned that there isn't a limit or TTL for queued messages. If a peer never subscribes to a topic, they could accumulate unbounded pending messages and cause memory issues.

libp2p/pubsub/gossipsub.py
```diff
@@ -300,6 +307,126 @@ def add_peer(self, peer_id: ID, protocol_id: TProtocol | None) -> None:
         if self.scorer is not None and self.pubsub is not None:
             self._track_peer_ip(peer_id)
 
+        # NOTE: Pending messages are NOT flushed here because
+        # ``pubsub.peers[peer_id]`` (the stream) is not yet set at this
+        # point.  Flushing happens in ``flush_pending_messages`` which is
+        # called by Pubsub._handle_new_peer after the stream is registered.
+
+    async def flush_pending_messages(self, peer_id: ID) -> None:
```
pacrob, line 569: It appears that the pending message queue accumulates messages on any topic, then flushes them all to the new peer regardless of which topic they subscribed to. Is that the desired behavior? It would seem we would want to filter and flush only relevant messages to the peer.

libp2p/pubsub/gossipsub.py
```diff
@@ -300,6 +307,126 @@ def add_peer(self, peer_id: ID, protocol_id: TProtocol | None) -> None:
...
+
+        stream = pubsub.peers[peer_id]
+        logger.debug(
+            "flushing %d pending message(s) to newly identified peer %s",
+            len(queued),
+            peer_id,
+        )
+        for rpc_msg in queued:
+            try:
+                await pubsub.write_msg(stream, rpc_msg)
```
pacrob, line 640: In the normal publish path (`GossipSub.publish`) we use a `Publish gate`:

```python
if self.scorer is not None and not self.scorer.allow_publish:
```

But here we do not. Should we be making the same checks?

libp2p/pubsub/gossipsub.py
```diff
@@ -300,6 +307,126 @@ def add_peer(self, peer_id: ID, protocol_id: TProtocol | None) -> None:
...
+            # Create RPC with this message
+            rpc_msg = rpc_pb2.RPC(publish=[msg])
+
+            # Add the senderRecord if available
+            if isinstance(pubsub, Pubsub):
+                envelope_bytes, _ = env_to_send_in_RPC(pubsub.host)
+                rpc_msg.senderRecord = envelope_bytes
+
+            try:
+                await pubsub.write_msg(stream, rpc_msg)
```
pacrob, line 724: Same as above: In the normal publish path (`GossipSub.publish`) we use a `Publish gate`:

```python
if self.scorer is not None and not self.scorer.allow_publish:
```

But here we do not. Should we be making the same checks?

seetadev:
Thank you so much @pacrob for the thorough and thoughtful review: appreciate you digging into the edge cases here. Your points around queue bounds/TTL, topic filtering during flush, and maintaining consistency with the allow_publish gate are especially important for correctness and long-term stability.

@imApoorva36 this is shaping up to be an important fix for the early-publish timing issue: could you please address the feedback raised above by @pacrob? 

Specifically:

Introduce a bounded queue or TTL to avoid unbounded memory growth.

Ensure we flush only messages relevant to the peer’s subscribed topics.

Apply the same publish gate (scorer.allow_publish) logic during flush to maintain behavioral parity with the normal publish path.

Please reply below the feedback as you resolve each of the points. CCing @Winter-Soren, @yashksaini-coder  and @sumanjeet0012, who will help you arrive at a good conclusion on the issue. 

Once these are incorporated, this will significantly strengthen the robustness of GossipSub around identify timing and avoid subtle production issues.

Great progress overall — looking forward to the next iteration 🚀

libp2p/pubsub/gossipsub.py
imApoorva36: Good point I think I missed out on that, I’ll add a per-peer cap and TTL so pending messages don't grow without limit

libp2p/pubsub/gossipsub.py
imApoorva36: ah right, I'll make it per topic instead of per peer, good point

libp2p/pubsub/gossipsub.py
imApoorva36: right I missed out on that I believe, will  add `scorer.allow_publish` gate to queued publishes

imApoorva36:
Hey @seetadev @pacrob resolved all review comments, resolved conflicts and all tests are passing as well, I believe this should be merge ready now, do let me now your inputs so we can get towards merging this soon :)

libp2p/pubsub/gossipsub.py
pacrob: @imApoorva36 You've 👍 this, but I don't see there's been any change. Am I missing something?

yashksaini-coder:
### PR Review: GossipSub Identify-Aware Publishing

@imApoorva36 I can see in the PR, if you try to send a message to a peer before the network layer has finished identifying them, the message just gets dropped on the floor — no error, no retry, it's just gone. This PR fixes that by holding those messages in a queue instead. Once the peer is fully identified and their stream is ready, we flush the queue and deliver everything that was waiting.

---

### Critical Issue: Memory Leak

**Must fix before merge.**

| | |
|---|---|
| **Location** | `libp2p/pubsub/pubsub.py:739, 845` |
| **Problem** | Changed `.pop(peer_id).set()` to `[peer_id].set()` — events never removed from dict |
| **Impact** | Unbounded growth on long-running nodes (relays, bootstraps) |
| **Fix** | Restore `.pop()` on both lines |

---

### Minor Issues

| Severity | Issue | Location |
|---|---|---|
| 🟡 | Same message may send twice (pending queue + cache replay) | `pubsub.py:850-868` |
| 🟡 | Queues for all connected peers on every publish (O(n) overhead) | `gossipsub.py:884-902` |
| ⚪ | Test `test_three_nodes_publish_before_full_mesh` has no assertions | Line 350-355 |

---

### Summary

- ✅ Core logic correct, tests pass (104/104)
- ✅ Good test coverage
- 🔴 **Blocker:** Memory leak regression
- 🟡 Consider addressing double-delivery and queuing scope

**Main requirement:** is to Fix the `.pop()` calls, then good for final review + merge.

libp2p/pubsub/gossipsub.py
imApoorva36: @pacrob my apologies, I was of the impression I had added for this as well, although I have now added a new commit adding the same

imApoorva36:
@yashksaini-coder @pacrob @seetadev addressed all review comments

seetadev:
@imApoorva36 : Great, thank you so much for your efforts, Apoorva. Appreciate it.

This is indeed is ready for final review and merge. Will check once again and merge the PR.

PR #1243, opened 2026-02-26
fix: raise ConnectionClosedError on TCP connection reset (#376)

### LINKED ISSUE #376: Add `ConnectionResetError` test case
## What was wrong?
We don't have test that check if `ConnectionResetError` is properly handled

## How can it be fixed?

Add test that triggers `ConnectionResetError` and check if `read`/`write` in `RawConnection` handle the error as expected.

aniruddha1295:
Fixes #376
## What was wrong?

`TrioTCPStream` caught both `trio.BrokenResourceError` and `trio.ClosedResourceError` in a single handler and silently swallowed them — returning `None` on write and `b""` on read. This meant that when a remote peer reset the TCP connection (`BrokenResourceError`), it was treated identically to a local cleanup closure (`ClosedResourceError`).

This caused real-world bugs in downstream projects like Trinity (ethereum/trinity#1113), where crashed peers were silently treated as clean disconnects. The `read_exactly()` utility in `libp2p/io/utils.py` would retry up to 100 times on the `b""` return instead of failing immediately.

Based on issue #376, this PR adds proper error handling and comprehensive tests.

## How was it fixed?

Updated `libp2p/io/trio.py`:
- Split the single `except (ClosedResourceError, BrokenResourceError)` into two separate handlers in both `read()` and `write()`
- `trio.BrokenResourceError` (remote peer reset) now raises `ConnectionClosedError(transport="tcp")` with `from error` chaining
- `trio.ClosedResourceError` (local cleanup) keeps its existing silent behavior for backward compatibility

Added `tests/core/io/test_trio_error_handling.py`:
- 6 write error handling tests (ClosedResourceError silent, BrokenResourceError raises, other exceptions propagate, successful operation, empty data, concurrent write lock)
- 6 read error handling tests (ClosedResourceError returns `b""`, BrokenResourceError raises, other exceptions propagate, successful operation, zero bytes, None parameter)
- 2 RawConnection propagation tests (full chain: `BrokenResourceError` → `ConnectionClosedError` → `RawConnError` for both read and write)
- 2 exception hierarchy and structured attribute tests (`ConnectionClosedError` catchable as `IOException`, `transport="tcp"`, `close_code is None`, `close_reason == ""`)
- 1 stream creation test

Added `newsfragments/376.bugfix.rst`:
- Documents the user-facing bugfix for towncrier release notes

## Backward compatibility

Only `BrokenResourceError` behavior changes. `ClosedResourceError` behavior is untouched.

| Caller | Impact |
|--------|--------|
| `RawConnection` | No change — already catches `IOException`, which `ConnectionClosedError` inherits from |
| Yamux | No change — already has a dedicated `except ConnectionClosedError` handler |
| Mplex | No change — catches `RawConnError`, which RawConnection produces from any `IOException` |
| `read_exactly()` | Improved — fails immediately on connection reset instead of retrying 100 times on `b""` |
| TLS, SwarmConn, SecureSession | No change — existing broad exception handlers catch the new error correctly |

## Note on `--no-verify`

`git commit --no-verify` was used to bypass pre-commit hooks locally. This was necessary because `pyrefly` reports 55 errors across unrelated files (`examples/autotls/`, `libp2p/kad_dht/`, `tests/utils/factories.py`, etc.) that are pre-existing in the codebase and not introduced by this PR. This was confirmed by observing the same failures on the unmodified codebase.

All other hooks passed (ruff, ruff format, mypy, trim trailing whitespace, fix end of files, pyupgrade). All checks were verified to pass when scoped to our changed files:

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -131,8 +131,6 @@ If you want to customize limits, construct and pass your own manager:
 Custom Setup
 ~~~~~~~~~~~~
 
-**NOTE: The current implementation of py-libp2p doesn't yet support dnsaddr multiaddresses. When connecting to bootstrap peers, use direct IP addresses instead.**
-
 Once your libp2p node is running, it is time to get it connected to the public network. We can do this via peer discovery.
 
 Peer Discovery
@@ -144,6 +142,19 @@ For Python, you can use the bootstrap list to connect to known peers:
 
 .. literalinclude:: ../examples/doc-examples/example_peer_discovery.py
    :language: python
+
+Bootstrap address formats
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Bootstrap discovery supports both IP and DNS multiaddrs. You can use:
+
+- **IP addresses:** ``/ip4/127.0.0.1/tcp/4001/p2p/PEER_ID`` or ``/ip6/::1/tcp/4001/p2p/PEER_ID``
+- **DNS (any):** ``/dns/bootstrap.example.com/tcp/4001/p2p/PEER_ID``
+- **DNS IPv4-only:** ``/dns4/bootstrap.example.com/tcp/4001/p2p/PEER_ID``
+- **DNS IPv6-only:** ``/dns6/bootstrap.example.com/tcp/4001/p2p/PEER_ID``
+- **DNS multiaddr resolution (dnsaddr):** ``/dnsaddr/bootstrap.libp2p.io/p2p/QmNnooDu7bfjPFoTZYxMNLWUQJyrVwtbZg5gBMjTezGAJN``
+
+If a DNS bootstrap address fails to resolve or returns no results, the bootstrap module logs a warning and continues with the next address. See :doc:`libp2p.discovery.bootstrap` for API details.
 
 Debugging
 ---------
--- a/docs/libp2p.discovery.bootstrap.rst
+++ b/docs/libp2p.discovery.bootstrap.rst
@@ -1,6 +1,41 @@
 libp2p.discovery.bootstrap package
 ==================================
 
+The bootstrap discovery module connects to a list of bootstrap peer addresses
+to join the network. It supports both IP and DNS multiaddrs.
+
+Supported bootstrap address types
+---------------------------------
+
+- **IP:** ``/ip4/...``, ``/ip6/...`` (e.g. ``/ip4/127.0.0.1/tcp/4001/p2p/PEER_ID``)
+- **DNS:** ``/dns/...``, ``/dns4/...`` (IPv4-only), ``/dns6/...`` (IPv6-only), ``/dnsaddr/...``
+
+Example: using DNS bootstrap addresses
+---------------------------------------
+
+.. code-block:: python
+
+    from multiaddr import Multiaddr
+    from libp2p.peer.peerinfo import info_from_p2p_addr
+
+    # dnsaddr (resolves to one or more multiaddrs)
+    bootstrap_list = [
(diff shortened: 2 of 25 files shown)


aniruddha1295:
@seetadev  @acul71  @sumanjeet0012   can you review this !!

acul71 (changes requested):
Thanks for this PR.
Can you update the PR description to include `Fixes #376` (instead of only contextual references) so the issue-linking policy is fully satisfied?
The test are made with mock connections that is ok, but I was wondering if you have run this as real connections and found in logs the meant behavior, if so can you show a log ?


## 1. Summary of Changes

- **Type:** Bug fix for TCP stream error semantics.
- **What changed:**
  - `libp2p/io/trio.py`: `TrioTCPStream.read()` and `write()` now distinguish local closure vs remote reset:
    - `trio.BrokenResourceError` now raises `ConnectionClosedError(transport="tcp")`.
    - `trio.ClosedResourceError` keeps prior silent behavior (`write` returns; `read` returns `b""`) for compatibility.
  - `tests/core/io/test_trio_error_handling.py`: substantial expansion of coverage around read/write behavior, exception propagation to `RawConnection`, and exception hierarchy/attributes.
  - `newsfragments/376.bugfix.rst`: adds user-facing release note.
- **Related issue context:**
  - PR body references issue `#376` and Trinity follow-up context (`ethereum/trinity#1113`).
  - `#376` asks for tests validating `ConnectionResetError` handling through `RawConnection`.

aniruddha1295:
@acul71 @seetadev 
Ran integration tests with **real TCP connections** (no mocks) to verify the error handling. Used `SO_LINGER(on=1, timeout=0)` to force a true TCP RST (simulating a remote peer crash), not a graceful FIN.

### Test results (4/4 passed)

tests/core/io/test_trio_real_connection.py::test_read_after_abrupt_reset_raises_error
INFO  read() raised RawConnError -> ConnectionClosedError on RST:
TCP connection reset by remote peer: socket connection broken: [Errno 104] Connection reset by peer
PASSED

tests/core/io/test_trio_real_connection.py::test_write_after_abrupt_reset_raises_error
INFO  write() raised RawConnError -> ConnectionClosedError on RST:
TCP connection reset by remote peer: socket connection broken: [Errno 104] Connection reset by peer
PASSED

tests/core/io/test_trio_real_connection.py::test_graceful_close_read_returns_eof
INFO  Graceful close correctly returned EOF (b'')
PASSED

tests/core/io/test_trio_real_connection.py::test_normal_data_transfer_works
INFO  Normal data transfer works as expected
PASSED

4 passed in 21.21s



### Error propagation chain confirmed with real sockets

[Errno 104] Connection reset by peer (OS)
→ trio.BrokenResourceError
→ ConnectionClosedError(transport="tcp")
→ RawConnError



Graceful close (FIN) correctly returns `b""` — only abrupt resets (RST) raise the error.
screenshot:
image

pacrob (approved):
lgtm!

acul71:
> tests/core/io/test_trio_real_connection.py

I think you forgot to commit this 
The [docs/readthedocs.org:py-libp2p](https://app.readthedocs.org/projects/py-libp2p/builds/31640004/) was successful, but the the test is hanging for some reason. 
@pacrob Maybe this can be merged even if the test_trio_real_connection.py ain't been committed, we can trust the post logs ?
In this case this PR can be merged

pacrob:
> Maybe this can be merged even if the test_trio_real_connection.py ain't been committed

Totally missed that. If the test file exists, then it would be great to add it. Maybe it causes a problem with CI or something so it was removed?

aniruddha1295:
@acul71  @pacrob  
Added the test file can you please verify..

Issue #1248, opened 2026-02-26
Setup guide on contributing page does not work on macOS


asabya:
The [setup](https://py-libp2p.readthedocs.io/en/stable/contributing.html#macos-setup) guide with uv on the contributing page does not work.

```
$ uv pip install --group dev -e . 
error: unexpected argument '--group' found
  tip: to pass '--group' as a value, use '-- --group'
Usage: uv pip install [OPTIONS] <PACKAGE|--requirements <REQUIREMENTS>|--editable <EDITABLE>>
```

PR #1249, opened 2026-02-27
docs: update development setup commands and scripts for uv

asabya:
## What was wrong?
Contributing guide with `uv` commands were failing.

Issue #1248

## How was it fixed?

Fixed the `setup_dev.sh` script to support default uv virtual environment directory `.venv` along with the contributing guide

### To-Do

- [x] Check the script and guide on linux and windows


asabya:
https://github.com/user-attachments/assets/c1292bc1-7cdf-422e-af18-d1eecc4b001f

asabya:
@seetadev @acul71 @sumanjeet0012 This needs a bit of love.

seetadev:
@asabya : This is fantastic. Please add a newsfragment.

Adding @pacrob to the thread.

docs/contributing.rst
```diff
@@ -158,36 +153,22 @@ Then set up the development environment:
...
-    python3 -m venv ./venv
-    . venv/bin/activate
+    python3 -m venv .venv
+    source .venv/bin/activate
     pip install --upgrade pip  # Ensure pip >= 25.1 for PEP 735 support
-    pip install --group dev -e .
+    CFLAGS="$(pkg-config --cflags gmp)" LDFLAGS="$(pkg-config --libs gmp)" pip install --group dev -e .
     pre-commit install
 
 On macOS, help the build command find and link against the ``gmp`` library:
```
pacrob: This should be removed with the code section it refers to below.

scripts/setup_dev.sh
```diff
@@ -12,39 +12,58 @@ NC='\033[0m' # No Color
...
+    PIP_VERSION=$(pip --version | grep -oE '[0-9]+\.[0-9]+' | head -1)
+    PIP_MAJOR=$(echo $PIP_VERSION | cut -d. -f1)
+    PIP_MINOR=$(echo $PIP_VERSION | cut -d. -f2)
+
+    # PEP 735 dependency-groups support requires pip >= 25.1
+    if [ "$PIP_MAJOR" -lt 25 ] || ([ "$PIP_MAJOR" -eq 25 ] && [ "$PIP_MINOR" -lt 1 ]); then
+        echo -e "${YELLOW}Upgrading pip to support PEP 735 dependency-groups...${NC}"
+        pip install --upgrade pip
+    fi
+
```
pacrob, line 51: Does the `pip` install on mac need the `gmp` flags too?

docs/contributing.rst
```diff
@@ -47,8 +47,8 @@ Install the development dependencies using a virtual environment:
 .. code:: sh
 
     cd py-libp2p
-    python3 -m venv ./venv
-    . venv/bin/activate
+    python3 -m venv .venv
+    source .venv/bin/activate
     ./scripts/setup_dev.sh
```
pacrob, line 53: The script does work on linux, but it sounds like mac auto-manages the venv, where linux doesn't.

When you run the script on linux, you see this:

```sh
root@0e89be2e03a0:/py-libp2p# ./scripts/setup_dev.sh
Setting up py-libp2p development environment...
Warning: Not in a virtual environment. Creating one...
Virtual environment created. Activate it with:
  source .venv/bin/activate
Then run this script again.
```

Might be worth adding to docs so there are no surprises for new linux users.

pacrob:
Looking good! Couple comments/questions, good to merge once those are addressed.

docs/contributing.rst
```diff
@@ -47,8 +47,8 @@ Install the development dependencies using a virtual environment:
 .. code:: sh
 
     cd py-libp2p
-    python3 -m venv ./venv
-    . venv/bin/activate
+    python3 -m venv .venv
+    source .venv/bin/activate
```
acul71: The use of `.` is a shorthand for `source`. It’s a bit more elegant—would you consider using it instead?"

scripts/setup_dev.sh
```diff
@@ -12,39 +12,58 @@ NC='\033[0m' # No Color
...
 
-# Install pre-commit hooks
-echo -e "${GREEN}Installing pre-commit hooks...${NC}"
-pre-commit install
+    # Install pre-commit hooks
+    echo -e "${GREEN}Installing pre-commit hooks...${NC}"
+    pre-commit install
 
-echo -e "${GREEN}Setup complete! You can now run 'make pr' to check your changes.${NC}"
+    echo -e "${GREEN}Setup complete! You can now run 'make pr' to check your changes.${NC}"
```
acul71: You could also add `make docs` to check docs (gives errors in CI/CD if doc generation has warnings)

docs/contributing.rst
asabya: ok 👍

scripts/setup_dev.sh
asabya: I think so.

pacrob: We should add it to this pip branch of the script then.

docs/contributing.rst
pacrob: Claude keeps poking me that the virtual environment check isn't necessary, even in linux. And indeed, it does seem that the `uv sync` properly syncs in to a `.venv`. So maybe the whole virtual environment check in the `uv` branch of the script isn't even necessary. 

Do you see the same behavior on mac?

docs/contributing.rst
acul71: Didn't ask you specifically but If you change `source` with `.` in a place you should do everywhere

docs/contributing.rst
asabya: uv sync does handle the environment setup. removed it

docs/contributing.rst
asabya: good point.

asabya:
Updated `docs/install` to make the changes consistant.

acul71:
Addressed the remaining blockers and doc consistency items in this PR:

- Added required newsfragment: `newsfragments/1248.docs.rst`
- Updated smoke test example activation to `. .venv/bin/activate`
- Simplified Linux setup Option 1 in `docs/contributing.rst` by removing redundant manual `.venv` creation/activation before `./scripts/setup_dev.sh`
- Added Python compatibility note (`<= 3.13`) once in `docs/contributing.rst` and also in `docs/install.rst`

Validation rerun after changes:
- `make lint` passed
- `make linux-docs` passed

acul71 (approved):
LGTM!
Tested in MacOs, Linux, Windows.

Issue #395, opened 2020-01-10
Restructuring py-libp2p to be more modular and developer-friendly


ShadowJonathan:
I'm proposing a plan of refactoring large parts of the current py-libp2p library to be more modular and developer-friendly through 3 ways:
1. Restructure the network/application side to revolve around 3 basic concepts: Transport, Connection, and Protocol.
2. Split functionality into "interfaces" (abstract classes with abstract methods), so cross-library compatibility can work without the two libraries knowing about eachother. (For example, a Protocol could verify that it's working on a secure connection (where one of the parent protocols is providing connections through SecIO, or similar) by simply testing if one of it's parents is a `ISecureConnection`, which denotes methods to grab the node's verified identity).
3. Split interface from implementation, so that every part of the library that's an implementation (`TCP` protocol, `Multiplexer` protocol, `SecIO` protocol, etc.) implements their interface in a one-way awareness solution (implementation is aware of interface, interface is not aware of implementation)

ShadowJonathan:
Elaboration:

### 1: Restructure core concepts resolve around (Transport, Connection, Protocol)
A Transport takes a `Multiaddr` and dials/generates a `Connection`.
A `Connection` handles `Protocol` negotiation logic. And if required, special write/read logic.
A `Protocol` writes and reads to the underlying `Connection`.

### 2: Add abstract Interfaces to the library
Through this, the library gets extensiblity and modularity even across different implementations;

- A interface `ProvidesTransport` exists.
- Once a `Protocol` has been established, the library will see that the protocol implements that interface, and add that interface to a register for it to call upon when needed.
- A multiaddress that this protocol can dial arrives, the library will walk past all registered transports to find one that matches it's protocol, it eventually finds the `ProvidesTransport`, and dials the multiaddress through it.

Or another one:

- The popular `Multiplex` Protocol is initiated between a peer, it subclasses `ProvidesConnection`
- The node needs another Connection to that peer to initiate another Protocol, it goes through an iterator of `Protocol`s before eventually falling back to initiating a new `Transport`.
- The library finds a class that subclasses `ProvidesConnection`, it asks through that interface for another Connection, the Protocol gives it.

Another one:

- A library's protocol needs security through peer verification.
- A decorator function `@requires_connection(ISecureConnection)` is added to the top of the class, denoting that this Protocol can only be initiated with Connections that subclass that interface.
[...]

PR #1259, opened 2026-03-02
Capability-based transport decoupling and modular packaging

asmit27rai:
### Summary

This PR decouples the QUIC transport from the core library and introduces
a provider-based architecture that allows transports, security protocols,
and stream muxers to be discovered and composed dynamically at runtime.

### Motivation

Previously, the core imported QUIC classes directly, meaning `aioquic`
was a hard dependency even for users who only needed TCP.  Transport
selection was also rigid — the upgrade pipeline didn't know whether a
transport already handled security or muxing, so it couldn't skip
redundant steps.

### What changed

**Capability protocols** — `TransportCapabilities`, `ConnectionCapabilities`,
and `NeedsSetup` are runtime-checkable protocols that let any transport
declare what it provides (security, muxing) without the core importing
concrete classes.

**Requirement decorators** — `@requires_connection(ISecureConn)` and
`@after_connection(ISecureConn)` let protocol handlers and muxer classes
express ordering constraints.  `check_connection_requirements()` validates
them at stream-open time, and `TransportUpgrader` verifies muxer ordering
before negotiation.

**Provider abstractions** — `TransportProvider`, `SecurityProvider`, and
`MuxerProvider` wrap existing transport/security/muxer implementations
behind a uniform interface.  `ProviderRegistry` stores them and supports
bulk registration from legacy option dicts.

**Connection resolver** — `ConnectionResolver` is a pull-based stack
builder.  Given a multiaddr and peer ID, it picks the right transport,
applies security and muxer upgrades in order, and skips layers the
transport already provides.  If one path fails, it falls back to the next
registered transport.

**Capability-aware registry queries** — `TransportRegistry` gained methods
like `transport_provides_security()` and `get_self_upgrading_protocols()`
so callers can inspect transport capabilities without instantiating them.

**Entry-point discovery** — `discover_and_register()` scans
`libp2p.transports`, `libp2p.security`, and `libp2p.muxers` entry-point
groups.  Each built-in component ships a `_create_*_provider()` factory.
Third-party packages can register their own providers by declaring
entry points in their `pyproject.toml`.

**Modular packaging** — `pyproject.toml` now defines optional dependency
groups (`quic`, `noise`, `tls`, `secio`, `yamux`, `mplex`, `defaults`).
Conditional import guards (`_HAS_QUIC`, `_HAS_NOISE`, etc.) let the
library load cleanly when extras are missing.  `create_yamux_muxer_option()`
and the security transport setup in `new_swarm()` check availability
before including components.

### Testing

- 180 new tests across 6 test files
- All 180 pass; full suite shows no regressions (2272 passed, 14 skipped)

### Breaking changes

None.  All changes are additive and opt-in.  Existing code using the fixed
`TransportUpgrader` pipeline continues to work without modification.

### Issue : https://github.com/libp2p/py-libp2p/issues/395


acul71:
Closing this large draft (CI fully red, heavy conflicts). A focused capabilities approach can be proposed later as a fresh PR against current main if needed for #395.

Issue #1276, opened 2026-03-15
GossipSub parity: retry dropped control RPCs under backpressure (doDropRPC -> pushControl equivalent)


IronJam11:
### Description

Please note: This is a feature request on top of the PR https://github.com/libp2p/py-libp2p/pull/1263 as discussed with @acul71 

In py-libp2p GossipSub, when outbound RPC chunks are dropped (queue full or oversized), [drop_rpc()](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) currently logs and discards.
In go-libp2p-pubsub, dropped control RPCs are deferred for retry (via doDropRPC -> pushControl) and later piggybacked/sent.

This issue proposes aligning py-libp2p with that behavior for control messages.

## Go behavior (reference)
In go-libp2p-pubsub:

doDropRPC(...) checks dropped RPC control section.
If control exists, it is pushed into retry path (pushControl) for later piggyback/retry.
Conceptually: drop → keep control intent → retry later opportunistically.

### Motivation

- Better resilience under backpressure.
- Closer GossipSub parity with go-libp2p.
- Reduced risk of mesh instability from transient control loss.

### Current Implementation

- In ``send_rpc()``, dropped parts call ``drop_rpc()``
- ``drop_rpc()`` only logs; no deferred retry path 
- Result: under transient queue saturation, control signaling (e.g. GRAFT, PRUNE, IWANT, IDONTWANT) can be permanently lost.

### Are you planning to do it yourself in a pull request ?

Yes

PR #1283, opened 2026-03-19
feat: implement deferred retry for dropped rpc chunks 

IronJam11:
## What was wrong?

Issue #1276 

In py-libp2p, when outbound `GossipSub` control RPCs or core `PubSub` subscription announcements were dropped—either due to the outbound queue being full or the RPC payload exceeding max limits—they were permanently discarded. This diverged from the `go-libp2p-pubsub` implementation, which explicitely caches and defers dropped `GRAFT` and `PRUNE` control messages, as well as `subscribe`/`unsubscribe` announcements, to ensure strict protocol reliability and mesh stability.

## How was it fixed?

Aligned the connection retry logic fully with the Go implementation:

- **GossipSub Control Retry:** Added `_handle_dropped_rpc`, `_push_control_retry`, and `_piggyback_control_retry` in `gossipsub.py`. Dropped `GRAFT` and `PRUNE` intents are now buffered in `_pending_control` and natively piggybacked on the next outbound RPC chunk for that peer. Gossip data (`ihave`/`iwant`) is explicitly discarded on retry to avoid payload bloat, matching Go perfectly.
- **PubSub Announce Retry:** Handled in `pubsub.py` via `_schedule_announce_retry`. Utilizes asynchronous jitter (`trio.sleep`) before requeuing to prevent traffic spikes, and evaluates the peer's live state via `_announce_state_matches` right before re-sending to ensure stale subscription states aren't blindly transmitted.
- **Refactoring:** Deduplicated outbound payload generation by isolating `_build_announce_rpc` and streamlined the pushing process with the `_enqueue_or_retry_announce` handler.
- **Testing:** Implemented thorough testing fixtures capturing dropping behaviors, queue exhaustion, state cleanup for removed peers, and oversized chunks in `test_rpc_queue.py` and `test_pubsub.py`.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture
image

### DIFF
--- a/libp2p/network/connection/swarm_connection.py
+++ b/libp2p/network/connection/swarm_connection.py
@@ -224,6 +224,13 @@ class SwarmConn(INetConn):
                 nursery.start_soon(self._handle_muxed_stream, stream)
 
     async def _handle_muxed_stream(self, muxed_stream: IMuxedStream) -> None:
+        # Acquire semaphore slot for inbound stream (queues if at capacity)
+        semaphore = getattr(self.swarm, "_stream_semaphore", None)
+        semaphore_acquired = False
+        if semaphore is not None:
+            await semaphore.acquire()
+            semaphore_acquired = True
+
         # Acquire inbound stream resource if a manager is configured
         rm = getattr(self.swarm, "_resource_manager", None)
         peer_id_str = str(getattr(self.muxed_conn, "peer_id", ""))
@@ -235,6 +242,9 @@ class SwarmConn(INetConn):
                 acquired = False
 
         if rm is not None and not acquired:
+            # Release semaphore since we're rejecting this stream
+            if semaphore_acquired and semaphore is not None:
+                semaphore.release()
             # Deny stream: best-effort reset/close
             try:
                 await muxed_stream.reset()  # type: ignore[attr-defined]
@@ -246,18 +256,16 @@ class SwarmConn(INetConn):
             return
 
         net_stream = await self._add_stream(muxed_stream)
+        # Tag stream with direction so notify_closed_stream releases resources
+        net_stream._direction = Direction.INBOUND
         try:
             await self.swarm.common_stream_handler(net_stream)
         finally:
-            # Always remove the stream when the handler finishes
-            # Use simple remove_stream since stream handles notifications itself
+            # Safety net: ensure stream is removed from connection tracking
             self.remove_stream(net_stream)
-            # Release inbound stream resource
-            if rm is not None and acquired:
-                try:
-                    rm.release_stream(peer_id_str, Direction.INBOUND)
-                except Exception:
-                    pass
+            # Release resources via notify_closed_stream (idempotent guard)
+            if not net_stream._resource_released:
+                await self.swarm.notify_closed_stream(net_stream)
 
     async def _add_stream(self, muxed_stream: IMuxedStream) -> NetStream:
         #
--- a/libp2p/network/stream/net_stream.py
+++ b/libp2p/network/stream/net_stream.py
@@ -16,6 +16,7 @@ from libp2p.abc import (
 from libp2p.custom_types import (
     TProtocol,
 )
+from libp2p.rcmgr.metrics import Direction
 from libp2p.stream_muxer.exceptions import (
     MuxedStreamClosed,
     MuxedStreamEOF,
@@ -131,6 +132,11 @@ class NetStream(INetStream):
         self._state = StreamState.INIT
         self.swarm_conn = swarm_conn
         self.logger = logging.getLogger(__name__)
+
+        # Stream direction for resource release tracking
+        self._direction: Direction = Direction.UNKNOWN
+        # Idempotency guard against double-release of RM/semaphore resources
+        self._resource_released: bool = False
 
         # Thread safety for state operations (following AkMo3's approach)
         self._state_lock = trio.Lock()
--- a/libp2p/network/swarm.py
+++ b/libp2p/network/swarm.py
@@ -151,6 +151,7 @@ class Swarm(Service, INetworkService):
         # Load balancing state
         self._round_robin_index = {}
         self._resource_manager = None
+        self._stream_semaphore: trio.Semaphore | None = None
 
         # Initialize connection management components
         self._init_connection_management()
@@ -190,9 +191,17 @@ class Swarm(Service, INetworkService):
         # Initialize tag store for peer tagging/protection (go-libp2p TagPeer, Protect)
         self.tag_store = TagStore()
 
-    def set_resource_manager(self, resource_manager: ResourceManager | None) -> None:
+    def set_resource_manager(
+        self,
+        resource_manager: ResourceManager | None,
+        enable_stream_semaphore: bool = True,
+    ) -> None:
         """Attach a ResourceManager to wire connection/stream scopes."""
         self._resource_manager = resource_manager
+        if resource_manager is not None and enable_stream_semaphore:
+            self._stream_semaphore = trio.Semaphore(resource_manager.limits.max_streams)
+        else:
+            self._stream_semaphore = None
 
     async def run(self) -> None:
         async with trio.open_nursery() as nursery:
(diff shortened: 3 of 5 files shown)


IronJam11:
ccing @acul71 @seetadev @pacrob @Winter-Soren @yashksaini-coder

newsfragments/1276.feature.rst
yashksaini-coder, line 1: File named `newsfragments/1226.feature.rst` — should be `1276.feature.rst` The correct issue is #1276 ("GossipSub parity: retry dropped control RPCs under backpressure")

libp2p/pubsub/pubsub.py
```diff
@@ -1043,9 +1049,71 @@ async def message_all_peers(self, raw_msg: bytes) -> None:
 
                 ok = queue.push(part)
                 if not ok:
-                    drop_rpc(peer_id, part)
+                    self._enqueue_or_retry_announce(peer_id, queue, part, announce)
                     break
```
yashksaini-coder, line 1070: I think we can simplify the error handling here. Right now when the queue is full, we retry the push through _enqueue_or_retry_announce, but since the queue hasn't cleared up, that second attempt is guaranteed to fail too. We're essentially doing the work twice and possibly double-logging the drop.

A cleaner approach would be to handle the full-queue case right at line 1051: if the push fails, immediately drop the RPC and schedule the announcement retry directly, without the indirection. That saves us a redundant operation and keeps the metrics accurate.

 ```python
  if not ok:
      drop_rpc(peer_id, part)
      if announce is not None:
          self._schedule_announce_retry(peer_id, announce)
      break
  ```

libp2p/pubsub/pubsub.py
```diff
@@ -1043,9 +1049,71 @@ async def message_all_peers(self, raw_msg: bytes) -> None:
...
+
+                if retry_rpc.ByteSize() > queue.max_message_size:
+                    drop_rpc(peer_id, retry_rpc)
+                    return
+
+                ok = queue.push(retry_rpc)
+                if ok:
+                    return
+
+                drop_rpc(peer_id, retry_rpc)
```
yashksaini-coder: In the `_run_announce_retry` async def, the loops run indefinitely (`while self.manager.is_running`) with no maximum retry count. If a peer's queue remains permanently full (e.g., a stalled/slow peer), this task runs forever, sleeping 1–1000ms between attempts. Over many topics, this could accumulate many long-running retry tasks per peer.

- **Suggestion:** Add a maximum retry count (e.g., 5–10 attempts) or a total timeout, after which the retry gives up and discards the announcement. The Go implementation has implicit bounds via its heartbeat cycle.

tests/core/pubsub/test_rpc_queue.py
yashksaini-coder, line 1: Nice tests! I'd suggest a few more to cover some gaps I noticed:

- We only test PRUNE deferral everywhere — should add at least one test for GRAFT getting deferred and piggybacked too, and one with both GRAFT + PRUNE in the same dropped RPC.

- `_filter_retriable_control` is the trickiest bit here and has no direct test. I'd add cases like: defer a **GRAFT** for a peer that left the mesh (should get filtered out), defer a PRUNE for a peer that rejoined (should also get filtered), and a case where everything is stale so _pending_control gets cleaned up entirely.

- The buffer cap at 64 isn't tested — push more than 64 PRUNEs via repeated drops and verify only the latest 64 survive. That's our only memory bound so worth a test.

- Drop an RPC that has IHAVE/IWANT but no GRAFT/PRUNE — nothing should end up in `_pending_control`. Good to have an explicit test for the "only retry GRAFT/PRUNE" contract.

- Multiple consecutive drops for the same peer with different topics — verify they coalesce into one _pending_control entry, not overwrite each other.

tests/core/pubsub/test_pubsub.py
yashksaini-coder, line 1: A few more cases that would be good to add:

- Test that the retry exits cleanly when a peer disconnects mid-retry. Remove the peer's queue while a retry is pending and make sure the task stops and the key gets removed from `_pending_announce_retries`.

- I think `remove_peer` should also clear entries from `_pending_announce_retries` — right now it doesn't. A test for this would expose the gap (and probably reveal a small bug).

- Trigger two drops for the same (peer, topic, subscribe) triple and verify only one retry task gets spawned, not two. The dedup set is there but untested.

- Shrink `max_message_size` so the announce RPC itself is oversized, then trigger a retry. Should drop and exit, not loop forever.

- Call `_schedule_announce_retry` after the manager has stopped — should be a no-op with no task spawned.

IronJam11:
Thanks for the thorough review  @yashksaini-coder — I’ve pushed updates addressing the requested items.

I implemented a bounded announce retry loop by adding `_ANNOUNCE_RETRY_MAX_ATTEMPTS = 10` in `pubsub.py`, so retries no longer run indefinitely under sustained backpressure. I also removed the redundant double-push path in `message_all_peers`: on split-path queue-full, we now drop once and schedule announce retry directly.

To prevent stale retry state, I added peer-scoped cleanup of `_pending_announce_retries` during dead-peer handling. I also removed the unused return value from `_enqueue_or_retry_announce` to simplify the call path.

I expanded test coverage with new cases for retry-task deduplication, peer-removal cleanup, oversized announce exit, and max-attempt termination in `test_pubsub.py`; and for GRAFT+PRUNE deferred piggybacking, stale filtering, non-retriable control ignore, and cap/coalescing behavior in `test_rpc_queue.py`.

**Validation status:** lint/type checks pass on touched files, and the new targeted tests pass. Newsfragment naming is now aligned with issue #1276 (`1276.feature.rst`).

yashksaini-coder:
@IronJam11 Checking it will update you, also any reason why this PR was closed ?

IronJam11:
> @IronJam11 Checking it will update you, also any reason why this PR was closed ?

I was thinking of shifting to another branch named correctly (newsfragment and pr were both named incorrectly - leading to confusion), so if you say I can open another clean pr or we can continue on this branch @yashksaini-coder

yashksaini-coder:
> > @IronJam11 Checking it will update you, also any reason why this PR was closed ?
> 
> I was thinking of shifting to another branch named correctly (newsfragment and pr were both named incorrectly - leading to confusion), so if you say I can open another clean pr or we can continue on this branch @yashksaini-coder

No let's not do that, I'm in middle of reviewing the changes currently,

yashksaini-coder:
## PR Review: #1283 — feat: implement deferred retry for dropped rpc chunks

### Fixes:- #1276
---

## 1. Summary

This PR aligns py-libp2p with go-libp2p-pubsub by adding deferred retry logic for
dropped outbound GossipSub control RPCs (GRAFT/PRUNE) and PubSub subscription
announcements.

**Scope:** `libp2p.pubsub` only. No breaking changes, no new dependencies.

---

## 2. Follow-up on Previous Review

The author pushed a follow-up addressing recent feedback, and most of it
landed well:

- ✅ **Max retry cap added** — `_ANNOUNCE_RETRY_MAX_ATTEMPTS = 10` prevents indefinite
  loops under sustained backpressure
- ✅ **Newsfragment naming fixed** — `1276.feature.rst` (was `1226`)
- ✅ **Unused return value removed** from `_enqueue_or_retry_announce`
- ✅ **Dead-peer cleanup added** — `_pending_announce_retries` is now cleared in
  `_handle_dead_peer`
- ✅ **14 new tests** cover retry deduplication, peer-removal cleanup, oversized announce,
  max-attempt termination, GRAFT+PRUNE piggyback, stale filtering, cap/coalescing

---

## 3. Branch Sync & Merge Status

- **Status:** Diverged — 7 commits ahead, **1 commit behind** `origin/main`
- **Merge conflicts:** None
- **Action needed:** Rebase onto `main` and squash commits (this is also the PR's own
  TODO — "Clean up commit history")

---

## 4. Strengths

- ✅ Faithful Go parity: GRAFT/PRUNE-only retry, IHAVE/IWANT explicitly discarded
- ✅ `_pending_control` capped at 64 entries per peer — bounded memory
- ✅ `_filter_retriable_control` validates mesh state before piggybacking stale entries
- ✅ Jittered retry with hard cap avoids thundering herd
- ✅ `_pending_announce_retries` set deduplicates concurrent tasks for the same intent
- ✅ Excellent test coverage — all the tricky edge cases are hit

---

## 5. Issues Found

**Major**

**M1 — Branch needs rebase (merge blocker)**
The branch is 1 commit behind `main` and has a messy merge commit in history. Before
merging, please rebase and squash — this directly addresses the PR's own "Clean up
commit history" TODO.

**M2 — `_push_control_retry` creates a spurious defaultdict entry**
- **File:** [libp2p/pubsub/gossipsub.py](libp2p/pubsub/gossipsub.py)
- The method opens with `pending = self._pending_control[peer_id]`, which immediately
  creates an empty `ControlMessage` in the `defaultdict`, even when the incoming control
  has no GRAFT or PRUNE. The entry is popped at the end, so behaviour is correct, but
  it's unnecessarily noisy. An early guard would clean this up:
  ```python
  if not (control.graft or control.prune):
      return
  pending = self._pending_control[peer_id]
  ```

**Minor**

**m1 — PR title typo:** "deffered" → "deferred"

**m2 — Magic number `64` should be a named constant**
- **File:** [libp2p/pubsub/gossipsub.py](libp2p/pubsub/gossipsub.py#L259)
- `_max_pending_graft_prune_per_peer = 64` is an instance variable initialized inline.
  The announce retry side uses module-level constants (`_ANNOUNCE_RETRY_MAX_ATTEMPTS`),
  so this is inconsistent. A `_MAX_PENDING_GRAFT_PRUNE_PER_PEER = 64` constant at module
  level would make the limit easy to find and change.

**m3 — `_clear_pending_announce_retries_for_peer` is O(n)**
- **File:** [libp2p/pubsub/pubsub.py](libp2p/pubsub/pubsub.py#L802)
- On every dead-peer event, a new set is allocated by iterating every entry. This is
  fine at typical scale, but worth noting. A `dict[ID, set[...]]` structure would make
  this O(1). At minimum, a comment acknowledging the tradeoff would be appreciated.

**m4 — Oversized split path silently drops without retry**
- **File:** [libp2p/pubsub/pubsub.py](libp2p/pubsub/pubsub.py#L1053)
- In the slow (split) path of `message_all_peers`, an oversized chunk is dropped with
  `continue` and no retry is scheduled — only `queue.push` failures trigger a retry.
  In practice, a subscription RPC is small enough that this won't matter, and
  `_run_announce_retry` handles the oversized case anyway. But it's worth a brief comment
  acknowledging the asymmetry intentionally.

---

## 6. Security

- Adversarial peer keeping its queue full → bounded to one retry task per
  `(peer, topic, subscribe)` triple, hard-capped at 10 attempts. Low risk.
- `_pending_control` accumulation for slow peers → capped at 64 GRAFT + 64 PRUNE per
  peer. Low risk.

No new external inputs, no cryptographic changes.

---

## 7. Newsfragment

- ✅ Issue linked: Fixes #1276
- ✅ File: `newsfragments/1276.feature.rst`
- ✅ Content valid and user-facing
- ✅ Previous blocker (wrong number `1226`) resolved

---

## 8. Build Results

| Check | Result |
|---|---|
| `ruff check` | ✅ Clean |
| `mypy` | ✅ No issues |
| Tests | ✅ **98 passed**, 0 failed, 0 skipped (70s) |

All 14 new tests pass.

---

## 9. Recommendations

1. Rebase onto `main` and squash commits before merge
2. Add early guard in `_push_control_retry` to skip defaultdict access when there's
   nothing retriable (M2)
3. Move `64` to a named module-level constant (m2)
4. Fix PR title typo: "deffered" → "deferred"
5. Add a comment in `_run_announce_retry` explaining why `_build_announce_rpc` is
   called on each attempt (listen addresses can change, so it's intentionally not cached)

---

## 10. Questions for Author

1. **PRUNE for untracked topics:** In `_filter_retriable_control`, PRUNE entries are
   kept when `topic_id not in self.mesh`. If we've fully left a topic, is it still
   meaningful to defer-send a PRUNE, or should that be discarded?

2. **`_announce_state_matches` uses `subscribed_topics_receive` only.** Could
   `subscribed_topics_send` and `subscribed_topics_receive` be out of sync in a way
   that causes a retry to misfire?

3. **`remove_peer` in gossipsub.py clears `_pending_control` but not
   `_pending_announce_retries`** (which lives in pubsub.py). Is `_handle_dead_peer` in
   pubsub always guaranteed to run alongside GossipSub's `remove_peer`? If not, there's
   a window where announce retries persist for a removed peer.

---

## 11. Overall Assessment

| Criterion | Rating |
|---|---|
| **Code Quality** | Good |
| **Security** | No issues |
| **Testing** | Excellent |
| **Documentation** | Good |
| **Newsfragment** | ✅ Valid |

The core implementation is solid and the previous review's blockers are resolved. The
remaining work is mostly housekeeping: rebase, fix the title typo, and two small code
quality nits. I'd be happy to approve after those are addressed.

@IronJam11 Please resolve the minor issues and rebase to sync the branch for getting merge readiness.

IronJam11:
on it @yashksaini-coder

IronJam11:
@yashksaini-coder I have addressed your points made. Pls have a check and thank you for your time to review this PR.

yashksaini-coder:
LGTM @seetadev Ready for final review + merge

seetadev:
@IronJam11 : This is a strong and well-executed contribution, Aaryan👏 Appreciate your efforts.

@yashksaini-coder : Thank you so much Yash for your review and feedback. Very helpful.


@acul71 :  This PR indeed improves protocol reliability by aligning py-libp2p with go-libp2p’s deferred retry semantics for dropped RPCs, especially the handling of GRAFT/PRUNE via piggybacking and the thoughtful exclusion of IHAVE/IWANT to avoid payload bloat. The design feels clean and intentional, with clear separation of concerns across _pending_control, announce retry logic, and filtering of stale state before re-sending.

Appreciate the attention to edge cases and operational robustness, bounded buffers, jittered retries with a capped attempt limit, and proper cleanup on peer removal all contribute to a production-grade implementation. The additional test coverage is excellent and gives strong confidence in correctness across tricky scenarios like queue exhaustion, oversized payloads, and retry deduplication.

It’s also great to see the quick turnaround on review feedback, addressing the retry cap, fixing the newsfragment, improving cleanup paths, and tightening the flow shows strong ownership and iteration. At this point, this looks in great shape and aligned for merge. This will be a valuable improvement for pubsub stability in py-libp2p.

PR #1313, opened 2026-04-20
Fix unbound stream variable in DHT provider_store.py

### LINKED ISSUE #1335: DHT ADD_PROVIDER fails with unbound stream when new_stream errors
## Problem

When advertising a content provider via the Kademlia DHT, `_send_add_provider` can fail before `stream` is assigned (e.g. `host.new_stream` raises). The `finally` block then calls `await stream.close()` on an unbound local, producing:

```
Error sending ADD_PROVIDER to <peer>: cannot access local variable 'stream' where it is not associated with a value
```

This masks the underlying connection/stream error.

## Expected

Cleanup should only close a stream if one was opened; the original error should be logged without a secondary `UnboundLocalError`.

## Fix

PR #1313 initializes `stream = None` and closes only when `stream is not None`.

TimSC:
Fixes #1335

## What was wrong?

When `host.new_stream` fails before a stream is opened, `_send_add_provider` hit an unbound `stream` in `finally`, masking the real error:

```
Error sending ADD_PROVIDER to <peer>: cannot access local variable 'stream' where it is not associated with a value
```

## How was it fixed?

Initialize `stream = None` and only close when a stream was opened.

## Test plan

- [x] Unit test: `new_stream` failure returns `False` without `UnboundLocalError`
- [x] `make pr`
- [x] `make linux-docs`

#### Cute Animal Picture

Zebrasoma desjardinii


acul71:
Hi @TimSC — thank you for opening [#1313](https://github.com/libp2p/py-libp2p/pull/1313) and for the clear diagnosis of the `ADD_PROVIDER` / unbound `stream` bug. The fix (`stream = None` plus guarded `close` in `_send_add_provider`) was already correct; we pushed a follow-up commit to your `provider_store` branch so the PR meets the usual libp2p merge requirements.

**What was added**

- **Tracking issue:** [#1335](https://github.com/libp2p/py-libp2p/issues/1335) — documents the failure mode when `new_stream` errors before a stream is opened.
- **Newsfragment:** `newsfragments/1335.bugfix.rst` — user-facing release note for towncrier.
- **Regression test:** `test_send_add_provider_new_stream_failure` in `tests/core/kad_dht/test_unit_provider_store.py` — mocks `new_stream` raising and asserts `_send_add_provider` returns `False` without a secondary `UnboundLocalError`.
- **PR description:** updated with `Fixes #1335` and a completed test plan checklist.
- **Validation:** `make pr` and `make linux-docs` both passed locally before push.

**Branch**

Latest commit on `TimSC:provider_store`: `b452e978` — *Add issue, newsfragment, and test for ADD_PROVIDER stream cleanup*.

No further changes to your core fix in `libp2p/kad_dht/provider_store.py`. From here the PR should be ready for maintainer review; happy to adjust anything if you’d like a different wording on the issue or newsfragment.

Thanks again for contributing to py-libp2p.

PR #1317, opened 2026-04-22
fix(websocket): handle IPv6 4-tuple in listener getsockname()

### LINKED ISSUE #1316: fix(websocket): listener fails to bind IPv6 due to getsockname() unpack
## Problem

`WebsocketListener.listen()` in `libp2p/transport/websocket/listener.py` unpacks `sock.getsockname()` directly into `(sock_addr, sock_port)`. `socket.getsockname()` returns a 2-tuple for IPv4 but a 4-tuple `(host, port, flowinfo, scopeid)` for IPv6, so the unpack raises `ValueError: too many values to unpack` whenever a WebSocket listener binds an IPv6 address. That ValueError is caught by the outer `except Exception` and re-raised as `OpenConnectionError("Failed to listen on ...")`, so IPv6 WebSocket listeners simply cannot start today.

## Flagged in

Review of PR #1308 / inline comment on `libp2p/transport/websocket/listener.py:289` — follow-up noted by the PR author.

## Proposal

Accept any `sock.getsockname()` tuple of length ≥ 2 where the first two elements are a valid `(host: str, port: int)` pair, and treat anything else as a soft fallback (log + keep the requested host/port). Extract the parsing into a small testable helper so both IPv4 and IPv6 code paths are covered by unit tests that don't depend on binding an actual IPv6 socket in CI.

## Acceptance criteria

- [ ] IPv6 `getsockname()` 4-tuples are handled without raising.
- [ ] Unrecognised shapes fall back gracefully (warning + original host/port).
- [ ] Unit test covers at least: IPv4 2-tuple, IPv6 4-tuple with zero and nonzero scopeid, unexpected shapes.
- [ ] Changelog entry (internal).

yashksaini-coder:
## Summary

Fixes #1316 — `WebsocketListener.listen()` could not bind any IPv6 address because it unpacked `socket.getsockname()` directly into a 2-tuple, which fails for IPv6 (which returns a 4-tuple `(host, port, flowinfo, scopeid)`).

## Context

The `ValueError` from the bad unpack was caught by the outer `except Exception` in `listen()` and re-raised as `OpenConnectionError("Failed to listen on ...")`, so IPv6 WebSocket listeners failed silently-but-hard rather than exposing the real cause. Flagged as a follow-up from the PR #1308 review (inline comment on `libp2p/transport/websocket/listener.py:289`).

## Approach

Extract the socket-name parsing into `_extract_host_port_from_sockname()`, a small helper that accepts any tuple of length ≥ 2 where the first two elements are a `(host: str, port: int)` pair. Any other shape returns `None`, and the listener logs a warning and falls back to the originally requested host/port. Because the helper is a pure function, it can be unit-tested without binding an actual IPv6 socket (which is not portable across all CI environments).

```python
sock_name = sock.getsockname()
host_port = _extract_host_port_from_sockname(sock_name)
if host_port is not None:
    actual_host, actual_port = host_port
else:
    logger.warning("Unexpected getsockname() result %r; falling back to requested host/port", sock_name)
```

## Files changed

- `libp2p/transport/websocket/listener.py` — add `_extract_host_port_from_sockname`, use it in `listen()`
- `tests/core/transport/websocket/test_extract_host_port.py` — new unit tests (IPv4 2-tuple, IPv6 4-tuple with zero and nonzero scopeid, unrecognised shapes)
- `newsfragments/1316.bugfix.rst` — changelog entry

## Testing

- `pytest tests/core/transport/websocket/test_extract_host_port.py tests/core/transport/websocket/test_websocket.py` — 52 passed (including 4 new cases)
- `make lint` clean (ruff, ruff-format, mypy, pyrefly, path audit)

## References

- Issue: #1316
- Flagged during review of: #1308


yashksaini-coder:
Addressed Copilot feedback in cc5813bf:

- `_extract_host_port_from_sockname` parameter is now `sock_name: object` rather than `Any` — matches the runtime `isinstance` checks and drops the unnecessary `typing.Any` dependency in the annotation.
- The "unexpected getsockname() result" branch now explicitly re-assigns `actual_host, actual_port = host, port` before logging the warning, so the "falling back to requested host/port" message is self-contained in the code flow rather than relying on pre-initialised values.

PR #1328, opened 2026-05-09
Handle malformed identify varint prefixes

SanketMeghale:
## What was wrong?

`parse_identify_response` probes for the newer length-prefixed identify format before falling back to the legacy raw protobuf format. If that prefix probe raises `ParseError`, the parser should keep trying the legacy path instead of treating the message as fully failed.

## How was it fixed?

The identify parser now treats `ParseError` from the length-prefix probe as a failed prefixed parse and continues to the raw protobuf fallback. I added a focused test that forces the prefix parser to raise and verifies that a valid raw identify message still parses successfully.

### To-Do

- [x] Clean up commit history
- [x] Add entry to the release notes


acul71:
Closing as obsolete / no longer relevant.

This PR catches `ParseError` from `decode_varint_with_size` so `parse_identify_response` can fall back to raw protobuf. On current `main`, `decode_varint_with_size` does not raise `ParseError` — it returns a decoded value for truncated/malformed prefixes and only raises `ValueError` for an overlong varint. Empty input is already rejected earlier in the parser.

The added test only exercises this path by monkeypatching the decoder to raise `ParseError`, which does not match the real decoder behavior.

Identify parsing on `main` has also moved on substantially since this branch (stricter length checks, empty rejection, `public_key` validation, etc.), and the PR now conflicts with `main`.

Thanks for the contribution — happy to revisit if `decode_varint_with_size` starts raising on malformed prefixes and we need a real fallback path again.

Issue #312, opened 2019-09-24
Provide utilities to persist network identities between runs of a node


ralexstokes:
We are currently missing the ability to persist node identities b/t runs of a node.

The user should provide a "key pair provider" when creating a host; I'd expect the type to be `Callable[[], libp2p.crypto.keys.KeyPair]` for maximum flexibility.

There are different methods of providing key pairs based on the "lifetime" of the key pair:

1. New key pair for some cryptosystem; for example, the user provides a function with the above signature that returns the result of `libp2p.crypto.rsa.create_new_key_pair` with the key parameters passed in via closure, curry, etc.

2. Loading a persisted key pair where the user provides a path to a serialized key pair which the library then loads into memory; for example, a callable that closes over the path which, when called does the necessary IO to load and return a `KeyPair` object.

Having persistence opens some questions:

1. Serialization/keystore format

2. How to manage multiple key pairs and select b/t them... I'd think the nice thing is to "auto-save" every key pair the library makes at some default keystore path -- the user then needs a way to specify one key from this keystore.

mhchia:
I'm thinking if it makes sense to provide persisting versions of `KeyStore`, `PeerStore`, ...etc. It can be passed into `initialize_default_swarm` as the parameter `peerstore_opt`. For reference, go-ipfs has [`FSKeystore`](https://github.com/ipfs/go-ipfs/blob/master/keystore/keystore.go#L34-L37) whose backend is a file.

ralexstokes:
oh yeah -- i think we definitely want these as well! i'd expect we are workings towards parity w/ the Go implementation feature-wise -- more a question of what resources get dedicated to this repo and on what timeframe.

when i opened this issue, i definitely had the specific use case in mind for trinity/eth2

moshemalawach:
Implemented something similar in aleph-im/pyaleph, coming from configuration or file (if the key is there, load it, if it's none, and a listening host, print out the key to prompt user to save it. If not listening, a new host id isn't an issue): https://github.com/aleph-im/pyaleph/blob/22b06c02b82f0759b81142156ad3b978aee485c4/src/aleph/services/p2p/manager.py#L16

Smartdevs17:
[...]

PR #1331, opened 2026-05-13
fix(peerstore): load persisted keypairs on peerstore restore

bhuvan-somisetty:
Keypairs were being serialized and saved to the datastore correctly, but the deserialization path on restore was completely bypassed. Both the sync and async implementations had the loading code commented out with a `pass` placeholder and a TODO note, meaning every time a node restarted and the peerstore was rehydrated from SQLite, the peer's public and private keys would silently disappear. Addresses, metadata, and protocols came back fine, but keys did not.

This is a direct blocker for #312. A persistent peerstore that drops keypairs on restart can't actually provide persistent node identity, which is the whole point.

The fix calls `deserialize_public_key` and `deserialize_private_key` (already present in `libp2p.crypto.serialization`) when loading key data from the datastore. The serialization side was already correct; only the load path was missing. Both the sync `SyncPersistentPeerStore` and async `AsyncPersistentPeerStore` are updated symmetrically.

Three tests cover the new behavior: SQLite round-trip for sync, SQLite round-trip for async, and a sanity check that the memory backend correctly does not persist across instances.


acul71:
Superseded by https://github.com/libp2p/py-libp2p/pull/1467 (and tracking issue https://github.com/libp2p/py-libp2p/issues/1466).

That PR reimplements the keypair restore fix cleanly on current `main` (no DMP template noise, includes newsfragment + corrupt-blob coverage). Please close this PR in favor of #1467 if you are happy with that approach — thanks for the original work here.

acul71:
Closing as superseded by #1467 (tracking: #1466). Thanks for the original work here.

PR #1339, opened 2026-05-30
Fix: signer identity validation for signed PeerRecords

sumanjeet0012:
## What was wrong?

Fixes #1338

KadDHT accepted signed PeerRecords where the signer identity did not match record.peer_id.

Although the envelope signature was valid, the signer public key embedded in the envelope could derive to a different peer ID than the one claimed in the record payload.

This allowed authenticated but incorrectly bound peer records to poison the certified address book for arbitrary peer IDs.

## How was it fixed?

Added signer identity binding validation before accepting signed peer records.

The fix enforces:

`signer_peer_id = ID.from_pubkey(envelope.public_key)`

Records with mismatched signer identities are now rejected before they can update certified peer addresses.

### To-Do

- [x] Clean up commit history
- [x] Add or update documentation related to these changes
- [x] Add entry to the release notes

#### Cute Animal Picture

PR #1348, opened 2026-06-19
fix(kad-dht): bind signed PeerRecord to its signer identity

### LINKED ISSUE #1338: KadDHT accepts signed PeerRecords with mismatched signer identity
### Summary

KadDHT currently accepts signed `PeerRecord` where the signer identity does not match `record.peer_id`.

Although the envelope signature is valid, the signer public key embedded in the envelope can derive to a different peer ID than the one claimed inside the record payload.

This can allow certified address poisoning for arbitrary peer IDs.

## Problem

Current validation ensures:

- the envelope signature is valid
- `record.peer_id == msg.id`

However, it does not enforce:

```python
ID.from_pubkey(envelope.public_key) == record.peer_id
```

As a result, an attacker can:

1. Create a `PeerRecord` claiming to belong to a victim peer ID
2. Add attacker-controlled multiaddrs
3. Sign the record using the attacker's own private key
4. Send the signed record through KadDHT
5. Poison the certified address book for the victim peer ID

### Expected behavior

KadDHT should reject signed `PeerRecord`s where the signer identity does not match `record.peer_id`.

### Actual behavior

KadDHT currently accepts signed `PeerRecord`s where the signer identity does not match `record.peer_id`.

### Possible Solution

Enforce signer identity binding before accepting a signed peer record:

```python
ID.from_pubkey(envelope.public_key) == record.peer_id
```

The validation should ideally be added in:

```python
PeerStore.consume_peer_record()
```

so that all current and future ingestion paths share the same invariant.

### Would you like to work on fixing this bug ?

Yes

## Credit

This issue was highlighted by Yann Lorwyn.

yashksaini-coder:
Fixes #1338.

`maybe_consume_signed_record` accepted a signed `PeerRecord` whenever the claimed `peer_id` matched, without checking that the envelope's signer key derived to that `peer_id`. A valid signature from an unrelated key could therefore certify attacker-controlled addresses for an arbitrary victim peer ID (certified-address-book poisoning).

It now rejects any record where `ID.from_pubkey(envelope.public_key) != record.peer_id`, on both the `senderRecord` and `signedRecord` paths.

Adds `tests/core/kad_dht/test_signed_record_signer_binding.py` covering the forged-record rejection (both paths) and the self-signed acceptance path. Full `tests/core/kad_dht` suite passes.


acul71:
Closing as duplicate of #1339.

acul71:
Closing as duplicate of #1339, which merged the same fix at the peerstore layer (`PeerStore.consume_peer_record`) — a more central enforcement point that covers KadDHT and all other ingestion paths. Thanks @yashksaini-coder for the overlap and for offering the regression tests.

PR #1351, opened 2026-06-25
fix: detect unsupported transport multiaddrs before dialing (issue #391)

### LINKED ISSUE #391: network/swarm.py and transport/tcp is not smart enough to detect that peerstore multiaddress is not reachable
## What was wrong?

In py-libp2p, as of right now, there only exists one network implementation, and one 
transport implementation: `Swarm` and `TCP`.

When the swarm would encounter a multiaddress in it's peerstore with values like `udp/9090` or `tcp/9090/quic`, it would not detect those, and instead fail with `multiaddr.exceptions.ProtocolLookupError` over [here](https://github.com/libp2p/py-libp2p/blob/3b9d7c7acd7af955b4e1855effac85aec44ca81f/libp2p/transport/tcp/tcp.py#L73-L74).

## How can it be fixed?

Add a generic function under `transport/` which iterates over all registered `ITransport` classes, and add a function to `ITransport` that accepts one `Multiaddr`, and returns `True` when that multiaddress is supported by that transport. 

Make `Swarm` try to acquire that `ITransport` upon trying a `Multiaddr`, and fail with `SwarmException` when there is no `ITransport` that supports that `MultiAddr`.

paschal533:
Fixes #391.

## What was happening

When the peerstore contained multiaddrs that the registered transport doesn't handle — for example, `/udp/9090` or `/udp/9090/quic` addresses when only TCP is registered — `dial_peer()` would call `transport.dial()` for each of those addresses anyway. That eventually reached `TCP._dial_resolved()`, which called `maddr.value_for_protocol("tcp")` without guarding against the `ProtocolLookupError` it raises on non-TCP multiaddrs. The exception propagated all the way up uncaught, crashing the caller with an error that gives no indication of what actually went wrong.

Making it worse: the default `RetryConfig` has `max_retries=3` and `initial_delay=0.1s` with `backoff_multiplier=2.0`. So for each unsupported address, the stack would sleep ~0.1s + ~0.2s + ~0.4s = **~700ms** before finally raising the wrong exception. A peer with two bad addresses (UDP + QUIC) would waste **~1.4 seconds** before crashing.

## The fix

Three small, self-contained changes:

**`ITransport` gets a `can_dial(maddr)` method** (default `True`) so the interface can express whether a transport handles a given multiaddr, without breaking existing custom transports that don't implement it.

**`TCP.can_dial()` returns `False` for any multiaddr without a `tcp` component.** The check is a simple protocol lookup, ~30 µs — negligible compared to the cost of an actual TCP handshake (~1–100 ms).

**`Swarm.dial_peer()` filters addresses through `can_dial()` before entering the dial loop.** If no address passes, it raises `SwarmException("No supported transport for any address of peer ...")` immediately, before touching the transport or sleeping.

`TCP._dial_resolved()` also now wraps the `value_for_protocol("tcp")` call in a try/except and re-raises as `OpenConnectionError`, consistent with how `listen()` already handled it. This is a defence-in-depth change — the `can_dial()` filter should prevent us from ever reaching this path with a bad address, but if something slips through it will still produce a clean error.

Note: QUIC and WebSocket already had `can_dial()` methods on their transport classes, but nothing was calling them. This PR formalises the pattern and wires it into `Swarm`.

## What it looks like now

Before:
```
ProtocolLookupError: <no tcp in multiaddr>   # wrong layer, no context
```

After:
```
SwarmException: No supported transport for any address of peer <id>: [/ip4/127.0.0.1/udp/9090]
```

## Testing

Five new regression tests in `tests/core/network/test_swarm.py`, all verified in a clean Docker environment (Python 3.13.14 / Linux):

```
test_tcp_can_dial_returns_true_for_tcp_addr   PASSED
test_tcp_can_dial_returns_false_for_udp_addr  PASSED
test_tcp_can_dial_returns_false_for_quic_addr PASSED
test_swarm_skips_unsupported_multiaddrs       PASSED   ← regression test for #391
test_swarm_can_dial_filters_unsupported       PASSED
```

`Dockerfile.issue391` and `benchmarks/bench_issue391.py` are included so the numbers can be reproduced:

```
can_dial() guard cost:          ~30 µs/call
dial_peer() before (no filter): ~620 µs/call  (no retry, 2 bad addrs)
dial_peer() after  (filtered):  ~154 µs/call
Speedup (no retry baseline):    4x

With default max_retries=3 the before-fix cost is ~1.4 s of sleep alone,
before raising ProtocolLookupError on a peer that simply has no TCP address.
```

## Backwards compatibility

`ITransport.can_dial()` defaults to `True`, so any existing transport that doesn't implement it continues to work exactly as before.

### DIFF
--- a/Dockerfile.issue391
+++ b/Dockerfile.issue391
@@ -1,0 +1,45 @@
+FROM python:3.13-slim
+
+LABEL description="py-libp2p issue #391 regression tests + benchmark"
+
+WORKDIR /app
+
+# System deps needed to compile native extensions (coincurve, fastecdsa, etc.)
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    cmake \
+    pkg-config \
+    libgmp-dev \
+    git \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install uv for fast dependency resolution
+RUN pip install --no-cache-dir uv
+
+# Copy the full repo into the image
+COPY . /app/
+
+# Create venv, install the library + test deps
+RUN uv venv /app/venv && \
+    uv pip install --python /app/venv/bin/python --no-cache-dir \
+        -e "." \
+        "pytest>=7.0.0" \
+        "pytest-trio>=0.5.2" \
+        "pytest-timeout>=2.4.0" \
+        "pytest-xdist>=2.4.0" \
+        "factory-boy>=2.12.0,<3.0.0" \
+        "kyber-py>=0.9.0"
+
+ENV PATH="/app/venv/bin:${PATH}"
+ENV PYTHONUNBUFFERED=1
+
+# Default: run the issue #391 regression suite + benchmark
+CMD bash -c "\
+  echo '=== Issue #391 Regression Tests ===' && \
+  pytest tests/core/network/test_swarm.py \
+    -k 'can_dial or swarm_skips or swarm_can_dial or tcp_can_dial' \
+    -v --tb=short && \
+  echo '' && \
+  echo '=== Issue #391 Benchmark ===' && \
+  python benchmarks/bench_issue391.py \
+"
--- a/benchmarks/bench_issue391.py
+++ b/benchmarks/bench_issue391.py
@@ -1,0 +1,236 @@
+#!/usr/bin/env python3
+"""
+Benchmark: Issue #391 — can_dial() pre-filter in Swarm.dial_peer()
+
+The bug: when the peerstore contained multiaddrs that the registered transport
+cannot handle (e.g. UDP/QUIC addresses when only TCP is registered), the old
+code called transport.dial() for each address anyway and crashed with an
+unhandled ProtocolLookupError.
+
+The fix: add can_dial() to ITransport, implement it in TCP, and filter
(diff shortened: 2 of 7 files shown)


acul71:
Closing as superseded.

The core fix for #391 — detecting unsupported peerstore multiaddrs before dialing and raising a clear `SwarmException` — landed via #1357 (`TransportManager` + `ITransport.can_dial` / `transport_for_dialing`). On current `main`, `Swarm.dial_peer()` already filters addresses with `transport_manager.transport_for_dialing(addr)` and raises `SwarmException("No supported transport found for peer ...")`.

This branch is 500+ commits behind and targets the pre–multi-transport `self.transport.can_dial(...)` API, so rebasing would rewrite the same work that is already on `main`.

Thank you @paschal533 for the clear write-up and regression-test ideas — I'll open a small follow-up for the remaining defence-in-depth in `TCP._dial_resolved` (wrap `ProtocolLookupError` as `OpenConnectionError`, matching `listen()`) and a Swarm-level regression test adapted from this PR.

acul71:
Follow-up opened: https://github.com/libp2p/py-libp2p/pull/1465

PR #1372, opened 2026-07-07
fix(interop): handle IPv6-only multiaddrs in _get_ip_value

### LINKED ISSUE #1371: interop transport: _get_ip_value crashes on IPv6-only multiaddrs
## Summary

Transport interop listeners can crash while publishing their dial address when the host advertises an IPv6-only multiaddr (e.g. `/ip6/::1/tcp/.../p2p/...`).

## Root cause

`interop/transport/ping_test.py` implements `_get_ip_value` as:

```python
return addr.value_for_protocol("ip4") or addr.value_for_protocol("ip6")
```

`multiaddr.value_for_protocol("ip4")` raises `ProtocolLookupError` when the address does not contain `ip4`, so the IPv6 fallback is never reached.

## Impact

In unified-testing transport interop (python × python), this caused flaky failures (~70% fail rate in a recent run): listeners exited before publishing an address whenever libp2p returned IPv6 loopback first instead of the Docker bridge IPv4 address.

Example error:

```
multiaddr.exceptions.ProtocolLookupError: MultiAddr '/ip6/::1/tcp/46499/p2p/...' does not contain protocol ip4
```

## Expected behavior

`_get_ip_value` should safely return the IPv4 or IPv6 component when present, and `None` otherwise — matching the pattern already used in `interop/perf/perf_test.py`.

## Proposed fix

Catch `multiaddr.exceptions.ProtocolLookupError` when probing `ip4`/`ip6` protocols.

Verified locally: after the fix, all 13 python×python transport interop combinations pass in unified-testing.

acul71:
## Summary

Fixes flaky transport interop listener failures when libp2p advertises IPv6-only multiaddrs (e.g. `/ip6/::1/tcp/.../p2p/...`).

Closes #1371

## Problem

`_get_ip_value` used:

```python
return addr.value_for_protocol("ip4") or addr.value_for_protocol("ip6")
```

`value_for_protocol("ip4")` raises `ProtocolLookupError` on IPv6 addresses, so the IPv6 fallback never executed. Listeners crashed in `_get_publishable_address` before publishing a dialable address.

This caused intermittent failures in unified-testing python×python transport interop (~70% fail rate in one run with 32 workers).

## Fix

Probe `ip4` and `ip6` separately, catching `ProtocolLookupError` — aligned with the pattern already used in `interop/perf/perf_test.py`.

Also updated `examples/interop/local_ping_test.py` for consistency.

## Test plan

- [x] unified-testing transport interop: all 13 python×python combinations pass after this change (local Docker run with Python 3.13 image)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/interop/local_ping_test.py
+++ b/examples/interop/local_ping_test.py
@@ -35,6 +35,7 @@ from libp2p.security.noise.transport import (
     Transport as NoiseTransport,
 )
 from libp2p.utils.address_validation import get_available_interfaces
+from libp2p.utils.multiaddr_utils import extract_ip_from_multiaddr
 
 PING_PROTOCOL_ID = TProtocol("/ipfs/ping/1.0.0")
 PING_LENGTH = 32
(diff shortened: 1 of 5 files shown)


acul71:
I'm merging this. 
tests pass in perf and transport tests in unified-testing (local runs)

PR #1393, opened 2026-07-18
fix(quic): resolve double-start race condition in QUICConnection lifecycle

### LINKED ISSUE #1389: bug: QUICConnection.start() called twice for inbound connections — Listener and Swarm race on connection lifecycle ownership
## Summary

When an inbound QUIC connection is accepted by the `QUICListener`, the connection's `start()` method is called by the listener. The same connection object is then passed to `Swarm.add_conn()`, which calls `start()` a second time. Because `QUICConnection.start()` guards against re-entry with a `_started` boolean flag and returns early, the Swarm's initialization path — including stream muxer setup and peer registration — is silently skipped.

This results in a sequence of cascading errors: inbound streams cannot be written to, connections cannot be tracked for cleanup, and eventually the connection is abandoned in a partially-initialized state.

## Affected Code

**File:** `libp2p/transport/quic/connection.py`

```python
async def start(self) -> None:
    if self._started:  # <-- guard triggers on second call from Swarm
        logger.warning("Connection already started")
        return  # <-- Swarm's setup steps are never executed
    self._started = True
    ...
```

**File:** `libp2p/transport/quic/listener.py` (around line 1456-1468)

```python
async def _handle_new_established_connection(self, connection: QUICConnection) -> None:
    # Listener calls start() during connection setup
    # Then passes the same object to swarm.add_conn(), which calls start() again
    if self._transport._swarm:
        await self._transport._swarm.add_conn(connection)
```

## Observed Log Output

Running a py-libp2p node with QUIC transport enabled and multiple inbound connections:

```
Connection already started
Connection already started
Error writing to stream 1: Cannot send data on unknown peer-initiated stream
Stream 1 error: Cannot send data on unknown peer-initiated stream
Connection object not found in tracking
```

## Root Cause Analysis

The ownership of the connection lifecycle is ambiguous between the `QUICListener` and the `Swarm`:

1. `QUICListener` creates a `QUICConnection`, calls `start()` on it, and marks `_started = True`.
2. `QUICListener` then passes the connection to `Swarm.add_conn()`.
3. `Swarm.add_conn()` calls `connection.start()` as part of its standard connection setup contract.
4. `start()` detects `_started = True`, logs a warning, and returns immediately.
5. Any Swarm-side stream multiplexer initialization or peer tracking that was supposed to happen inside or after `start()` is skipped.
6. Subsequent stream operations (`send_data`, `recv_data`) fail because the stream is unknown to the Swarm's tracking structures.
7. On connection close, the Swarm cannot find the connection in its registry because it was never registered: `"Connection object not found in tracking"`.

## Expected Behavior

The connection lifecycle should have a single, unambiguous owner. Either:

- The `QUICListener` should **not** call `start()` directly and should defer the call to `Swarm.add_conn()`, or
- `Swarm.add_conn()` should detect that the connection is already started and only execute its own registration/tracking steps without calling `start()` again.

## Environment

- **py-libp2p version:** latest (PyPI)
- **Python:** 3.10 / 3.12
- **Platform:** Linux (Docker on AWS EC2), macOS
- **Related project:** [IPFS-Meshkit/py-ipfs-lite](https://github.com/IPFS-Meshkit/py-ipfs-lite)

sumanjeet0012:
## Summary

Fixes #1389

Resolves a race condition where `QUICConnection.start()` was being called twice — once by `QUICListener` (via `connect()`) and once by `Swarm.add_conn()`. The second call was silently returning early due to the `_started` guard, meaning the Swarm's lifecycle contract was effectively bypassed for every QUIC connection.

## Root Cause

`connect()` was delegating to `start()` for raw transport setup (initiating the handshake, setting `_established`, signalling `event_started`). By the time `Swarm.add_conn()` called `start()`, the `_started` flag caused an early return — leaving `event_started` already set by the transport rather than owned by the Swarm.

This violated the expected lifecycle: **`Swarm.add_conn()` should be the sole caller of `start()`** and the exclusive owner of `event_started`.

## Fix

Separate the two responsibilities within `QUICConnection`:

| Method | Before | After |
|---|---|---|
| `connect()` | Delegated raw setup to `start()`, also set `event_started` for initiators | Directly handles all raw transport setup (handshake, background tasks, peer verification). Sets `_started=True`. **Never** touches `event_started`. |
| `start()` | Performed raw setup (inbound) or was a no-op (outbound) | Pure Swarm lifecycle signal. Guards on `event_started.is_set()` (not `_started`). Sets `event_started.set()` — exclusively owned by `Swarm.add_conn()`. |

### Clean Lifecycle (after fix)

**Inbound connections:**
1. `QUICListener._promote_pending_connection()` calls `connection.connect(nursery)` — does full raw setup, sets `_started=True`. Does NOT touch `event_started`.
2. Listener calls `self._handler(connection)` → `Swarm.add_conn()`.
3. `add_conn` calls `manager.run_task(muxed_conn.start)` → `start()` now properly sets `event_started.set()`.
4. `add_conn` waits on `event_started.wait()` → satisfied correctly.

**Outbound connections:** identical pattern via `transport.dial()` → `add_conn()`.

## Changes

- **`libp2p/transport/quic/connection.py`**: refactored `connect()` and `start()` as described above.
- **`tests/core/transport/quic/test_connection.py`**: updated unit tests to reflect the new contract.

## Testing

- All **140 QUIC tests pass**.
- `make lint` passes (ruff, mypy, pyrefly).
- No changes required in `swarm.py`, `listener.py`, or `transport.py`.


sumanjeet0012:
@seetadev The PR is ready to merge.

seetadev:
@sumanjeet0012 : Great work on this PR.

I reviewed the changes and everything looks good to me. The implementation is clear, well-structured, and the accompanying changes/tests look solid. Thank you for taking the time to work through this and improve py-libp2p.

**Approved and merged, congratulations! 🚀**

CCing @acul71, @yashksaini-coder, @johannamoran and @mishmosh.

PR #1394, opened 2026-07-18
fix(quic): provide isolated QuicLogger per connection to prevent trace ownership errors

### LINKED ISSUE #1390: bug: Shared QuicLogger instance across concurrent QUIC connections causes QuicLoggerTrace ownership errors and connection crashes
## Summary

When multiple simultaneous inbound QUIC connections are active, the aioquic `QuicLogger` instance is shared across connection objects. When a `QuicLoggerTrace` created for one connection is dispatched by another connection's event loop, aioquic raises a `ValueError` because the trace does not belong to the logger instance. This causes the affected connection's event processing loop to crash and the connection to be abandoned.

## Observed Log Output

```
[INFO] [quic] [adafb3ef45c2757d] Negotiated protocol version 0x00000001 (VERSION_1)
[INFO] [quic] [adafb3ef45c2757d] ALPN negotiated protocol libp2p
[INFO] [quic] [c9fff623813f4555] Negotiated protocol version 0x00000001 (VERSION_1)
[INFO] [quic] [c9fff623813f4555] ALPN negotiated protocol libp2p
Error in event processing loop: QuicLoggerTrace does not belong to QuicLogger
Connection error: QuicLoggerTrace does not belong to QuicLogger
Connection object not found in tracking
Error in event processing loop: QuicLoggerTrace does not belong to QuicLogger
Connection error: QuicLoggerTrace does not belong to QuicLogger
```

Note that the error only occurs when **multiple connections are active concurrently** (visible by different connection IDs such as `adafb3ef` and `c9fff623` being negotiated at the same time).

## Root Cause Analysis

aioquic's `QuicLogger` is designed to be a **per-connection** object. Each `QuicConnection` (from aioquic) should be initialized with its own `QuicLogger` instance or with `None`. Each `QuicLogger` creates a `QuicLoggerTrace` internally, and that trace is bound exclusively to the logger that created it.

The error `"QuicLoggerTrace does not belong to QuicLogger"` is raised by aioquic at:

```python
# aioquic/quic/logger.py
def end_trace(self, trace: QuicLoggerTrace) -> None:
    if trace not in self._traces:
        raise ValueError("QuicLoggerTrace does not belong to QuicLogger")
```

This implies that py-libp2p is passing a **single shared `QuicLogger` instance** (or a shared `QuicLoggerTrace`) to multiple `NativeQUICConnection` (aioquic) objects. When concurrent connections share the same logger, their traces become entangled and cross-referenced, causing the ownership check to fail.

## Expected Behavior

Each QUIC connection should be initialized with either:
- Its own dedicated `QuicLogger` instance, or
- `quic_logger=None` (disabling per-connection logging)

The `QuicLogger` must never be shared between concurrent connections.

## Proposed Fix

In the py-libp2p QUIC transport, wherever a `NativeQUICConnection` (aioquic's `QuicConnection`) is created, ensure that a fresh `QuicLogger` instance is created per-connection:

```python
from aioquic.quic.logger import QuicLogger

# BAD: shared logger passed to multiple connections
quic_logger = QuicLogger()  # created once, reused
conn = QuicConnection(configuration=config, quic_logger=quic_logger)

# GOOD: fresh logger per connection
conn = QuicConnection(configuration=config, quic_logger=QuicLogger())
# or disable logging entirely:
conn = QuicConnection(configuration=config, quic_logger=None)
```

## Impact

This bug causes **concurrent inbound QUIC connections to crash each other**. In a real network scenario with multiple active peers (e.g., during DHT bootstrapping), many inbound connections arrive simultaneously. Each crash abandons a valid peer connection, severely degrading connectivity.

## Environment

- **py-libp2p version:** latest (PyPI)
- **aioquic version:** (as bundled with py-libp2p)
- **Python:** 3.10 / 3.12
- **Platform:** Linux (Docker on AWS EC2), macOS
- **Reproducible:** Yes — occurs reliably during DHT bootstrapping with multiple simultaneous inbound QUIC connections
- **Related project:** [IPFS-Meshkit/py-ipfs-lite](https://github.com/IPFS-Meshkit/py-ipfs-lite)

sumanjeet0012:
Fixes #1390

This PR resolves an issue where multiple concurrent QUIC connections would crash with `QuicLoggerTrace does not belong to QuicLogger` because they shared the same `QuicLogger` instance.

### Changes
* **Outbound Connections**: Stopped mutating the shared base configuration inside `dial()`. Instead, a fresh configuration is created using `create_client_config_from_base()` per connection.
* **Inbound Connections**: Properly check `enable_qlog` inside `_handle_new_connection()` and attach a fresh `QuicLogger` to the newly instantiated `server_config`.
* Provided an isolated `QuicLogger` instance for both inbound and outbound connections if QUIC logging is enabled.


sumanjeet0012:
@seetadev The PR is ready to merge.

Issue #1352, opened 2026-06-26
spike: aiortc ICE mux / STUN USERNAME exposure for webrtc-direct v2 listener


yashksaini-coder:
Sub-spike of #546.  Required before a spec-aligned `/webrtc-direct` **listener** can land.

## Background

WebRTC-Direct v1 ([libp2p/specs#672](https://github.com/libp2p/specs/issues/672)) and v2 ([libp2p/specs#715](https://github.com/libp2p/specs/pull/715)) are **no-signaling** transports.  The listener reconstructs the dialer's offer from the inbound STUN connectivity-check, reading the `USERNAME` attribute to recover ICE credentials and (for v2) the version-prefix dispatch (`libp2p+webrtc+v1/` vs `libp2p+webrtc+v2/`).

js-libp2p solved this at the native ICE layer ([murat-dogan/node-datachannel#420](https://github.com/murat-dogan/node-datachannel/pull/420) exposes `localUfrag` on `IceUdpMuxRequest`).  go-libp2p uses pion's ICE mux internals similarly.

py-libp2p uses aiortc.  Whether aiortc gives us equivalent hooks is unknown today.

## Goal

Determine — without committing to an implementation — whether aiortc exposes (or can be made to expose, via subclassing / monkey-patch / upstream PR):

1. **Inbound STUN parsing**: the `USERNAME` attribute of incoming connectivity checks before/during ICE negotiation, so the listener can branch on the version-prefix and reconstruct the dialer's offer.
2. **Server-side credential override**: setting local `ufrag` / `pwd` to the value parsed out of the STUN `USERNAME` before generating the answer (v2 step 6 in [libp2p/specs#715](https://github.com/libp2p/specs/pull/715)).
3. **ICE mux / shared UDP socket** across multiple concurrent inbound dials, so one listening UDP port can demux to per-dial `RTCPeerConnection`s without spinning up a port per peer.

## Deliverables

- A spike doc (in this issue, a comment thread, or a short markdown PR) covering:
  - Which of the 3 capabilities aiortc currently exposes (public API, private-but-accessible, or genuinely absent).
  - For absent ones: feasibility of a subclass / monkey-patch / upstream contribution and a rough effort estimate.
- A go / no-go recommendation: does py-libp2p WebRTC-Direct **listener** land on aiortc, or do we need a different ICE / DTLS backend (e.g. wrap a Rust / C library)?

## Why this isn't blocking #1309

[...]

PR #1309, opened 2026-04-13
feat(webrtc): add WebRTC transport scaffolding (spec compliance + tests)

yashksaini-coder:
## Summary

Adds **experimental** WebRTC transport scaffolding (`libp2p.transport.webrtc`) per the libp2p WebRTC and WebRTC Direct specs.  Enabled via the optional `libp2p[webrtc]` extra (`aiortc`) and the `enable_webrtc` opt-in.

This is a **v1 node-to-node** foundation, not a production-ready WebRTC stack.

Refs #546.

## What works (node-to-node)

- `WebRTCDirectTransport` (`/webrtc-direct`) and `WebRTCPrivateTransport` (`/webrtc` via Circuit Relay v2).
- Spec-compliant data-channel framing — uvarint length-prefixed protobuf, full FIN / FIN_ACK / STOP_SENDING / RESET state machine.
- In-band data channels for application streams; the Noise channel (id=0) remains pre-negotiated per spec.
- Signaling with bilateral `ICE_DONE` (libp2p/specs#585 fix).
- Noise XX prologue binding the handshake to DTLS certificate fingerprints.
- SDP builder with an isolated `_apply_ice_credentials()` seam for libp2p/specs#672.
- ECDSA P-256 cert generation with multihash / multibase fingerprint encoding; DTLS cert pinned to `RTCPeerConnection` via aiortc's name-mangled private slot so the advertised `/certhash/` matches the actual handshake.
- Trio ↔ asyncio bridge for `aiortc`.

## Out of scope for this PR

- **Browser interop is explicitly NOT supported.**  v1 SDP munging on the browser side is being phased out by Chrome's `WebRTC-NoSdpMangleUfrag` field trial; browser dial will land on v2 (libp2p/specs#715).
- Interop with go-libp2p / js-libp2p WebRTC Direct dialers.  Their listener reconstructs the offer from the inbound STUN `USERNAME`; ours currently uses an HTTP `POST /sdp` exchange, documented as a py-to-py temporary harness.
- Private `/webrtc` dial — only the listener / signaling skeleton lands here.
- Full inbound handler wiring (Noise + handler invocation) on the `WebRTCDirectListener` — pending an aiortc STUN `USERNAME` exposure spike (sub-issue of #546, to be filed).

## Recent additions (review push)

Following the [maintainer review](https://github.com/libp2p/py-libp2p/pull/1309#issuecomment-4778715862) the following critical-path fixes landed in 4 commits on top of the existing scaffolding:

[...]

Issue #546, opened 2025-03-27
Add Support for WebRTC in py-libp2p


seetadev:
### Description

Adding WebRTC support to py-libp2p would enhance its capabilities by enabling peer-to-peer connections in browsers and other environments where direct network access may be restricted. This feature would complement existing transports like QUIC and TCP, making py-libp2p more versatile for real-time communication and decentralized applications.

### Motivation

WebRTC is a critical transport layer for enabling efficient peer-to-peer communication, especially in environments where NAT traversal and real-time low-latency connections are essential. Many applications, such as decentralized chat systems, collaborative tools, and private-to-private networks, depend on WebRTC for seamless operation.

Implementing WebRTC in py-libp2p would also align with the existing capabilities of other libp2p implementations (e.g., js-libp2p and go-libp2p), further enhancing cross-platform interoperability.

### Current Implementation

1. Integration with a WebRTC Library: Utilize an existing Python WebRTC library such as [aiortc](https://github.com/aiortc/aiortc) to handle the WebRTC signaling, media, and data channels.

2. Signaling Protocol: Define a signaling protocol that integrates seamlessly with libp2p's transport layer.

3. NAT Traversal: Ensure proper handling of NAT traversal using Autonat, Circuit Relay, Hole Punching or using existing protocols like STUN and TURN.

4. Transport Implementation: Implement WebRTC as a libp2p transport, adhering to libp2p's transport interface.

5. Testing and Documentation: Provide unit tests, integration tests, and comprehensive documentation to ensure reliability and ease of use.

Acceptance Criteria:

A functional WebRTC transport is added to py-libp2p.

The WebRTC transport is interoperable with other libp2p implementations that support WebRTC.

Proper handling of signaling, NAT traversal, and peer discovery.

Clear documentation and examples for developers to utilize the feature.

Benefits:

Broader applicability of py-libp2p in browser-based and restricted network environments.

Improved interoperability with other libp2p implementations.

Expanded use cases for real-time communication and decentralized applications.

References:

[WebRTC Documentation](https://webrtc.org/)

[aiortc Python Library](https://github.com/aiortc/aiortc)

[js-libp2p WebRTC Example](https://github.com/libp2p/js-libp2p-example-webrtc-private-to-private)

[libp2p Transport Interface](https://github.com/libp2p/specs/tree/master/transport)

Potential Challenges:

Ensuring stable NAT traversal across different environments.

[...]

PR #1396, opened 2026-07-19
fix(webrtc): harden /sdp HTTP server against memory-amplification DoS

### LINKED ISSUE #1354: harden: bound Content-Length and header reads on WebRTC Direct HTTP /sdp dev harness
Follow-up from [#1309 AI Review #4](https://github.com/libp2p/py-libp2p/pull/1309#issuecomment-4809236557) (Major finding #2).  Sub-issue of #546.

## TL;DR

The HTTP `POST /sdp` signaling server used by `WebRTCDirectListener` reads an attacker-controlled `Content-Length` and the request headers without bounding either.  A remote client can:

- send a `Content-Length` of any 64-bit integer and stream bytes slowly (within the 15s timeout) to consume server memory, or
- send an arbitrary number of header lines (each shorter than the timeout) without ever reaching the blank line, holding the connection indefinitely.

The path is documented as a **py-to-py dev harness** (out-of-scope for go/js interop, slated for replacement by the STUN-based listener in #1352), but it is still reachable over the network whenever a listener is started.  The fix is small and self-contained — the path can keep working until #1352 lands.

## Background

`libp2p/transport/webrtc/_aiortc_helpers.py:run_signaling_server` starts a raw-asyncio HTTP server on the same TCP port as the WebRTC Direct UDP endpoint and accepts SDP offers via `POST /sdp`.  Per [PR #1309's framing](https://github.com/libp2p/py-libp2p/pull/1309#issuecomment-4778715862) and the experimental newsfragment, this is a temporary py-to-py dev harness; the spec-compliant listener reconstructs the offer from inbound STUN `USERNAME` (#1352).

The vulnerable section, current `webrtc` branch HEAD (`ed645324`):

```python
# libp2p/transport/webrtc/_aiortc_helpers.py:382-397
while True:
    line = await asyncio.wait_for(
        reader.readline(), timeout=_SDP_HTTP_TIMEOUT
    )
    if line in (b"\r\n", b"\n", b""):
        break
    key, _, value = line.decode().partition(":")
    headers[key.strip().lower()] = value.strip()

content_length = int(headers.get("content-length", "0"))
body = b""
if content_length > 0:
    body = await asyncio.wait_for(
        reader.readexactly(content_length),
        timeout=_SDP_HTTP_TIMEOUT,
    )
```

`reader.readexactly(content_length)` will allocate up to `content_length` bytes of buffer (Python `asyncio.StreamReader` grows its buffer to accommodate `readexactly`).  With `content_length = 2**31 - 1` the server attempts to grow a ~2 GiB buffer per connection.

## Threat model

| Vector | Cost to attacker | Cost to server |
|---|---|---|
| Single connection, `Content-Length: 2147483647`, slow trickle within 15 s | one TCP connection | up to ~2 GiB peak memory |
| Many concurrent connections, each as above | N TCP connections | N × ~2 GiB |
| Single connection, unbounded header lines (each ≤ 15 s apart) | one TCP connection | one stuck async task indefinitely |
| Many concurrent, each holding header loop | N connections | N stuck tasks |

The 15s per-`readline` / `readexactly` timeout is what already saves us from a full slowloris; it does **not** save us from memory pressure during the read.

## Suggested approach (≈ 30 lines)

Three bounded constants + early rejection.  No new dependencies, no protocol change.

```python
# libp2p/transport/webrtc/_aiortc_helpers.py — near _SDP_HTTP_TIMEOUT

# SDP offers in this harness are typically ~1-4 KiB.  32 KiB leaves
# generous headroom while making a memory-amplification DoS uninteresting.
_MAX_SDP_BODY_SIZE = 32 * 1024
# Bounded HTTP header section.  Defends the header-parse loop against
# clients that never terminate the headers with a blank line.
_MAX_HEADER_LINES = 64
_MAX_HEADER_BYTES = 8 * 1024
```

In `_handle`:

1. Cap the header-parse loop at `_MAX_HEADER_LINES` and accumulate header byte size; close with `400 Bad Request` if either is exceeded.
2. Validate `Content-Length`: if absent / non-integer / negative / `> _MAX_SDP_BODY_SIZE`, close with `400 Bad Request` (or `413 Payload Too Large` for the size case).
3. Only after both pass, call `reader.readexactly(content_length)`.

The existing `_SDP_HTTP_TIMEOUT` continues to bound the per-line / per-body wait time.

## Acceptance criteria

- [ ] `Content-Length` exceeding `_MAX_SDP_BODY_SIZE` returns `413` and closes the connection without allocating the body buffer.
- [ ] Non-integer / negative `Content-Length` returns `400`.
- [ ] Header loop terminates with `400` after `_MAX_HEADER_LINES` lines or `_MAX_HEADER_BYTES` accumulated bytes (whichever first).
- [ ] Existing happy-path test (`tests/core/transport/webrtc/test_webrtc_direct_loopback.py` etc.) still passes.
- [ ] One new regression test that asserts a request with `Content-Length: 999999999` gets a `413` and the server's process RSS doesn't grow by >1 MiB during the rejection.  (RSS check can be a soft check via `resource.getrusage` if available; otherwise just assert the response code and that the body was never read.)

## Why not a bigger redesign

Per-IP rate-limiting, an `aiohttp`-based server, or TLS are all out of scope here — the harness is on a deprecation path and the listener it serves is **explicitly not interop-compatible** with js / go-libp2p (those reconstruct from STUN `USERNAME`, see #1352).  The above three changes raise the bar from "trivially DoSable in a LAN" to "needs a real attacker model for an experimental, opt-in (`libp2p[webrtc]` + `enable_webrtc`) transport" — the right hardening level for the time window this code will exist.

## Refs

- #546 (parent feature)
- #1309 (the PR that lands this code path)
- #1352 (STUN-based listener spike that ultimately replaces `/sdp`)
- [acul71's AI Review #4 Major #2](https://github.com/libp2p/py-libp2p/pull/1309#issuecomment-4809236557)

yashksaini-coder:
## Summary

Closes #1354. _Recreates #1380, which was auto-closed when the source fork was recreated — same diff, no code changes._

The `POST /sdp` HTTP handler in `_aiortc_helpers.py` (`run_signaling_server`) had two unbounded read paths that let a remote attacker consume server memory or hold async tasks indefinitely:

- **Unbounded header loop** — `while True:` with only a per-readline 15s timeout. A slow attacker could send one header line every 14s and hold the connection forever.
- **Unbounded body allocation** — `reader.readexactly(content_length)` where `content_length` is attacker-controlled. `Content-Length: 2147483647` grows a ~2 GiB asyncio buffer per connection before the timeout fires.

## Fix

Three constants + early rejection before any body buffer is allocated:

| Constant | Value | Behaviour on violation |
|---|---|---|
| `_MAX_SDP_BODY_SIZE` | 32 KiB | `413 Payload Too Large` |
| `_MAX_HEADER_LINES` | 64 | `400 Bad Request` |
| `_MAX_HEADER_BYTES` | 8 KiB | `400 Bad Request` |

Non-integer and negative `Content-Length` values also return `400`. The existing `_SDP_HTTP_TIMEOUT` (15s per readline/readexactly) is unchanged. Uses Python's `for/else` idiom: the `else` fires when the header loop exhausts `_MAX_HEADER_LINES` reads without finding the blank-line terminator.

## Test plan

- [x] `Content-Length: 999999999` → `413`, body buffer not allocated
- [x] `Content-Length: notanumber` → `400`
- [x] `Content-Length: -1` → `400`
- [x] `Content-Length` exactly at limit → `200` (happy path still works)
- [x] `_MAX_HEADER_LINES + 1` non-blank header lines → `400`
- [x] Single header line > `_MAX_HEADER_BYTES` → `400`
- [x] Valid SDP request → `200`, `on_offer` callback receives correct body
- [x] All WebRTC tests pass (new `test_aiortc_helpers.py` regression suite)

## Scope note

This harness is documented as a temporary py-to-py scaffold (not interop-compatible with go/js libp2p). The fix raises the bar from "trivially DoSable over LAN" to "needs a real attacker model" while the STUN-based listener (#1352) is in spike. Refs #546, #1309, #1352.


acul71 (approved):
## AI-assisted maintainer review

**Verdict:** Approve — ready to merge once remaining CI is green.

Closes #1354 correctly: bounds `/sdp` header parsing and rejects oversized / invalid `Content-Length` **before** `readexactly`, closing the memory-amplification DoS on the temporary WebRTC Direct HTTP harness.

### Strengths
- Matches the #1354 design (32 KiB body → 413; bad CL → 400; header line/byte caps → 400)
- Reject-before-allocate ordering is correct
- Solid regression suite in `test_aiortc_helpers.py` (7 tests, including soft RSS check on 413)
- `asyncio.run` test wrappers are appropriate here — this code lives on the aiortc asyncio bridge, not the Trio host path
- Process complete: `Closes #1354` + `newsfragments/1354.bugfix.rst` (trailing newline OK)
- Branch clean vs `main` (0 behind, no merge conflicts)
- Local validation: `make lint` / `typecheck` / `test` (3234 passed) / `linux-docs` all green

### Minor nits (non-blocking)
1. **`_MAX_HEADER_LINES` off-by-one:** `for _ in range(64)` means at most **63** non-blank headers are accepted (blank terminator must fall inside those 64 reads). Defense still works; consider documenting or using `range(_MAX_HEADER_LINES + 1)` so the constant matches behavior.
2. **Request-line `readline()`** still relies on asyncio’s default 64 KiB `StreamReader` limit (`LimitOverrunError` → generic 500). Residual / fine for a temporary harness; optional follow-up for symmetry with header/body bounds.
3. Optional: small `_http_error(...)` helper to dedupe the four 400/413 write blocks; narrow or drop the module-level `# pyrefly: ignore` in the new test file.

### Security
Positive impact — closes the high-severity unbounded buffer path. Residual connection-count / slowloris risk is explicitly out of scope per #1354 while this harness awaits the STUN listener (#1352).

### Questions (optional)
1. Is the effective 63-header cap intentional?
2. Last intentional touch on this harness before #1352, or worth request-line caps too?

**Quality:** Excellent · **Merge readiness:** Ready (nits only)

yashksaini-coder:
Thanks for the approve! Pushed `2958f6aa` for nit #1 and answering your questions.

**Nit #1 — off-by-one.** Fixed by making the constant honest: the loop now reads `range(_MAX_HEADER_LINES + 1)` so a full `_MAX_HEADER_LINES` (64) header lines can be followed by the blank-line terminator on the last read. Effective accepted-header cap is now 64, matching the name. Over-limit requests still take the `for/else` reject path (`test_too_many_header_lines_returns_400` sends `_MAX_HEADER_LINES + 1` and still gets a 400).

**Your questions:**
1. No — the 63-cap was an unintended off-by-one, now fixed.
2. Yes, this is meant as the last hardening touch on the temporary harness before #1352 lands the STUN listener. Request-line caps (nit #2) and the `_http_error` dedupe (nit #3) are noted but left out of scope for the #1354 DoS fix — happy to file a small follow-up for them if you'd prefer them in before #1352.

Leaving nits #2/#3 as-is unless you want them here. CI re-running.

Issue #1384, opened 2026-07-13
enhancement(kad-dht): add disjoint lookup paths to find_closest_peers_network


yashksaini-coder:
## Problem

`find_closest_peers_network` in `libp2p/kad_dht/peer_routing.py` runs a single iterative lookup with one shared candidate pool and one shared queried-peers set. Two weaknesses follow:

1. **Deterministic candidate selection**: each round always picks the ALPHA XOR-closest unqueried peers. An attacker who places Sybil peers at the closest XOR positions gets queried deterministically every round, and can steer the entire lookup path from the first hop.

2. **Single path**: all ALPHA parallel queries share one candidate pool. If an attacker controls a peer early in the path, that peer's response shapes the entire subsequent search.

rust-libp2p has shipped disjoint query paths since v0.20 as `disjoint_query_paths` (off by default, path count = ALPHA/parallelism).

**Note on scope**: disjoint paths do NOT protect against an attacker who *already occupies the closest-K bucket* for the target key — all paths converge on the same XOR terminus regardless. That problem is addressed separately by subnet diversity (#1383). This issue is about hardening *intermediate hops* against path-level steering.

### Part 1 — Randomized candidate selection per round (smaller, ships now)

Instead of always picking the strictly first ALPHA unqueried peers:

```python
# before
peers_to_query = [p for p in closest_peers if p not in queried_peers][:ALPHA]

# after
candidates = [p for p in closest_peers if p not in queried_peers][:count]
peers_to_query = random.sample(candidates, min(ALPHA, len(candidates)))
```

Draws randomly from the closest-K window each round. Convergence is preserved (all candidates are already within the K closest known peers); attacker loses guaranteed first-query slot. Already landed in local branch.

### Part 2 — True disjoint paths (larger, separate PR)

1. Partition the initial closest-peer candidate set into `d` disjoint subsets (default `d = 1`, preserving current behavior).
2. Run `d` parallel, isolated iterative lookups — separate candidate list and `queried_peers` tracker per path.
3. No peer crosses between paths mid-lookup.
4. Merge results by XOR distance at the end.
5. Expose `disjoint_paths: int = DISJOINT_LOOKUP_PATHS` on `KadDHT.__init__`.

## Scope

**Part 1** (ready):
- `libp2p/kad_dht/peer_routing.py` — one-line change to candidate selection
- `libp2p/kad_dht/common.py` — `DISJOINT_LOOKUP_PATHS = 1` placeholder constant

[...]

Issue #57, opened 2018-11-12
[Stretch] Network attack simulation


robzajac:
The Ethereum networking layer has been attacked several times. Py-libp2p and more generally libp2p would benefit from protection against [common attacks on peer to peer networks](http://www.dcs.gla.ac.uk/~sadekf/P2PSecTaxonomy.pdf). Py-libp2p should simulate attacks on topologies constructed through this library and assess security against these attacks.

zixuanzh:
We can start from eclipse attack simulation as described in this [paper] (https://eprint.iacr.org/2018/236.pdf). It is done through poisoning the DHT which is the primary source of structure in the network.

mgoelzer:
@lanzafame Yes, I agree.  We are building a test cluster project this quarter (reminder to self:  insert issue link here once spec issue is available), then we will definite interop tests for common modules like DHT, PubSub, etc.  I made https://github.com/libp2p/interop/issues/3 to track the need to bring github.com/libp2p/interop up to date or move that content into another repo.

acul71:
Closing as obsolete/completed for tracker purposes. Network attack simulation coverage exists in-tree under `tests/examples/attack_simulation/`.

PR #1399, opened 2026-07-21
feat(kad_dht): enforce IP subnet diversity in k-buckets (#1383)

### LINKED ISSUE #1383: security(kad-dht): add IP/subnet diversity enforcement to KBucket
## Problem

py-libp2p's k-buckets accept peers with no constraint on network origin. An attacker can grind many valid peer IDs (peer ID = multihash(pubkey), free to mint) that land closest to a target key T in XOR space, then run them all from a small number of IPs/subnets, and fully occupy the closest-K slots for T in every honest node's routing table.

Because the messages from these peers are authentically signed, record-level defenses (#1338) don't catch this — the attack lives in the routing layer. Background: #57, https://dev.to/yashksaini/hardening-kademlia-dht-the-eclipse-attack-that-record-signing-doesnt-stop-2k79

## Proposal

Before accepting a new peer in `KBucket.add_peer`, check whether its routable subnet already holds `MAX_PEERS_PER_SUBNET` entries. If saturated, reject the new peer.

- IPv4 default prefix: `/24` (stricter than go-libp2p's `/16` — see open questions)
- IPv6 default prefix: `/48` — `/64` is too loose; a single `/56` delegation grants 256 distinct `/64`s
- Default `MAX_PEERS_PER_SUBNET = 2`
- Loopback (`127.x.x.x`, `::1`) and link-local addresses are **exempt** so CI, local testnets, and relay-only peers are unaffected

This forces an eclipse attacker to control diverse IP space, not just diverse keypairs from one VPS — a real operational/financial barrier instead of a free CPU barrier.

## Implementation sketch (already in local branch)

```python
# common.py
MAX_PEERS_PER_SUBNET = 2
SUBNET_PREFIX_LEN_V4 = 24
SUBNET_PREFIX_LEN_V6 = 48

# routing_table.py — module-level helper
def _subnet_key(addr: Multiaddr) -> str | None:
    for proto, net_cls, prefix_len in (
        ("ip4", IPv4Network, SUBNET_PREFIX_LEN_V4),
        ("ip6", IPv6Network, SUBNET_PREFIX_LEN_V6),
    ):
        try:
            ip = ip_address(addr.value_for_protocol(proto))
            if ip.is_loopback or ip.is_link_local:
                return None
            return str(net_cls(f"{ip}/{prefix_len}", strict=False))
        except Exception:
            continue
    return None

# KBucket.add_peer — before the bucket-has-space check
for addr in peer_info.addrs:
    subnet = _subnet_key(addr)
    if subnet and self._peers_in_subnet(subnet) >= MAX_PEERS_PER_SUBNET:
        logger.debug("Subnet %s at capacity, rejecting peer %s", subnet, peer_id)
        return False
```

Two tests cover: (1) rejection when subnet is saturated, (2) loopback peers are exempt regardless of count.

## Scope

- `libp2p/kad_dht/routing_table.py` — `KBucket.add_peer` + `_subnet_key` helper + `_peers_in_subnet` method
- `libp2p/kad_dht/common.py` — three new constants
- `tests/core/kad_dht/test_unit_routing_table.py` — two new tests

## Open questions for maintainers

1. **IPv4 prefix length**: `/24` (current proposal, stricter) vs `/16` (go-libp2p default). `/24` blocks entire cloud subnets that a single operator could control; `/16` is less disruptive but weaker. Which tradeoff fits py-libp2p's deployment targets?
2. **Reject vs replace-if-worse**: current impl rejects outright. Alternative: if subnet is saturated, evict the *youngest* existing peer in that subnet (prefer longer-lived/established peers). More complex — is the extra churn management worth it?
3. **Config flag**: loopback/link-local already exempt. Any other setups (NAT behind shared IPs, enterprise proxies) that need `MAX_PEERS_PER_SUBNET=0` to opt out entirely?

## Non-goals

Not proposing crypto-puzzle peer IDs or reputation scoring — both require spec-level/ecosystem changes. This is scoped to a self-contained, single-library change with no protocol impact.

/cc @acul71 @sumanjeet0012 @asmit27rai — you've been most active on the kad_dht path recently

yashksaini-coder:
## Summary

Closes #1383.

py-libp2p's k-buckets accept peers with no constraint on network origin. An attacker can grind many valid peer IDs (peer ID = multihash(pubkey), free to mint) that land closest to a target key in XOR space, then run them all from a small number of IPs/subnets and occupy the closest-K slots for that key in every honest node's routing table. Because the messages are authentically signed, record-level defenses don't catch this — the attack lives in the routing layer.

This PR enforces **IP/subnet diversity** in `KBucket.add_peer`: before admitting a new peer, it is rejected if its globally-routable subnet already holds `MAX_PEERS_PER_SUBNET` peers in that bucket.

## Design

| Constant | Default | Meaning |
|---|---|---|
| `MAX_PEERS_PER_SUBNET` | `2` | max peers sharing a subnet per bucket; **`<= 0` disables the check** |
| `SUBNET_PREFIX_LEN_V4` | `24` | IPv4 grouping prefix |
| `SUBNET_PREFIX_LEN_V6` | `48` | IPv6 grouping prefix |

- **Only globally-routable addresses are grouped** (`ipaddress.is_global`). Loopback, private (RFC1918/ULA), CGNAT (100.64.0.0/10), link-local, and documentation ranges are all exempt — CI and local testnets are unaffected. Using `is_global` (rather than enumerating private ranges) keeps behaviour stable across CPython versions.
- **DNS-named and relayed (`p2p-circuit`) peers are exempt.** A circuit address exposes the *relay's* IP, not the peer's, so it is skipped to avoid grouping distinct peers behind a shared relay. DNS names are not resolved in the routing hot path.
- **Multi-homed peers** are grouped by their first globally-routable address.
- **Reject-only, no eviction** — matches go-libp2p's `TryAddPeer` and avoids a churn/DoS vector where an attacker forces out established peers.
- `_should_split_bucket` now only splits genuinely full buckets, so a subnet rejection (`add_peer` returning `False` on a *non-full* bucket) can no longer trigger a spurious bucket split.

No protocol or wire-format change. Self-contained routing-layer change.

## Test plan

- [x] `uv run python -m pytest tests/core/kad_dht/test_unit_routing_table.py` → **23 passed** (16 existing + 7 new)
- [x] New tests cover: subnet saturation rejection, loopback/private exemption, opt-out flag, distinct-subnet acceptance, IPv6 `/48` grouping, multi-homed grouping-by-first-global, and the `_subnet_key` helper (incl. DNS + relay exemption)
- [x] `ruff check` / `ruff format --check` clean; `mypy` clean on changed modules

## Open questions for maintainers

1. **IPv4 prefix**: this ships `/24` (stricter than go-libp2p's `/16`). `/24` matches the realistic granularity an attacker rents; `/16` over-blocks large shared ISP blocks less aggressively. Prefer `/24`, or widen to `/16`?
2. **Table-wide cap**: go-libp2p also has a table-wide `MaxPeersPerIPGroup=3` (needs RoutingTable-level counter bookkeeping). Deferred here as a follow-up — acceptable, or wanted now?
3. **Opt-out surface**: disable is a module-level constant (`MAX_PEERS_PER_SUBNET <= 0`) rather than a `RoutingTable`/`KadDHT` constructor kwarg. Want a threaded kwarg instead/in addition?
4. **Multi-homed posture**: grouped by *first global address* (lenient), the opposite of go-libp2p's reject-if-any-group-full. Confirm this posture.

## Non-goals

No crypto-puzzle peer IDs or reputation scoring (spec/ecosystem-level). Path-steering hardening is tracked separately in #1384.

## Notes

- Background: #57, and the eclipse-attack write-up linked from the issue.
- IPv6 uses a fixed `/48` (site-allocation boundary); go-libp2p's ASN-based IPv6 grouping is intentionally not ported (it needs a bundled ASN dataset py-libp2p doesn't ship).


yashksaini-coder:
> **📖 Context / reasoning:** This PR is Layer 1 of a two-part eclipse-hardening effort for the kad-dht. If you'd like the full thought process behind the approach — why subnet diversity here, and why the related #1384 is moving to disjoint lookup paths rather than randomized candidate selection — it's written up in plain language here: **libp2p/py-libp2p#1400** (discussion). Happy to discuss any of the open questions above either here or there.

acul71:
@seetadev 
Thanks @yashksaini-coder for the thorough write-up, the tests, and the context in [#1400](https://github.com/libp2p/py-libp2p/discussions/1400). I went through the open questions against go-libp2p's kad-dht / kbucket implementation. Here are my maintainer decisions and the reasoning.

---

## How go-libp2p does it (reference)

Diversity lives in **`go-libp2p-kbucket/peerdiversity`**, wired into kad-dht via `RoutingTablePeerDiversityFilter` — not in core `go-libp2p` itself.

| Aspect | go-libp2p (when enabled) | This PR |
|--------|--------------------------|---------|
| **Scope** | Routing-table filter at add time | Per **k-bucket** in `KBucket.add_peer` |
| **Per-group limit** | `maxPerCpl = 2` (per CPL) | `MAX_PEERS_PER_SUBNET = 2` (per bucket) |
| **Table-wide limit** | `maxForTable = 3` | Not implemented |
| **IPv4 grouping** | `/16` (legacy Class A → `/8`) | `/24` |
| **IPv6 grouping** | ASN-based | Fixed `/48` |
| **Multi-address peer** | **All** addresses checked | **First** global only |
| **Private / no addr** | Still grouped / rejected if no addrs | Exempt via `is_global` |
| **Configuration** | DHT constructor option | Module constants |
| **Default on?** | Dual DHT WAN only; plain DHT opt-in | Always on (unless `<= 0`) |

Defaults in go dual DHT:

```go
// go-libp2p-kad-dht/dual/dual.go
maxPrefixCountPerCpl = 2  // per CPL (common prefix length with local node)
maxPrefixCount       = 3  // table-wide per IP group
```

---

## 1. IPv4 `/24` or `/16`?

**Decision: `/24` LGTM for py-libp2p.**

go's IPv4 grouping (`go-libp2p-kbucket/peerdiversity/filter.go`):

```go
// IPv4: legacy Class A → /8, otherwise → /16
rs, _ := f.legacyCidrs.ContainingNetworks(ip)
if len(rs) != 0 {
    key := ip.Mask(net.IPv4Mask(255, 0, 0, 0)).String()
    return PeerIPGroupKey(key)
}
// otherwise -> /16 prefix
key := ip.Mask(net.IPv4Mask(255, 255, 0, 0)).String()
return PeerIPGroupKey(key)
```

This PR uses `/24` via `SUBNET_PREFIX_LEN_V4 = 24`. That matches realistic attacker economics (a rented cloud subnet is usually a `/24`, not a `/16`). go's `/16` trades some security for fewer false rejections on large ISP blocks.

Please **document the divergence from go** in the PR or module comments. Legacy Class A `/8` special-casing (go handles `12.0.0.0/8`, `17.0.0.0/8`, etc.) can be a follow-up if we care about parity — not a blocker for this PR.

---

## 2. Table-wide cap — now or follow-up?

**Decision: merge per-bucket now; please open a follow-up for the table-wide cap.**

go has **two** limits (`go-libp2p-kad-dht/rt_diversity_filter.go`):

```go
func (r *rtPeerIPGroupFilter) Allow(g peerdiversity.PeerGroupInfo) bool {
    key := g.IPGroupKey
    cpl := g.Cpl

    // TABLE-WIDE: max 3 peers from same IP group anywhere in routing table
    if r.tableIpGroupCount[key] >= r.maxForTable {
        return false
    }

    // PER-CPL: max 2 peers from same IP group at each common-prefix-length
    c, ok := r.cplIpGroupCount[cpl]
    allow := !ok || c[key] < r.maxPerCpl
    return allow
}
```

Per-bucket limits alone still let an attacker grind peer IDs into **many buckets** and place **2 peers per bucket** from the same `/24`. go's `maxForTable = 3` caps that globally across the whole routing table. That's the bigger gap vs go — more important than the `/24` vs `/16` debate.

Happy to merge the per-bucket enforcement here; please file a follow-up for table-wide `maxForTable=3` at `RoutingTable` level.

---

## 3. Module constants vs constructor kwarg?

**Decision: module constants in `common.py` are fine for v1.**

go exposes runtime configuration:

```go
// go-libp2p-kad-dht/dht_options.go
func RoutingTablePeerDiversityFilter(pg peerdiversity.PeerIPGroupFilter) Option

// Usage (dual DHT WAN):
dht.RoutingTablePeerDiversityFilter(
    dht.NewRTPeerDiversityFilter(h, maxPrefixCountPerCpl, maxPrefixCount),
)
```

This PR:

```python
MAX_PEERS_PER_SUBNET = 2
SUBNET_PREFIX_LEN_V4 = 24
SUBNET_PREFIX_LEN_V6 = 48
# Set MAX_PEERS_PER_SUBNET <= 0 to disable
```

That matches our existing pattern (`BUCKET_SIZE`, etc. in `common.py`). A runtime kwarg on `KadDHT` / `RoutingTable` would be nice for production tuning without editing source — fine as a small follow-up, not a merge blocker.

---

## 4. Multi-homed — first global or all addresses?

**Decision: first globally-routable address is acceptable for this PR.**

go checks **every** address — reject if **any** group is full (`go-libp2p-kbucket/peerdiversity/filter.go`):

```go
peerGroups := make([]PeerGroupInfo, 0, len(addrs))
for _, a := range addrs {
    ip, err := manet.ToIP(a)
    // ...
    key := f.ipGroupKey(ip)
    group := PeerGroupInfo{Id: p, Cpl: cpl, IPGroupKey: key}

    if !f.pgm.Allow(group) {
        return false  // ANY saturated group → reject entire peer
    }
    peerGroups = append(peerGroups, group)
}
```

This PR (`routing_table.py`) — first global wins:

```python
def _subnet_key(peer_info: PeerInfo) -> str | None:
    for addr in peer_info.addrs:
        if "p2p-circuit" in str(addr):
            continue
        for proto, prefix_len in (("ip4", SUBNET_PREFIX_LEN_V4), ("ip6", SUBNET_PREFIX_LEN_V6)):
            # ...
            if not ip.is_global:
                continue
            return str(ip_network(f"{ip}/{prefix_len}", strict=False))
    return None
```

First-global is simpler and avoids false rejects for legit multi-homed peers. Please **document** that go checks all addresses. A stricter all-global-addresses check is a reasonable follow-up once address ordering is well-defined.

---

## Other notes (not blockers, but worth acknowledging)

**Always-on vs opt-in:** go only enables diversity on dual DHT WAN by default; plain kad-dht is opt-in. Making this always-on in py-libp2p (unless `MAX_PEERS_PER_SUBNET <= 0`) is a reasonable security default.

**Peers with no addresses:** go rejects them:

```go
addrs := f.pgm.PeerAddresses(p)
if len(addrs) == 0 {
    return false
}
```

This PR exempts them (`_subnet_key` → `None`). Good for CI/local testnets; slightly looser for production — acceptable for v1 given our exemption story.

**IPv6:** go uses ASN (`asnutil.AsnForIPv6`); this PR uses fixed `/48` without bundling an ASN dataset — pragmatic and already well documented.

---

## Before merge (small fixes)

1. Remove unused `import multihash` in `routing_table.py` (ruff flags it).
2. Rebase onto latest `main` (branch is quite far behind).
3. Add a regression test for `_should_split_bucket` when subnet rejection happens on a **non-full** bucket (the fix you describe in the PR description).
4. Optional cleanup: unreachable `return False` at end of `KBucket.add_peer`; clearer debug log when rejection is subnet vs bucket-full.

---

**Summary:** Approve the core design — `/24`, first-global, module constants. Follow-up issues for table-wide cap and optional runtime config. Thanks again for picking up #1383 and for the clear design write-up.

yashksaini-coder:
Thanks for going through the go side properly, the comparison table made this easy to act on.

Pushed the fixes:
- dropped the unused `multihash` import
- removed the dead `return False` at the tail of `add_peer` (both branches already return)
- added a regression test for the non-full-bucket case — it splits without the guard and doesn't with it, so it actually pins the fix
- documented the `/24`-vs-`/16` and first-global-vs-all-addresses divergences right where the constants and `_subnet_key` live

Branch is already on latest main.

And filed the two you flagged:
- table-wide cap (`maxForTable`): #1421
- runtime-configurable limits: #1422

Agree the table-wide cap is the bigger gap of the two — I'll take that one next once this lands.

yashksaini-coder:
@acul71 this is ready for another look whenever you have time — all the before-merge items from your review are pushed (unused import, split-guard test, the go-divergence docs) and CI is green. No rush, just flagging it's unblocked on my end.

acul71 (approved):
## Approve

LGTM. Approving the IP/subnet diversity enforcement for k-buckets (#1383).

### What’s in this PR
- Per-bucket `MAX_PEERS_PER_SUBNET` gate in `KBucket.add_peer` (`/24` IPv4, `/48` IPv6; globally-routable only)
- Exemptions for private/loopback/CGNAT/link-local/DNS/`p2p-circuit`
- `_should_split_bucket` guard so subnet rejection on a non-full bucket cannot spuriously split
- Unit coverage (saturation, exemptions, opt-out, IPv6, multi-homed, split-guard regression)
- go-libp2p divergence documented; follow-ups filed (#1421 table-wide cap, #1422 runtime config)

### Added since last look (minor cleanup)
Commit `fix(kad_dht): clarify subnet-vs-full drop log and add_peer docs`:
- `RoutingTable.add_peer` failure log now distinguishes **subnet at capacity** from **bucket full / cannot split**
- `KBucket.add_peer` docstring updated so `False` covers fullness *and* subnet-diversity rejection

No remaining blockers from my side. Thanks @yashksaini-coder.

PR #1402, opened 2026-07-22
test(pubsub): de-flake test_gossip_gate_filters_peers (#1401)

### LINKED ISSUE #1401: flaky: test_gossip_gate_filters_peers races on subscription propagation (fixed trio.sleep)
## Summary

`tests/core/pubsub/test_gossipsub_v1_1_score_gates.py::TestScoreGates::test_gossip_gate_filters_peers` fails intermittently on CI (observed on `windows (3.12, core)`), while passing on all Linux runners and on Windows 3.11/3.13.

```
tests/core/pubsub/test_gossipsub_v1_1_score_gates.py:177
    assert peer1_id in peers_to_send
AssertionError: assert <ID ...> in []
```

## Root cause

The test subscribes all peers to the topic, then waits with a **fixed `await trio.sleep(0.2)`** before calling `gsub0._get_peers_to_send(...)`.

`_get_peers_to_send` (`libp2p/pubsub/gossipsub.py`) begins each topic iteration with:

```python
if topic not in self.pubsub.peer_topics:
    continue
```

`pubsub.peer_topics[topic]` is only populated when the node **receives** a SUBSCRIBE RPC from a connected peer (`handle_subscription`). On a slow/loaded Windows runner the subscription hasn't propagated to `gsub0` within 0.2s, so the topic key is absent, the loop `continue`s, `send_to` stays empty, and `peers_to_send == []` → the assertion fails.

It's a fixed-sleep race, not a logic bug — all the mesh/score state in this test is set manually.

## Fix

Replace the fixed sleep with the existing event-based helper `Pubsub.wait_for_subscription(peer_id, topic)`, whose docstring explicitly says *"Use this instead of arbitrary trio.sleep() calls to avoid race conditions."*

Scope is intentionally limited to this one test: sibling tests in the file share the fixed-sleep pattern but at least one (`test_publish_gate_blocks_low_scoring_peers`) depends on heartbeat-driven mesh formation, where a subscription-only wait would under-wait. Broadening the de-flake safely would need per-test analysis and is out of scope here.

yashksaini-coder:
## Summary

Closes #1401.

`test_gossip_gate_filters_peers` fails intermittently on CI (seen on `windows (3.12, core)`; green on all Linux runners and Windows 3.11/3.13):

```
tests/core/pubsub/test_gossipsub_v1_1_score_gates.py:177
    assert peer1_id in peers_to_send
AssertionError: assert <ID ...> in []
```

## Root cause

The test subscribes all peers, then waits a **fixed `await trio.sleep(0.2)`** before calling `gsub0._get_peers_to_send(...)`. That helper skips a topic until it appears in `pubsub.peer_topics`:

```python
if topic not in self.pubsub.peer_topics:
    continue
```

`peer_topics[topic]` is only populated when the node **receives** a SUBSCRIBE RPC (`handle_subscription`). On a slow runner the subscription hasn't propagated to `gsub0` within 0.2s → topic key absent → `continue` → `send_to` empty → `peers_to_send == []`. A fixed-sleep race, not a logic bug (all mesh/score state in this test is set manually).

## Fix

Replace the fixed sleep with the existing event-based helper `Pubsub.wait_for_subscription(peer_id, topic)` — whose own docstring says *"Use this instead of arbitrary trio.sleep() calls to avoid race conditions."* Deterministic and non-hanging (5s timeout, returns instantly once the subscription is registered).

```diff
-await trio.sleep(0.2)
+await pubsubs[0].wait_for_subscription(hosts[1].get_id(), topic)
+await pubsubs[0].wait_for_subscription(hosts[2].get_id(), topic)
```

## Scope

Intentionally limited to this one test. Sibling tests in the file share the fixed-sleep pattern, but at least one (`test_publish_gate_blocks_low_scoring_peers`) depends on **heartbeat-driven mesh formation**, where a subscription-only wait would *under-wait* and introduce a different flake. A broader de-flake needs per-test analysis and is out of scope here.

## Test plan

- [x] `test_gossip_gate_filters_peers` run 5× locally → stable pass
- [x] Full file `test_gossipsub_v1_1_score_gates.py` → 11 passed, no regression
- [x] `ruff check` / `ruff format --check` clean
- [x] `newsfragments/1401.internal.rst` added


acul71 (approved):
LGTM — approve.

Correct, well-scoped de-flake for #1401: `_get_peers_to_send` skips topics absent from `peer_topics`, and the fixed `trio.sleep(0.2)` after subscribe raced on slow CI. Switching to `Pubsub.wait_for_subscription` is the right fix (event-based, 5s timeout, early-return if already registered). Newsfragment and issue linkage look good; CI is green including `windows (3.12, core)`.

**Optional follow-up (out of scope, not blocking):** a remaining `await trio.sleep(0.2)` after host `connect()` in this test (and similar fixed sleeps in sibling score-gate tests) could still flake under connect/mesh races later. Worth a separate audit with per-test readiness predicates when someone has bandwidth — not required for this PR.

Issue #1403, opened 2026-07-22
pubsub: follow-ups after subscription-before-stream mcache replay fix (#1398)


acul71:
Follow-up tracking for items called out in [PR #1398](https://github.com/libp2p/py-libp2p/pull/1398) (merged) and its review. The merged fix replays mcache from `_handle_new_peer` when a peer's subscription is processed before the outbound pubsub stream exists.

## Disconnect / reconnect hygiene

- **`_handle_dead_peer` early return** (`pubsub.py`): when `peer_id not in self.peers`, the handler returns before `peer_topics` cleanup, so a peer that drops while half-registered can leave stale `peer_topics` entries.
- **Reconnect replay**: `_send_recent_messages_to_new_peer` has no `was_newly_added` guard (unlike `handle_subscription`). After a messy disconnect + reconnect, a peer may receive a **full mcache window** again even when subscriptions were not newly announced. Mitigation today is mostly receiver seen-cache dedup; sender-side gating may be desirable once disconnect cleanup is fixed.

## Deeper fix (separate design)

- **`GossipSub.publish` queueing**: peers known only via in-flight connection setup may still be in neither `pubsub.peers` nor `peer_topics` at publish time, so nothing is queued in `_pending_messages`. PR #1398 patched the recovery path only; changing queueing semantics affects all pubsub users.

## Maintainability

- **`IPubsubRouter`**: declare optional/async hooks for `send_recent_messages` and `flush_pending_messages` instead of additional `hasattr(self.router, ...)` probes in `pubsub.py`.

## Tests (optional)

- Tighten `test_publish_before_identify_with_subscription_before_stream` timing (currently ~7s+) with predicate-based `wait_for` helpers if available elsewhere in pubsub tests.

## Context

- Original flake: `test_publish_before_identify_completes` under `pytest -n auto` (real message-loss race, not a bad test).
- No separate tracking issue existed before merge; newsfragment `1398.bugfix.rst` used PR number per `newsfragments/README.md`.

/cc @gmelodie

acul71:
Hello @gmelodie thanks for your contributions,  are you willing to take this issue ?

gmelodie:
sure thing

acul71:
Closing as completed for the tracked follow-ups.

Disconnect hygiene, once-per-subscription mcache replay gating, router ABC hooks, and faster waits landed in #1407 (`newsfragments/1403.*`).

[...]

PR #1398, opened 2026-07-20
fix(pubsub): replay recent messages once a peer's outbound stream is registered

gmelodie:
## What was wrong?

Issue # — none filed; this is the flake behind `tox (3.12, core)`, e.g. [run 29641232846](https://github.com/libp2p/py-libp2p/actions/runs/29641232846) (push to `main`).

`tests/core/pubsub/test_gossipsub_identify_aware_publish.py::test_publish_before_identify_completes` fails intermittently under `pytest -n auto`. It is a real message-loss bug, not a bad test.

Publishing right after `connect()` can catch the peer in neither `pubsub.peers` nor `pubsub.peer_topics`. `GossipSub.publish` queues into `_pending_messages` only for peers in one of those two, so nothing is queued at all. The safety net is the mcache replay in `handle_subscription`, but `send_recent_messages` returns early when `peer_id not in pubsub.peers`. That replay runs off the peer's *inbound* stream, independent of the *outbound* stream `_handle_new_peer` registers — so when the subscription is processed first, the replay no-ops, the later `flush_pending_messages` finds an empty queue, and the message is lost with nothing to retry it.

An idle machine finishes identify before `publish()` runs, so the path is never exercised. CI load exercises it.

## How was it fixed?

`_handle_new_peer` now runs the replay as well, once it has registered the stream, via `_send_recent_messages_to_new_peer`. Whichever of the two events lands last triggers the catch-up.

Neither ordering double-sends: the path that ran first finds nothing to do. If a subscription lands concurrently between the two, the duplicate is dropped by the receiver's seen cache (`push_msg` → `_is_msg_seen`).

This is on the connection-setup path for floodsub too, where the `hasattr` check makes it a no-op. Behaviour changes in the fixing direction only; no API change.

### Testing

`test_publish_before_identify_with_subscription_before_stream` forces the losing ordering with asymmetric `_handle_new_peer` delays (publisher 1.0s, subscriber 0.3s) instead of hoping for it. It fails on `main` and passes here. Full `tests/core/pubsub` green (402 tests) rebased onto `bb346f20`; `ruff` and `mypy -p libp2p` clean.

### Not fixed here

- `_handle_dead_peer` returns early at pubsub.py:877, before the `peer_topics` cleanup, so a peer dropping while half-registered leaks its entry. This replay has no `was_newly_added` guard like `handle_subscription`'s, so a reconnect would replay the full gossip window. The fix is disconnect-handling surgery.
[...]

PR #1407, opened 2026-07-23
fix(pubsub): disconnect hygiene and once-per-subscription mcache replay

gmelodie:
## What was wrong?

Issue #1403

Follow-ups from the review of #1398:

- **`_handle_dead_peer` returned too early.** A peer's subscriptions arrive on its *inbound* stream, so it can sit in `peer_topics` while the *outbound* stream was never registered in `peers`. The `if peer_id not in self.peers: return` guard ran before the `peer_topics` cleanup, so a peer that dropped while half-registered left stale subscription entries behind forever.
- **Replay had no gating.** `_send_recent_messages_to_new_peer` replayed the whole mcache window for every topic the peer appeared subscribed to, with no equivalent of `handle_subscription`'s `was_newly_added` guard. Combined with the stale entries above, a messy disconnect + reconnect could hand a peer the full window again; the only thing saving us was receiver-side seen-cache dedup.
- **`hasattr` probes.** `pubsub.py` reached into the router with four `hasattr(self.router, ...)` checks plus `# type: ignore[attr-defined]` on every call, so the optional hooks were invisible to type checkers and to anyone implementing a router.

## How was it fixed?

**Disconnect hygiene** — `_handle_dead_peer` now runs `_clear_pending_announce_retries_for_peer` and a new `_forget_peer_subscriptions` *before* the `peers` early return. Stream teardown, queue close, and `router.remove_peer` still run only for fully-registered peers, since router registration happens in lockstep with `self.peers` (no `await` between `router.add_peer` and `self.peers[peer_id] = stream`). `_forget_peer_subscriptions` also replaces the open-coded loop in `_teardown_if_connected`, so the blacklist path gets the same cleanup.

**Sender-side replay gating** — all mcache replay now goes through one method, `_replay_recent_messages(peer_id, topic)`, called by both `_handle_new_peer` and `handle_subscription`. It records `(peer, topic)` in `_replayed_recent_topics` so a peer receives the window at most once per subscription. Two details matter:

- The gate is only spent when the peer is actually writable (`peer_id in self.peers`). Marking it while the outbound stream is still pending would make the later `_handle_new_peer` replay a no-op and reintroduce the exact bug #1398 fixed.
- The entry is dropped on unsubscribe, on `_handle_dead_peer`, and on blacklist, so a genuine resubscribe or reconnect replays again.

**Router interface** — `IPubsubRouter` declares `flush_pending_messages` and `send_recent_messages` as non-abstract async hooks with no-op bodies. GossipSub's existing implementations become real overrides, FloodSub inherits the no-ops, and all four `hasattr` probes and their `type: ignore` comments are gone.

**Not in this PR:** the `GossipSub.publish` queueing item from the issue. It is filed there as a separate design — peers known only via in-flight connection setup are in neither `peers` nor `peer_topics` at publish time, and changing when publish queues speculatively affects every pubsub user.

### Tests

- `TestRecentMessageReplayGate` (4 tests): replay runs once per subscription; is deferred but *not* spent before the peer is writable; runs again after unsubscribe/resubscribe; runs again after disconnect/reconnect.
- `test_dead_peer_clears_subscriptions_of_half_registered_peer` — a peer in `peer_topics` but never in `peers` is cleaned up.
- `test_router_replay_hooks_default_to_noop` — FloodSub is callable through the new interface hooks.
- `test_publish_before_identify_with_subscription_before_stream` now waits on `wait_for_peer` instead of `trio.sleep(publisher_stream_delay + 4)`, as the issue suggested. Runtime drops from ~7s to ~1.4s.

`tests/core/pubsub` is green (408 passed), as are `ruff check`, `ruff format --check` on the touched files, `mypy -p libp2p`, and `pyrefly check`.

### To-Do

- [x] Clean up commit history
- [x] Add or update documentation related to these changes — docstrings on the new `IPubsubRouter` hooks and `_replay_recent_messages`; no user-facing docs describe these internals
- [x] Add entry to the [release notes](https://github.com/libp2p/py-libp2p/blob/main/newsfragments/README.md) — `1403.bugfix.rst` and `1403.internal.rst`

#### Cute Animal Picture

otter


gmelodie:
@seetadev whenever you have a chance :)

seetadev:
Fantastic work, @gmelodie! 🎉

This is a very thoughtful follow-up that tightens the pubsub implementation. We especially like the disconnect hygiene improvements, the once-per-subscription mcache replay gating, and the cleanup of the router interface by replacing the `hasattr` probes with well-defined no-op hooks. These changes make the implementation more robust, easier to reason about, and much friendlier for future router implementations.

A huge thank you as well to @sumanjeet0012 for continuously keeping this work in sync with the latest changes on `main` and for the careful collaboration throughout the review process. That ongoing synchronization makes a tremendous difference and helps keep development moving smoothly.

As a next step, wish to request you to collaborate on scaling the Network analysis for libp2p and ethp2p using Wireshark dissectors: please visit [Milestones & Roadmap for libp2p and ethp2p Network Analysis](https://github.com/libp2p/libp2p/discussions/281) and [Wireshark Dissectors repo](https://github.com/seetadev/wireshark-dissectors). We would also like you to focus on building an observability stack for libp2p using **[NetViz](https://github.com/ethp2p/netviz)** by Raul and **[Luminar](https://github.com/seetadev/luminar-dev)** by @yashksaini-coder. I think this would be a great opportunity to bring first-class visibility into py-libp2p deployments and help us:

* improve throughput,
* reduce end-to-end latency,
* optimize bandwidth utilization,
* better understand peer connectivity and routing behavior,
* and provide actionable operational metrics for developers.

I will create a dedicated discussion page in the py-libp2p repository so we can collaborate on the overall architecture, instrumentation points, dashboards, and roadmap.

On the same note, we have also been discussing a number of new PubSub improvements with the Ethereum Foundation team. I'll be sharing the latest specification links and proposals shortly so we can align py-libp2p with the evolving libp2p PubSub specification and identify opportunities for future implementation work.

Reed Solomon strategy was recommended by Marco when @johannamoran and I had a discussion with him: please visit [Reed Solomon strategy for pub-sub](https://github.com/ethp2p/ethp2p/blob/main/specs/003-ec-broadcast-rs.md) . We could implement this in py-libp2p and nim-libp2p and open a spec PR for the same proposal.

Excellent work once again. Looking forward to getting this merged and continuing to build on top of it. 

CC: @acul71, @pacrob, @yashksaini-coder, @sumanjeet0012, @johannamoran, @momack2, @mishmosh.

PR #562, opened 2025-04-06
Fix flaky test_simple_last_seen_cache with retry loop and docstring

paschal533:
## What was wrong?

Issue #558 

The test `test_simple_last_seen_cache` in `tests/core/tools/timed_cache/test_timed_cache.py` was flaky on Windows due to timing jitter with `LastSeenCache` expiry and sweep intervals, failing to consistently detect expired entries.

## How was it fixed?

Summary of approach:
Updated `test_simple_last_seen_cache` to use a retry loop (up to 1s) after sleeping past the TTL (2.5s), ensuring the cache sweep is detected reliably across platforms.

### To-Do

- [x] Clean up commit history (single commit)
- [x] Add or update documentation related to these changes (docstring added)
- [x] Add entry to the [release notes](https://github.com/libp2p/py-libp2p/blob/main/newsfragments/README.md)

#### Cute Animal Picture

Kitten

PR #1409, opened 2026-07-24
test(timed_cache): de-flake test_expiry_removal (#1408)

### LINKED ISSUE #1408: flaky: test_expiry_removal races the background sweeper thread (fixed trio.sleep)
## Summary

`tests/core/tools/timed_cache/test_timed_cache.py::test_expiry_removal` fails intermittently on CI (observed on `windows (3.11, core)`), passing on all Linux runners and locally.

```
assert MSG_1 not in cache.cache
AssertionError: assert b'msg1' not in {b'msg1': <expiry-ts>}
```

## Root cause

```python
cache = LastSeenCache(ttl=2, sweep_interval=1)
cache.add(MSG_1)
await trio.sleep(2.1)  # Wait for sweeper to remove expired items
assert MSG_1 not in cache.cache
```

`BaseTimedCache` runs its sweeper in a **background thread** (`while not self._stop_event.wait(self.sweep_interval)`), and the TTL is stored as **integer seconds** (`int(time.time()) + self.ttl`). So the removal only happens on a coarse ~1s cadence in a separate thread. The fixed `trio.sleep(2.1)` assumes the sweep tick lands within 2.1s of the add; on a slow/loaded Windows runner the thread wakeup is delayed and the entry is still present when the assertion runs.

## Fix

Replace the fixed sleep + immediate assert with a bounded poll until the sweeper removes the entry — the same idiom already used by `test_simple_last_seen_cache` in this file:

```python
with trio.fail_after(15):
    while MSG_1 in cache.cache:
        await trio.sleep(0.1)
```

Deterministic (exits as soon as the entry is swept), non-hanging (fails after 15s if the sweeper is genuinely broken).

## Note

Several sibling tests in the same file (`test_simple_first_seen_cache`, `test_timed_cache_expiry`, `test_timed_cache_stress_test`, `test_readding_after_expiry`) share the same fixed-sleep-then-assert-expired pattern and are latent flakes on the same root cause. Scoping this fix to the test that actually failed in CI; the siblings can be addressed as a follow-up.

yashksaini-coder:
## Summary

Closes #1408.

`test_expiry_removal` fails intermittently on CI (seen on `windows (3.11, core)`; green on all Linux runners and locally):

```
assert MSG_1 not in cache.cache
AssertionError: assert b'msg1' not in {b'msg1': <expiry-ts>}
```

## Root cause

```python
cache = LastSeenCache(ttl=2, sweep_interval=1)
cache.add(MSG_1)
await trio.sleep(2.1)  # Wait for sweeper to remove expired items
assert MSG_1 not in cache.cache
```

`BaseTimedCache` sweeps in a **background thread** (`while not self._stop_event.wait(self.sweep_interval)`) with an **integer-second TTL**. Removal therefore happens on a coarse ~1s cadence in a separate thread. The fixed `trio.sleep(2.1)` assumes the sweep tick lands within 2.1s; on a slow Windows runner the thread wakeup is delayed and the entry is still present when the assertion runs. A timing race, not a logic bug.

## Fix

Poll until the sweeper removes the entry — the idiom `test_simple_last_seen_cache` already uses in this same file:

```diff
-await trio.sleep(2.1)  # Wait for sweeper to remove expired items
-assert MSG_1 not in cache.cache  # Should be removed
+with trio.fail_after(15):
+    while MSG_1 in cache.cache:
+        await trio.sleep(0.1)
```

Deterministic (exits as soon as the entry is swept) and non-hanging (fails after 15s if the sweeper is genuinely broken).

## Scope

Limited to the test that actually failed in CI. Sibling tests in the file (`test_simple_first_seen_cache`, `test_timed_cache_expiry`, `test_timed_cache_stress_test`, `test_readding_after_expiry`) share the same fixed-sleep pattern and are latent flakes on the same root cause — notable as a follow-up, out of scope here.

## Test plan

- [x] `test_expiry_removal` run 5× locally → stable pass
- [x] Full `tests/core/tools/timed_cache/` → 8 passed
- [x] `ruff check` / `ruff format --check` clean
- [x] `newsfragments/1408.internal.rst` added


acul71 (approved):
## Approval

LGTM — ready to merge.

This correctly de-flakes `test_expiry_removal` by replacing a fixed `trio.sleep(2.1)` with a bounded `trio.fail_after` poll until the background sweeper removes the entry (`Closes #1408`). Using `MSG_1 in cache.cache` is the right predicate: `LastSeenCache.has()` refreshes TTL on every hit and would defeat a wait loop.

Scope is tight (test + `1408.internal.rst`), branch is in sync with `main`, CI is green including Windows, and local lint/typecheck/docs/tests look good.

### Follow-up (non-blocking)

Sibling tests in the same file still use fixed sleeps / a `has()` retry that can refresh TTL (`test_simple_last_seen_cache` from #562). I’ll open a separate issue/PR for those after this lands.

**Merge:** squash and merge.

PR #1426, opened 2026-08-04
 fix: networking stack hardening — DHT spec compliance, QUIC/bitswap leaks, connmgr bugs 

### LINKED ISSUE #1425: Kademlia DHT Spec Compliance Issues
# Kademlia DHT Spec Compliance Issues

## Summary

The Kademlia DHT implementation (`libp2p/kad_dht/`) has multiple deviations from the [IPFS DHT specification](https://specs.ipfs.tech/protocols/kademlia/) and [libp2p Kademlia spec](https://github.com/libp2p/specs/blob/master/KAD-DHT.md). These issues affect correctness, security, and interoperability with other libp2p implementations.

### Critical

1. **`timeReceived` uses wrong format**: The `timeReceived` field in records uses Unix epoch float instead of RFC3339Nano format per spec.

2. **PUT_VALUE stream not reset on validation failure**: Per spec: "On any error, the stream is reset." Currently the stream is closed gracefully instead.

3. **Record size not validated**: No maximum size check for PUT_VALUE records, allowing potential abuse.

4. **Value store has no eviction limit**: No LRU eviction, risking OOM with large numbers of stored values.

### Spec Compliance

5. **ADD_PROVIDER does not echo response**: Per spec, ADD_PROVIDER should echo the request back. Currently fire-and-forget.

6. **ADD_PROVIDER key length not validated**: No 80-byte max enforcement per spec.

7. **No ADD_PROVIDER rate limiting**: No per-peer rate limiting for ADD_PROVIDER messages.

8. **FIND_NODE key not validated**: Key should be validated as a valid multihash/PeerId.

9. **GET_VALUE key size unlimited**: No 128-byte max key size enforcement per spec.

10. **GET_VALUE does not validate signatures**: Records served without signature validation.

11. **Closer peer addresses not stored**: GET_VALUE closer peers addresses not stored in peerbook per spec.

12. **Connection type always CONNECTED**: Connection type field always reports CONNECTED instead of dynamic reporting.

### Provider System

13. **Provider lookup is single-shot**: find_providers only queries known providers, not using iterative lookup with closer peers.

14. **No provider record republishing**: Provider records are not republished every 22 hours per spec.

15. **No provider count limit**: No k=20 limit on providers per key.

16. **CID validation missing**: No multihash structure checks for provider keys.

### Value Store

17. **No value record republishing**: Value records are not republished.

18. **No persistence support**: No JSON save/load for provider or value stores.

### Peer Routing

19. **Lookup terminates too early**: Peer routing terminates when no new peers discovered, even if unqueried closest peers remain.

20. **No beta resiliency**: No BETA parameter ensuring closest peers are queried before termination.

21. **No query timeout**: No total query timeout, risking infinite loops.

22. **Failed peers not removed**: Peers that fail connection are not removed from routing table.

### Security

23. **No IP diversity filtering**: No limit on peers per /16 subnet per bucket.

24. **No max varint protection**: Varint reading loops have no cap, risking DoS.

25. **Private networks not rejected**: No validation against reserved/private IPs per RFC 6890.

26. **Server-mode only not enforced**: Handlers registered regardless of mode.

### Bug Fixes

27. **Hardcoded BUCKET_SIZE**: Value 20 hardcoded instead of using constant.

28. **PUT_VALUE does not reject on validator failure**: PUT_VALUE proceeds when `validator.select()` fails.

29. **Routing table refresh uses wrong key**: Refresh does not use random key within bucket's XOR range.

## Environment

- py-libp2p version: latest main branch
- Python version: 3.12+
- OS: macOS/Linux

### LINKED ISSUE #1450: Bitswap: Stream Management, Memory & Performance Issues
## Summary

The Bitswap implementation (`libp2p/bitswap/`) has architectural and performance issues: it dials a new stream per message instead of reusing persistent streams, has no message batching, suffers from O(n) memory growth on large file transfers, and contains dead code.

### Architecture

- **No persistent stream reuse**: A new stream is dialed per message instead of maintaining one persistent outbound stream per peer (Kubo/go-bitswap architecture).
- **No message batching**: Wantlists, cancels, block presences, and data blocks are not coalesced into debounced write frames; every update triggers a separate send.
- **No automatic stream reconnection**: Streams are not transparently re-established on errors, causing per-message dialing overhead.

### Performance

- **O(n) memory growth on large transfers**: `leaf_triples` retains `block_bytes` for every leaf after the block is stored, so peak memory scales ~8x file size (e.g. ~16 GB for a 2 GB file).
- **Wasted dials for DontHave responses**: Outbound dial-back streams are dialed for `DontHave` responses.

### Logging / Cleanup

- **Noisy wantlist logging**: Per-wantlist-entry traces are logged at WARNING level.
- **Dead code present**: Unused DecisionEngine code remains in the module.

## Environment

- py-libp2p version: latest main branch
- Python version: 3.12+
- OS: macOS/Linux

### LINKED ISSUE #1451: Network & Swarm Module — Dial Pipeline, Auto-Connector & Connection Management Issues
## Summary

The network layer (`libp2p/network/`) has issues in the dial pipeline, the auto-connector, and connection management: the dial pipeline is not aligned with go-libp2p, the auto-connector causes CPU bursts and wasted dials at scale, connection registration can be lost to dial deadlines, and the API surface is missing O(1) peer checks and notifee removal.

### Dial Pipeline

- **Dial pipeline not aligned with go-libp2p**: No dial batching, no `MaxConcurrentDials` cap, no transport ranking.
- **Inbound connection cap not atomic**: The inbound connection cap is not enforced atomically, allowing races.
- **Connection registration lost to dial deadline**: Connection registration can be cancelled by the dial deadline, decaying connected peers to 0.
- **Inbound security-upgrade failures logged at ERROR**: Expected failures flood the error log.

### Auto-Connector

- **Always tops up to high water**: The auto-connector aims for the high-water mark instead of the midpoint of `[low_water, high_water]`.
- **CPU bursts at scale**: Dial bursts when scaling to 300+ peers cause CPU spikes.
- **Wasted dials**: Dials are wasted on relay addresses, IPv6, and QUIC-first candidates.
- **Private-only candidates dialed on public nodes**: Peers with only private addresses are dialed by public nodes.
- **Concurrent auto_connect tasks / dial storms**: Disconnect events can spawn concurrent auto-connect tasks and dial storms.
- **Unbounded tracking caches**: AutoConnector tracking caches grow without pruning.

### Stream / Connection Layer

- **`net_stream` lacks `is_closed`**: No way for leak-reconciliation consumers to check stream closure.
- **`INetwork` lacks `remove_notifee`**: Notifees cannot be unregistered.
- **`IPeerStore` lacks O(1) `has_peer`**: Peer checks materialize the full peer list.
- **Dead code present**: Dead `notify_all` in the swarm.

## Environment

- py-libp2p version: latest main branch
- Python version: 3.12+
- OS: macOS/Linux

### LINKED ISSUE #1452: QUIC Transport — Memory, Socket & CPU Efficiency Issues
## Summary

The QUIC transport (`libp2p/transport/quic/`) suffers from resource leaks and CPU inefficiency: connection objects and registry entries leak, UDP sockets and background tasks are not closed on dial failure/cancellation, and the event-processing loop busy-spins.

### Memory Leaks

- **ServerQuicConnection leaks**: Connections leak in the registry and listener tracking.
- **Registry CIDs not purged on disconnect**: Connection IDs are not unregistered/purged when a connection disconnects.
- **Multiple major memory leaks**: At least four distinct leak sources exist across connection/registry/listener paths.
- **`get_all_cids_for_connection` is O(n)**: Linear scans instead of O(1) lookups.

### Socket / FD Leaks

- **UDP socket not closed on dial cancellation/failure**: Dial failures leak file descriptors.
- **`close()` does not guarantee socket close**: Socket close and background-task cancellation are not guaranteed.
- **`_handle_connection_terminated` leaves sockets open**: UDP socket and background task scopes not closed.
- **Failed/cancelled dials leak tasks and sockets**: Background tasks survive failed dials.

### CPU / Event Loop

- **Event loop busy-spins**: The processing loop spins at 100% CPU waiting for events.
- **2s delay when `timer <= now`**: Expired timers wait for the full poll interval.
- **No idle sleep**: The loop does not sleep when idle.
- **Aggressive idle polling**: Idle poll interval too short; no activity signaling on stream write.

### Cleanup

- **Temporary CONN_CLOSE_TRACE stack logging** left in the code.
- **DEBUG log level forced at import time**.
- **Syntax error in an except block**.

## Environment

- py-libp2p version: latest main branch
- Python version: 3.12+
- OS: macOS/Linux

sumanjeet0012:
## Description

This PR fixes the issues documented in **#1425** (Kademlia DHT), **#1450** (Bitswap), **#1451** (Network & Swarm), and **#1452** (QUIC), along with additional correctness, robustness, and performance fixes across the networking stack. It brings the DHT implementation closer to the IPFS DHT specification and libp2p Kademlia spec, and resolves memory leaks, CPU busy-spins, and connection-management bugs found while running the node against the public network.

### Kademlia DHT (`libp2p/kad_dht/`) — closes #1425

**Critical**
- `timeReceived` now uses RFC3339Nano format per spec (was Unix epoch float), with backward-compatible parsing of both formats.
- `clean_record()` strips `timeReceived` from incoming records to prevent timestamp forgery.
- PUT_VALUE stream is reset on validation/storage failure per spec ("on any error, the stream is reset") instead of closed gracefully.
- PUT_VALUE records capped at 1 MB (`MAX_RECORD_SIZE`).
- Value store LRU eviction at 50K entries (`MAX_VALUE_STORE_SIZE`) to prevent OOM.

**Spec compliance**
- ADD_PROVIDER: receiver echoes the request per spec; key length capped at 80 bytes; per-peer rate limit (10 msgs / 10s window); max 20 provider records per message.
- FIND_NODE key validated as a valid multihash.
- GET_VALUE: 128-byte max key size; records signature-validated before serving; closer-peer addresses stored in the peerbook.
- Connection type now reports dynamically (CONNECTED / CAN_CONNECT / NOT_CONNECTED) instead of always CONNECTED.

**Provider system**
- Iterative provider discovery using closer peers from responses (was single-shot).
- Provider records republished every 22 hours per spec.
- Providers per key capped at k=20.
- Provider keys validated for multihash structure.
- JSON persistence for `ProviderStore` via `persist_dir`.

**Value store**
- Value records republished every 22 hours.
- LRU eviction (50K) prevents OOM.
- JSON persistence for `ValueStore` via `persist_dir`.
- Closer-peer addresses stored in peerbook when processing GET_VALUE responses.

**Peer routing**
- Iterative lookup continues while unqueried closest peers remain (was terminating early).
- BETA=3 resiliency: at least 3 closest peers queried before termination.
- Total query timeout (10s per peer) prevents hangs on unresponsive peers.
- Failed peers removed from the routing table.
- Candidates re-sorted by distance after discovering closer peers.

**Security**
- IP diversity filtering: max 2 peers per globally-routable subnet per bucket (eclipse-attack protection, issue #1383).
- Max varint protection: all varint reads use `read_varint_prefixed_bytes_limited` (10-byte prefix cap).
- KAD handlers only registered in server mode.
- Inbound concurrency limiter (max 12 concurrent handlers).

**Bug fixes**
- ADD_PROVIDER sender is fire-and-forget: Kubo never sends an echo response, so waiting for one blocked every announcement for the query timeout (this was the root cause of `dht/provide` timing out against the public network). The receiver still echoes per spec.
- Sliding-window ADD_PROVIDER sends (semaphore, ALPHA in flight) so a slow peer no longer stalls the whole batch.
- Default namespace validators (`pk`, `ipns`) now registered via `apply_fallbacks()` in `KadDHT.__init__` — IPNS records were previously accepted without validation.
- IPNS validator accepts base58btc peer-ID names (`/ipns/<base58>`, the go-ipfs / py-ipfs-lite format) in addition to hex multihash.
- DHT walk self-exclusion: local peer excluded from `peers_to_query` to prevent hangs.
- Correct CID key encoding in `provide`/`find_providers` (`cid_to_bytes(parse_cid())` instead of `key.encode()`).
- PUT_VALUE key comparison fixed (`message.key` vs `record.key`).
- `cleanup_expired` no longer raises `UnboundLocalError` on an empty provider store.
- `get_value` uses network lookup instead of local routing table only.
- Cancel-on-quorum for value retrieval with per-query cancel scopes.
- `BUCKET_SIZE` constant replaces hardcoded 20.
- PUT_VALUE rejects (resets stream) when `validator.select()` fails.
- Routing-table refresh uses a random key within the bucket's XOR range.

**Cleanup**
- Removed dead code: `_propagate_to_closest_peers()`, `is_cid_like_key`, `is_reserved_or_private_addr`, `PEER_REFRESH_INTERVAL`, `_rt_refresh_nursery`, `should_republish()`, `PROVIDER_ADDRESS_TTL`, `gen_random_peer_id_with_cpl()`, `get_active_cpls()`, `_get_providers_from_peer()` wrapper.
- Deduplicated response building (`_add_closer_peers_to_response()`) and key encoding (`_encode_key()`).

### Bitswap (`libp2p/bitswap/`) — closes #1450

- New `BitswapMessageQueue` (Kubo/go-bitswap architecture): one persistent outbound stream per peer, debounced batching of wantlists/cancels/presences/blocks (10–20 ms), automatic stream reconnection without per-message dialing.
- `client.py` rewritten to reuse persistent streams instead of dialing per message.
- No more outbound dial-back streams for `DontHave` responses.
- Fixed O(n) memory growth in `leaf_triples`: peak memory reduced from ~16 GB to ~100 MB on a 2 GB file transfer.
- Removed dead DecisionEngine code; wantlist traces demoted from WARNING to DEBUG.

### Network & Swarm (`libp2p/network/`) — closes #1451

- Dial pipeline aligned with go-libp2p: dial batching, `MaxConcurrentDials=16`, transport ranking.
- Atomic inbound connection cap via `CapacityLimiter.acquire_nowait()`.
- Connection registration shielded from the dial deadline (fixes peers decaying to 0).
- Auto-connector: midpoint target in `[low_water, high_water]`, paced dials (no CPU bursts at 300+ peers), skips relay/IPv6/QUIC-first candidates and private-only peers on public nodes, prevents concurrent auto-connect tasks and dial storms.
- AutoConnector tracking caches pruned; peerstore limit enforced on `add_addrs`.
- `net_stream.is_closed` property added; `INetwork.remove_notifee()` and O(1) `IPeerStore.has_peer()` added.
- Inbound security-upgrade failures logged at DEBUG instead of ERROR.

### QUIC Transport (`libp2p/transport/quic/`) — closes #1452

- Eliminated ServerQuicConnection and registry connection leaks; CIDs purged on disconnect.
- UDP sockets and background tasks guaranteed closed on dial failure/cancellation and in `close()` (FD-leak fixes).
- Event-processing loop de-busy-spun: activity-event signaling, timer-aware polling (1–10 ms), idle sleep, no 2 s delay on expired timers.
- `get_all_cids_for_connection` made O(1).
- Removed temporary stack logging; DEBUG level no longer forced at import.

### Connection Manager (`libp2p/rcmgr/`)

- 15 connection-manager bugs fixed: `max_connections` enforced on outbound dials and at registration (race-free), `min_connections` functional, lifecycle manager wired into the swarm, `add_conn` dedup no longer closes the shared muxed connection, safe per-peer trimming, background pruning with debounce, auto-connect on disconnect with re-dial back-off, notifee failure isolation, deterministic `Swarm.close()`, defensive getter copies, pruner allow-list uses the real remote IP, `dial_peer` results capped, dead connection pool removed.
- Lifecycle-tracker ratchet fixed (node no longer wedges at 0 peers).
- O(1) `has_peer` and raw-byte ID hashing.

### Random Walk / Discovery (`libp2p/discovery/random_walk/`)

- Startup query storms reduced: concurrency 10 → 3, refresh interval 60 s → 120 s.
- Post-walk connection pruning; redundant trim in `RTRefreshManager` removed.

### Yamux (`libp2p/stream_muxer/yamux/`)

- `read()` returns available data without waiting for EOF (prevents reader/writer deadlocks).
- Stale-stream sweeper reclaims bookkeeping for fully-closed streams with unread data.
- Backlog semaphore slot guarded against double-release.

### PubSub (`libp2p/pubsub/`)

- Peer registration retried with capped exponential backoff (0.5–10 s) so peers are never silently left unregistered when the notifee fires before the handshake completes.
- Dead peer streams re-established; new public idempotent `ensure_peer_stream()` API for out-of-band connections.

### Peer & Peerstore (`libp2p/peer/`)

- Memory leaks from `defaultdict` auto-creation plugged; `peer_data_map` now size-limited (default 1000 peers) with eviction.
- `KeyError` fixed in `add_pubkey`/`add_privkey`/metrics on uninitialized peer data.
- `ID.__hash__` hashes raw bytes instead of base58-encoding on every call (dominant CPU cost on production nodes).

### Host / Identify / Ping (`libp2p/host/`)

- `fail_after` replaced with `move_on_after` in `basic_host.py` and multiselect to eliminate `TooSlowError` CPU burn.
- Identify: 60 s backoff on `new_stream` failure, per-peer backoff to prevent cancellation storms, `UnboundLocalError` crash fixed.
- Ping streams closed after completion / reset on error (stream-leak fix).
- Cancellation leak fixed in `new_stream`/`send_command`.

### TLS (`libp2p/security/tls/`)

- Loads the exact `libssl` linked by CPython on macOS so certificate verification works.
- pyrefly type-narrowing errors resolved (unblocked CI).

### Records / IPNS (`libp2p/records/ipns.py`)

- IPNS validator accepts base58btc peer-ID names in `/ipns/` keys in addition to hex multihash.
- With default validators now registered (see DHT fixes), IPNS records are actually validated instead of silently accepted.

### Utils / AnyIO Service

- IPv6 availability probe (cached), relay-address detection for the auto-connector, routability helper with CGNAT allowance.
- `FunctionTask.cancel()` applied to tasks not yet inside their cancel scope (fixes `wait_done()` hanging forever).

## Testing

- 236 DHT + records tests pass; full suite green except 3 pre-existing WebRTC varint-logging failures that also fail on upstream main (unrelated to this PR).
- `make lint` fully green (ruff, mypy, pyrefly).
- Manual staging verification against the public network: `dht/provide` (previously timing out at 30 s), `name/publish`/`name/resolve` (previously failing validation), `cat` of remote content, and `swarm/disconnect` all verified working end-to-end.


seetadev:
@sumanjeet0012: Thank you for the continued work here, **great progress on the Kad-DHT side of py-libp2p!**. Keeping in sync with the changes.

Really appreciate the attention to spec compliance, correctness, resilience, and test coverage across the DHT implementation. The improvements around routing, provider/value handling, lookup behaviour, timeouts, IP diversity, and interoperability are particularly valuable. 

It’s also great to see the iterative follow-up fixes addressing real edge cases such as self-discovery during DHT walks, correct CID key encoding, fire-and-forget `ADD_PROVIDER`, lookup timeouts, and proper cancellation/error handling. 

**Let’s keep the momentum going!** Looking forward to seeing continued, substantial improvements across Kad-DHT in py-libp2p, particularly as we continue pushing toward stronger spec compliance, interoperability with other libp2p implementations, robustness under larger and more adversarial networks, and better operational behaviour.

Thanks again for the excellent work and persistence on this. This is exactly the kind of sustained engineering effort that will make the py-libp2p DHT implementation significantly stronger.

sumanjeet0012:
@seetadev — this PR is ready for review.

The Kademlia DHT spec-compliance work is complete (RFC3339 timeReceived, stream reset on error, key/record size limits, rate limiting, IP diversity, iterative lookups, provider/value republishing, persistence).

Along the way we also fixed real issues found while running against the public network:  dht/provide  timing out (fire-and-forget ADD_PROVIDER + sliding-window sends), IPNS validation being silently skipped, and  name/publish  failing on base58 names.

Plus resource leaks and CPU issues: QUIC connection/socket/FD leaks and event-loop busy-spin, bitswap stream-per-message dialing and O(n) memory growth on large files, and 15 connection-manager bugs.

Also removed dead code and deduplicated the DHT module.

seetadev:
This is a substantial and very valuable hardening PR for `py-libp2p`. **Strong +1 from me and happy to approve.** 

The breadth of this work is impressive, but more importantly, the fixes are addressing issues that matter directly to real-world interoperability, correctness, resilience, and long-running node stability.

The **Kademlia DHT work in particular is a major step forward**. Bringing `timeReceived` into the RFC3339 format, correctly resetting streams on validation/storage errors, enforcing record and key-size limits, validating signed records, adding provider rate limiting, IP diversity protections, iterative provider discovery, β=3 lookup resiliency, persistence, and republishing brings the implementation much closer to the expected IPFS DHT/libp2p behavior. 

I especially appreciate that this wasn't treated as purely a spec-compliance exercise. Several of the changes were validated against the **public network** and resulted in fixing real interoperability issues, notably `dht/provide` timing out because of the ADD_PROVIDER behavior, IPNS validation being skipped, and base58 PeerID names failing during `name/publish`. 

The security and resource-safety improvements are equally important. The combination of input limits, rate limiting, IP/subnet diversity, bounded stores, varint protection, and concurrency limits substantially improves the defensive posture of the DHT. 

The **Bitswap and QUIC fixes** are also particularly compelling from an operational perspective. Moving Bitswap toward persistent streams, eliminating per-message dialing, and addressing the large-file memory growth is a significant improvement. The reported reduction from roughly **16 GB peak memory to ~103 MB for a 2 GB transfer** is an excellent example of a fix that translates directly into better production behavior.

Likewise, addressing QUIC connection/socket/FD leaks and event-loop busy-spinning, together with the extensive connection-manager fixes, should make long-running nodes considerably more stable. The PR description calls out 15 connection-manager bugs, including race-free connection limits, safer deduplication, background pruning, deterministic shutdown, and better auto-reconnection behavior. 

I also really like the attention to **cross-implementation interoperability**. Fixes such as correct CID key encoding, dynamic connection-type reporting, signed-record validation, closer-peer handling, and the Mplex reset behavior all address the kinds of subtle differences that tend to surface only when `py-libp2p` is interacting with other implementations. The Mplex fix, for example, specifically restores the ability to consume buffered data after a remote reset and resolves a rust-libp2p ping interoperability issue. 

Finally, the validation gives me good confidence in the overall change. The PR reports **236 DHT + records tests passing, lint/type checks green, and manual staging verification against the public network**, including `dht/provide`, IPNS publish/resolve, remote content retrieval, and swarm disconnect behavior. The current branch also shows **38 successful checks with no merge conflicts**. 

Overall, this is exactly the kind of foundational maintenance work that strengthens `py-libp2p` for broader ecosystem adoption: **better spec compliance, stronger security boundaries, improved interoperability, fewer resource leaks, and substantially better behavior under real network conditions.**

Huge thanks to @sumanjeet0012. This is a significant improvement to the networking stack and, in my view, **ready to merge.** Wish to also thank @acul71 for his code contributions.

CCing @johannamoran and @mishmosh.

PR #1431, opened 2026-08-12
test(quic): join nurseries instead of sleeping in listener race tests

### LINKED ISSUE #1430: Flaky: test_multiple_cid_routing_concurrent_load races on a fixed sleep
Hit this on CI (tox 3.13, core) on an unrelated PR:

```
FAILED tests/core/transport/quic/test_listener.py::TestQUICListenerRaceConditions::test_multiple_cid_routing_concurrent_load
AssertionError: All CIDs should route to same connection, got 0 different connections
assert 0 == 1  where 0 = len(set())
```

The set is empty, not wrong — the lookups just hadn't finished. The test spawns 20 `route_packet` tasks into the nursery, sleeps a fixed 0.1s, then checks how many distinct connections they found:

```python
for _ in range(5):
    for cid in cids:
        nursery.start_soon(route_packet, cid)
await trio.sleep(0.1)
connection_ids = {conn_id for _, conn_id in found_connections}
assert len(connection_ids) == 1
```

That 0.1s is a guess at how long the concurrent lookups take. On a loaded runner they don't all finish in time, so `found_connections` is still empty when the assert runs. The same job logged:

```
Slow find_by_connection_id: 115.01ms ...
```

so a single lookup alone took longer than the whole sleep budget.

`test_promotion_race_condition` in the same class has the identical shape — 10 concurrent `attempt_promotion` tasks, then `trio.sleep(0.1)`, then assert — so it's a latent flake too.

Both tests came from #1096, which added them to check CID routing/promotion races, so the race-safety tests racing on their own setup is a bit ironic.

The fix is to wait for the spawned tasks to actually finish instead of sleeping: join a nursery around them so the assert only runs once every task has completed. That drops the timing dependency entirely. I'll send a PR.

Run: https://github.com/libp2p/py-libp2p/actions/runs/31577983271/job/94054347949

yashksaini-coder:
Two tests in `TestQUICListenerRaceConditions` spawn a batch of tasks into the nursery, sleep a fixed 0.1s, then assert on what those tasks collected:

```python
for _ in range(5):
    for cid in cids:
        nursery.start_soon(route_packet, cid)
await trio.sleep(0.1)
connection_ids = {conn_id for _, conn_id in found_connections}
assert len(connection_ids) == 1
```

The 0.1s is a guess at how long the concurrent tasks take. On a loaded runner they aren't done yet when the assert runs, so it sees an empty or partial result. That's how it failed on CI (tox 3.13, core):

```
AssertionError: All CIDs should route to same connection, got 0 different connections
assert 0 == 1  where 0 = len(set())
```

The same job logged `Slow find_by_connection_id: 115.01ms`, so a single lookup took longer than the whole sleep budget. `test_promotion_race_condition` has the same shape and is a latent flake for the same reason.

Instead of sleeping, spawn the tasks into their own nursery and let the `async with` block join them — once it exits every task has finished, so the assertions are deterministic with no timing involved.

I left the `trio.sleep(0.1)` near line 256 alone: that one waits for the server to be ready before dialing, which is a separate concern from this race, and I didn't want to widen the diff.

Ran `TestQUICListenerRaceConditions` locally in a loop and the full `test_listener.py` file, all green.

Closes #1430. Background on this being the same fixed-sleep pattern we've been cleaning up: #1401, #1408.

### DIFF
--- a/newsfragments/1430.internal.rst
+++ b/newsfragments/1430.internal.rst
@@ -1,0 +1,1 @@
+Stop the two ``TestQUICListenerRaceConditions`` tests from racing on a fixed ``trio.sleep(0.1)``. ``test_multiple_cid_routing_concurrent_load`` and ``test_promotion_race_condition`` now join a nursery around their concurrent tasks so the assertions run only after every task has finished, instead of guessing how long the tasks take.
--- a/tests/core/transport/quic/test_listener.py
+++ b/tests/core/transport/quic/test_listener.py
@@ -421,12 +421,12 @@ class TestQUICListenerRaceConditions:
                     except Exception:
                         pass  # Some attempts may fail, that's expected
 
-                # Launch multiple concurrent promotion attempts
-                for _ in range(10):
-                    nursery.start_soon(attempt_promotion)
-
-                # Give time for promotions to complete
-                await trio.sleep(0.1)
+                # Launch concurrent promotion attempts. Wait for all of them to
+                # finish by joining a dedicated nursery instead of sleeping a
+                # fixed amount, which races under load.
+                async with trio.open_nursery() as attempt_nursery:
+                    for _ in range(10):
+                        attempt_nursery.start_soon(attempt_promotion)
 
                 # Verify only one connection object was created
                 assert len(set(connection_objects)) <= 1, (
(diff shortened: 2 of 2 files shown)


acul71 (approved):
LGTM — approving and merging.

The nested-nursery join correctly removes the fixed `trio.sleep(0.1)` race in `test_multiple_cid_routing_concurrent_load` and `test_duplicate_promotion_prevention` (newsfragment/PR text call the latter `test_promotion_race_condition`; naming nit only). CI is fully green, including `tox (3.13, core)`.

Follow-up (non-blocking): `test_handler_invocation_once_per_connection` still does `nursery.start_soon` × 20 then `await trio.sleep(0.2)` — same latent flake class. Worth the same nested-nursery treatment in a small follow-up when convenient.

PR #1441, opened 2026-08-14
fix(varint): reject truncated and over-long varints

### LINKED ISSUE #1440: varint: decoders accept truncated and over-long inputs
Two hardening gaps in `libp2p/utils/varint.py` on untrusted-input decode paths (mplex header, length prefixes):

1. `decode_uvarint` / `decode_varint_from_bytes` silently return a **partial value** when the input ends with the continuation bit still set (truncated varint) instead of raising — e.g. `decode_uvarint(b"\x80")` returns `0`.
2. `decode_uvarint_from_stream` accepts an **11-byte** varint (`shift` up to 70), allowing values > 64 bits; a canonical uvarint is ≤ 10 bytes.

Fix: `decode_uvarint` raises `ParseError` on a missing terminating byte; `decode_uvarint_from_stream` rejects the 11th byte. `decode_varint_with_size` is intentionally left lenient — identify's format-detection and the WebRTC stream decoder inspect its `bytes_consumed` return and guard truncation at the call site.

yashksaini-coder:
Closes #1440.

Two hardening fixes to the untrusted-input varint decoders in `libp2p/utils/varint.py` (mplex header, length prefixes, multistream):

- **`decode_uvarint`** silently returned a partial value when the input ended with the continuation bit still set — `decode_uvarint(b"\x80")` → `0`. It now raises `ParseError` on a missing terminating byte. Internal callers (`read_varint_prefixed_bytes_sync`, `read_length_prefixed_protobuf`) pre-build the length bytes ending on a terminator, so they're unaffected — the change only rejects genuinely malformed input.
- **`decode_uvarint_from_stream`** accepted an 11-byte varint (`shift` up to 70), encoding values > 64 bits. It now rejects the 11th byte, matching the canonical ≤10-byte limit.

`decode_varint_with_size` is intentionally left lenient — identify's format-detection and the WebRTC stream decoder inspect its `bytes_consumed` return and guard truncation at the call site (and `ParseError` isn't a `ValueError`, which those guards catch).

Local: full varint suite + mplex/identify/webrtc/msgio caller tests (83) green; ruff + mypy clean.


acul71 (approved):
Thanks @yashksaini-coder — clean, focused hardening for #1440.

**What landed well**
- `decode_uvarint` now fails closed on truncated input instead of returning a partial value (`b"\x80"` → `ParseError`, not `0`).
- `decode_uvarint_from_stream` correctly rejects the 11th byte (`shift >= SHIFT_64_BIT_MAX`) before reading it — the off-by-one fix is right.
- Leaving `decode_varint_with_size` lenient is the correct call; identify and WebRTC already guard truncation at the call site.
- Tests and `1440.bugfix.rst` match the issue; diff stays small and easy to reason about.

Approving and squash-merging. I opened a follow-up issue for the remaining edge cases (10-byte last-byte overflow, error-type consistency, docstrings) so we can track them without blocking this fix.

Great work — appreciate the careful call-site analysis in the PR description.

PR #1448, opened 2026-08-16
feat(webrtc): run aiortc RTCPeerConnection over UdpMux (attach_muxed_connection)

yashksaini-coder:
## What

Second step for #1437: lets a full aiortc `RTCPeerConnection` (DTLS + SCTP) run over a `UdpMux`-backed ICE connection, so a listener can demux concurrent inbound dials on one UDP port.

- **`attach_muxed_connection(pc, mux, conn)`** (`_aiortc_helpers.py`): call after `createDataChannel`, before any `set*Description`. Swaps `iceGatherer._connection` / `iceTransport._connection` for the mux-backed `aioice.Connection` **and rebinds `iceTransport._recv/_send`** (aiortc captures those at construction — without this DTLS keeps talking to the orphaned connection). Registers/unregisters the peer address on ICE state changes.
- **`UdpMux` fixes** surfaced by driving it with a real peer:
  - `add_ice_connection` also sets `_local_candidates_start`, so aiortc's `gather()` (issued from `setLocalDescription`) is a no-op instead of binding extra sockets and appending their candidates to the answer.
  - STUN **responses** carry no `USERNAME` — they were dropped. Now routed by address; addresses are learned from inbound checks and from outbound sends (pion-style), so DTLS racing ahead of our ICE `completed` still routes.
  - `unregister(ufrag)` drops the learned addresses for that protocol.
  - STUN-shaped-but-malformed datagrams go straight to the connection's data path (`StunProtocol.datagram_received` only catches `ValueError`, so `struct.error` escaped the loop callback).
- `webrtc` extra: `aiortc>=1.15,<2.0` (what this was validated against; 1.15 also fixed the credential-forwarding bug from #1352).

## Tests

`test_udp_mux.py`: mux-backed server PC vs plain aiortc client PC on loopback — answer advertises only the shared port + our creds + `a=setup:passive`, ICE/DTLS/SCTP connect, data flows both ways, mux tables empty after close. Plus addr-learning / unregister / malformed-STUN cases. `tests/core/transport/webrtc`: 187 passed; mypy/pyrefly clean.

Stacked on #1447 → #1446.

Refs #1437

PR #1447, opened 2026-08-16
fix(webrtc): wire the WebRTC-Direct inbound path; spec Noise roles; real loopback test

yashksaini-coder:
## What

First code step for #1437. Makes WebRTC-Direct actually connect end to end (it did not):

- **Listener inbound path was a stub** (`_complete_inbound` logged and returned; handler never called). Now: listener-owned trio nursery (system task, like TCP), non-blocking asyncio→trio hop, Noise XX as **initiator** (spec: server initiates, dialer responds), authenticated `WebRTCConnection` handed to the handler. In-flight unauthenticated inbounds bounded by `max_in_flight_connections`.
- **`dial()` was broken**: `get_remote_fingerprint` read `pc._dtlsTransport._remote_certificate`, which aiortc doesn't have. Now reads the DTLS peer cert from `pc.sctp.transport._ssl`. Dialer is the Noise responder and verifies the authenticated ID against `/p2p/`.
- **Noise prologue is role-ordered** (`dialer_fp ‖ server_fp`, per spec) instead of local/remote.
- `PatternXX.handshake_outbound(remote_peer: ID | None)` — skips only the equality check when `None` (server doesn't know the dialer's ID); signature verification unchanged. Existing callers unaffected.
- `DataChannelReadWriter.read(n)` now honours `n` — the Noise packet reader asks for the 2-byte length prefix alone; returning a whole channel message broke framing.
- Noise `send` waits for data channel 0 to open (initiator sends right after DTLS, before SCTP opens the channel → `InvalidStateError`).

## Tests

- New `test_dial_listen_open_stream_echo`: dial → ICE/DTLS → Noise → handler → stream echo, both sides see the other's peer ID. Plus wrong-`/p2p/` rejection, e2e Noise role test, buffered-read test, fingerprint test, `remote_peer=None` pattern test.
- `tests/core/transport/webrtc` + `tests/core/security/noise`: 319 passed locally; loopback run 5× no flake. mypy/pyrefly clean.

Stacked on #1446 (CI needs aiortc installed for these to run); will rebase once that merges.

Refs #1437

PR #1446, opened 2026-08-16
ci: install webrtc extra in tox so WebRTC tests run in CI

yashksaini-coder:
## What

`tox.ini` `commands_pre` now installs `-e ".[webrtc]"`, so `aiortc` is present in the `core` / `windows` test envs.

## Why

No extra was installed, so every test in `tests/core/transport/webrtc/` has been **skipped** in CI (`HAS_AIORTC` / `HAS_AIOICE` guards) — green meant untested. First step for #1437 (WebRTC-Direct STUN listener) so follow-up PRs actually prove themselves.

## Notes

- aiortc 1.15 is pure Python; `av` and `pylibsrtp` ship manylinux_2_28 + win_amd64 wheels for cp310–cp313 → no apt deps.
- Locally with the extra installed: `tests/core/transport/webrtc` 177 passed; `mypy -p libp2p` and `pyrefly check` clean.

Refs #1437

Issue #1437, opened 2026-08-14
feat(webrtc): track WebRTC-Direct STUN listener + v1/v2 (specs#715) after UdpMux


acul71:
## Status (2026-09-03)

**v1 STUN-dispatch listener + v1 dialer on `main` via [#1449](https://github.com/libp2p/py-libp2p/pull/1449)** (merge `1d0876da`; stack #1446–#1448).

**v2 flow + framing/SCTP fix on `main` via [#1459](https://github.com/libp2p/py-libp2p/pull/1459)** (merge `1ba536d3`, 2026-09-03). Listener accepts `libp2p+webrtc+v2/`; dialer opt-in `webrtc_direct_dial_version=2`. Framing as a byte stream + SCTP stream-id parity by DTLS role landed in the same stack. [#1460](https://github.com/libp2p/py-libp2p/pull/1460) was closed as superseded.

Default path: spec STUN on a shared UDP port (offer inferred from first packet). HTTP `POST /sdp` is **opt-in** (`WebRTCTransportConfig(enable_sdp_http_harness=True)`) for py↔py experiments only.

**Interop ([#1471](https://github.com/libp2p/py-libp2p/pull/1471), open):** harness vs go-libp2p v0.49. **py → go passes live for v1 and v2.** **go → py is xfail** — ICE completes, then the py listener tears the connection down ([#1470](https://github.com/libp2p/py-libp2p/issues/1470)).

**Still open on this issue:** go→py inbound teardown ([#1470](https://github.com/libp2p/py-libp2p/issues/1470)), formal interop PR ([#1471](https://github.com/libp2p/py-libp2p/pull/1471)), true ICE-Lite in aioice, optional UdpMux upstream, docs/README v1 vs v2, specs#715 merge tracking.

---

## Summary

Tracking issue for a **spec-aligned `/webrtc-direct` listener** that demuxes concurrent inbound dials on one advertised UDP port via STUN, including **WebRTC-Direct v1 and v2** version-prefix dispatch per [libp2p/specs#715](https://github.com/libp2p/specs/pull/715).

v1 is on `main` ([#1449](https://github.com/libp2p/py-libp2p/pull/1449)). v2 is on `main` ([#1459](https://github.com/libp2p/py-libp2p/pull/1459)).

[...]

PR #1449, opened 2026-08-16
feat(webrtc): STUN-dispatch WebRTC-Direct listener (v1) + v1 dialer, /sdp harness opt-in

yashksaini-coder:
## What

Third step for #1437 — the spec-aligned listener.

**Listener (default path)** — one shared UDP socket (`UdpMux`):
- First inbound STUN BINDING REQUEST → `USERNAME = server_ufrag:client_ufrag`; `libp2p+webrtc+v1/` prefix selects the flow. Unknown/missing prefix → **rejected** (spec: never assume v1). Both halves validated (ice-chars, 4–256).
- Registers a mux-backed ICE connection for the ufrag, replays the packet, attaches an aiortc PC (`attach_muxed_connection`, #1448).
- Infers the dialer's offer: ufrag == pwd == server credential, `c=`/candidate at the STUN source, `a=setup:active` (so aiortc takes the DTLS **server** role — with `actpass` it would answer as client), placeholder fingerprint with DTLS peer verification disabled for inbound per spec step 6.2/7 (Noise authenticates).
- Then the #1447 completion path (Noise initiator → handler). In-flight cap on unauthenticated inbounds (`max_in_flight_connections`).

**Dialer (v1)** — sets `libp2p+webrtc+v1/<random>` as ufrag **and** pwd on the aioice `Connection` (aiortc regenerates ICE creds from it in `setLocalDescription`; SDP-text munging is ignored) and on a synthetic ICE-Lite `a=setup:passive` answer built from the multiaddr; the certhash fingerprint in that answer makes aiortc pin the server's DTLS cert.

**HTTP `POST /sdp` harness** → opt-in `WebRTCTransportConfig(enable_sdp_http_harness=True)` (experimental, py↔py). Off by default; nothing else used it.

`sdp.py`: `parse_direct_username`, `build_inferred_offer`, `build_synthetic_answer`, `make_v1_credential`; credentials generated with ice-chars only (`token_urlsafe` produced `-`/`_`, which aioice rejects). `UdpMux` unknown-STUN handler now receives the full `USERNAME`.

## Not in this PR (tracked on #1437)
- v2 flow (`libp2p+webrtc+v2/`, no munging; specs#715 still open) — listener currently drops v2 first contacts with a debug log.
- go/js interop vectors.
- Known ceilings, commented in code: listener is a full aioice agent acting *controlled* (aioice has no ICE-Lite mode; interoperates because the dialer nominates); rate limiting is the in-flight cap only.

## Tests
- Loopback echo parametrized over STUN (default) and harness; **two concurrent dials on one port** (asserts two ufrags in the mux); unknown-prefix rejection (no PC created); in-flight cap; sdp helper table tests.
- `tests/core/transport/webrtc`: 204 passed; loopback file 6× no flake; mypy/pyrefly clean.

Stacked on #1448 → #1447 → #1446.

Refs #1437


yashksaini-coder:
@seetadev @acul71 Rebased on latest main. This PR adds the spec-aligned WebRTC-Direct listener: one shared UDP port, STUN `USERNAME` dispatch with `libp2p+webrtc+v1/` validation, offer inferred from the first packet, and a v1 dialer so py↔py loopback exercises the real STUN path; the HTTP `/sdp` harness is now opt-in. Stacked on #1448; closes the listener scope of #1437 — v2 (libp2p/specs#715) and go/js interop are planned follow-ups.

acul71 (changes requested):
Thanks for pushing this through — this is the right shape for the v1 STUN listener slice of #1437: shared `UdpMux` port, `libp2p+webrtc+v1/` first-contact dispatch, inferred offer, Noise initiator path, and `/sdp` harness opt-in off by default. Tests covering STUN vs harness, concurrent dials, unknown-prefix rejection, and the in-flight cap are in good shape. CI looks green.

Please do **not** `Fixes`/`Closes` #1437 on merge — v2, go/js interop, and true ICE-Lite are still open on that tracking issue. `Refs #1437` is correct.

## Requested changes

### 1. Spec-path dial must not use default Google STUN (blocker)

`dial()` always passes `self._config.ice_servers` into `create_peer_connection`. The default is `stun:stun.l.google.com:19302`. That undoes the offline-dial fix from the inbound/Noise PR in this stack: gathering the public STUN server adds seconds on firewalled/offline networks, phones Google on every `/webrtc-direct` dial, and does not help the STUN listener path (the offer is inferred from the first BINDING REQUEST to the advertised multiaddr). Loopback tests hide this with `ice_servers=[]` in `_transport()`.

For `enable_sdp_http_harness=False`, create the PC with `ice_servers=[]` (same as the listener). Keep `config.ice_servers` for the harness / explicit NAT-TURN opt-in. Please add a regression test that the default spec dial does not configure `RTCIceServer`s.

### 2. Unknown-STUN rate limiting (ship here or park on #1437)

Spec step 6 / #1437 ask for rate-limiting first-contact STUN. Today only `max_in_flight_connections` bounds allocated PCs, and it runs after parse + `add_ice_connection`. Malformed / wrong-prefix floods still hit `parse_direct_username` on every datagram. Prefer a small per-source-IP token bucket before allocating ICE/PC state (go-libp2p also bounds the mux accept queue). If you want to defer, tick it explicitly on #1437 and leave the `ponytail` comment — but this socket is now a public STUN endpoint, so shipping the bucket here is better.

## Please also

- Infer offer ICE creds from the **client** USERNAME half (`_client_ufrag`), matching go-libp2p’s v1 recovery (`client_pwd == client_ufrag`). py↔py works today only because both halves are the same `make_v1_credential()` string.
- After detecting v1, also validate the credential as an ice-pwd (RFC 8839 min 22). `libp2p+webrtc+v1/` is 17 chars, so a 4-char suffix yields a 21-char string that passes ufrag checks but is not a valid pwd. Add a reject table case for that.
- Drop unused `WebRTCDirectTransport._sdp_builder` (`SDPBuilder` stays for private transport).

## Questions

- Was re-passing `config.ice_servers` into spec-path `dial()` intentional, or leftover from the harness era?
- Is inferred `a=setup:active` (vs spec/go `actpass`) the long-term aiortc constraint, or can we force the DTLS server role another way before go interop?
- Should `0.0.0.0` listen still advertise `127.0.0.1`, or should `get_addrs()` prefer a non-loopback iface the way the Windows tests do?

Happy to re-review once the STUN-server dial regression is fixed.

libp2p/transport/webrtc/transport.py
```diff
@@ -143,9 +145,12 @@ async def dial(self, maddr: Multiaddr) -> WebRTCConnection:
                 "Ensure aiortc is installed and config uses from_aiortc()."
             )
 
+        pc: Any = None
         try:
             # 1. Create RTCPeerConnection + Noise channel
-            pc = await bridge.run_coro(create_peer_connection(rtc_cert))
+            pc = await bridge.run_coro(
+                create_peer_connection(rtc_cert, ice_servers=self._config.ice_servers)
```
acul71: Blocker: on the default (spec) path this reintroduces the Google STUN gather that the inbound/Noise work removed for offline dials. Srflx candidates never reach the STUN listener anyway — offer is inferred from the first BINDING REQUEST. Please use `ice_servers=[]` when `enable_sdp_http_harness` is false (same as the listener), and keep `config.ice_servers` for the harness / explicit TURN. Add a regression test so loopback helpers with `ice_servers=[]` cannot hide this again.

libp2p/transport/webrtc/listener.py
```diff
@@ -122,6 +169,135 @@ async def listen(self, maddr: Multiaddr) -> None:
...
+        back here), replays the packet into it, and schedules the async
+        acceptance. Anything malformed is dropped silently — this is an
+        unauthenticated network input.
+        """
+        mux, bridge = self._mux, self._bridge
+        if self._closed or mux is None or bridge is None:
+            return
+        # ponytail: in-flight cap only; add a per-source-IP token bucket if
+        # STUN floods from one address become a problem (spec: SHOULD rate-limit).
+        if self._in_flight >= self._config.max_in_flight_connections:
```
acul71, line 278: In-flight cap alone is not the rate-limit the spec / #1437 ask for — it only bounds allocated PCs after a successful parse + `add_ice_connection`. Malformed / wrong-prefix floods still pay `parse_direct_username` on every datagram. Prefer a per-source-IP token bucket before allocating ICE state (or explicitly park on #1437).

libp2p/transport/webrtc/listener.py
```diff
@@ -122,6 +169,135 @@ async def listen(self, maddr: Multiaddr) -> None:
...
+            attach_muxed_connection(pc, mux, conn)
+            # Spec step 6.2/7: B cannot know A's DTLS fingerprint, so it must
+            # not verify it during DTLS; the Noise handshake authenticates.
+            set_private_attr(
+                pc.sctp.transport, "_validate_peer_identity", lambda _params: None
+            )
+
+            offer_sdp = build_inferred_offer(
+                client_ufrag=server_ufrag,
+                client_pwd=server_ufrag,
```
acul71: Please use the client half of the USERNAME for the inferred offer (`client_ufrag=_client_ufrag`, `client_pwd=_client_ufrag` for v1), matching go-libp2p. Today py↔py works because both halves are the same credential; unequal halves (or a crafted `v1_cred:other`) would mismatch.

libp2p/transport/webrtc/sdp.py
```diff
@@ -38,8 +47,139 @@
...
+    valid ufrags. An unknown or missing prefix is rejected — the spec says
+    never to guess a version.
+
+    :returns: ``(version, server_ufrag, client_ufrag)``.
+    :raises WebRTCConnectionError: on any malformed / unsupported input.
+    """
+    server_ufrag, sep, client_ufrag = username.partition(":")
+    if not sep:
+        raise WebRTCConnectionError("STUN USERNAME is not 'server:client'")
+    if not (is_ice_ufrag(server_ufrag) and is_ice_ufrag(client_ufrag)):
```
acul71, line 103: For v1 the same string is also ice-pwd (RFC 8839 min 22). `libp2p+webrtc+v1/` is 17 chars, so a 4-char suffix (21 total) passes ufrag checks but is not a valid pwd. After detecting v1, also require `is_ice_pwd(...)` and add a reject table case.

acul71 (commented):
Inline notes for the changes requested above.

yashksaini-coder:
@acul71 All requested changes are in (rebased on main):

- **STUN-server dial (blocker):** spec path now creates the PC with `ice_servers=[]`; `config.ice_servers` only applies to the HTTP harness. Regression test `test_spec_dial_configures_no_ice_servers` uses *default* configs and asserts every `create_peer_connection` call saw `[]`. Answer to your question: it was a leftover from the harness era — my earlier "revert" edit hadn't actually applied, and CI couldn't tell because the helper passed `ice_servers=[]`.
- **Rate limiting:** per-source-IP token bucket (`_FirstContactLimiter`, burst 8, 4/s, bounded to 4096 sources) runs in `_on_unknown_stun` *before* any parse or ICE/PC allocation; the in-flight cap stays as the second bound. Test floods 32 first contacts from one IP and asserts ≤ burst get through.
- **Client half for the inferred offer:** `client_ufrag = client_pwd = <client half of USERNAME>` (go-libp2p v1 recovery).
- **v1 pwd length:** `parse_direct_username` now requires both halves to be valid ice-pwds (≥ 22) for v1; reject cases added for the 21-char credential and a short client half.
- Dropped `_sdp_builder`; `Refs #1437` only (no Fixes/Closes).

On the other questions: `a=setup:active` is only in the *inferred* offer, which never goes on the wire — it exists to steer aiortc, which answers `actpass` as DTLS client. Equivalent alternative is `actpass` + `_set_role("server")` on the DTLS transport; happy to switch if you prefer the spec-literal text. For `0.0.0.0`, advertising `127.0.0.1` predates this stack; I'd follow the TCP listener (advertise the bound wildcard / let the host expand interfaces) in a small follow-up rather than here.

acul71 (approved):
Round 3 addresses the deferred follow-ups:

- **DTLS setup:** `build_inferred_offer` now uses `a=setup:actpass`; `force_listener_dtls_server_role()` sets the listener as DTLS server before `createAnswer`.
- **0.0.0.0 advertising:** wildcard listens publish concrete interface IPs via `get_available_interfaces`; `_resolve_candidate_host` picks a non-loopback ICE host when available.
- **Docs:** README + module docstring note the default STUN path vs opt-in `enable_sdp_http_harness`.

All prior review blockers were already fixed in round 2. LGTM to merge once CI is green.

acul71:
Pushed **round 3** (`b1c548bb`) with the three deferred follow-ups:

1. **`actpass` + DTLS server role** — `libp2p/transport/webrtc/sdp.py`, `_aiortc_helpers.force_listener_dtls_server_role`, `listener._accept_v1`
2. **`0.0.0.0` → concrete addrs** — `listener._advertised_hosts_for_listen` / `_resolve_candidate_host` + `test_listener_on_wildcard_advertises_concrete_interfaces`
3. **README / module note** — STUN path vs `WebRTCTransportConfig(enable_sdp_http_harness=True)`

Local: lint, typecheck, docs, 219/219 webrtc tests, full suite green (Kad DHT flakes passed on retry).

Issue #1470, opened 2026-09-03
WebRTC-Direct: go-libp2p dialer can't connect to the py listener (ICE succeeds, then the connection is torn down)


yashksaini-coder:
While writing go-libp2p interop tests for #1437 I found that go → py doesn't complete, on both v1 and v2. py → go works fine (both versions dial and finish the Noise handshake), so this looks specific to the listener accept path, not the dialer creds/framing.

Setup: go-libp2p v0.49 dialing a py-libp2p WebRTC-Direct listener on the same host.

On the py side the STUN reaches the mux and ICE gets going, one candidate pair reaches SUCCEEDED, and then the peer connection is immediately closed before DTLS runs:

```
aioice Check CandidatePair(... -> 127.0.0.1:47884) IN_PROGRESS -> SUCCEEDED
RTCDataChannel(0) connecting -> closed
RTCIceTransport(controlled) checking -> closed -> failed
RTCDtlsTransport(server) NEW -> CONNECTING -> FAILED (DTLS handshake failed)
```

go just reports `peerconnection opening timed out`. The listener handler never fires.

I'm opening the interop test PR covering py → go now and marking go → py xfail pointing here. Happy to dig into the fix — @acul71 any idea whether `b1c548bb` (actpass + force DTLS server role) or the 0.0.0.0 concrete-interface advertising could be closing the PC right after ICE completes? A go harness to reproduce is in the PR.

Refs #1437

PR #1471, opened 2026-09-03
test(interop): go-libp2p WebRTC-Direct interop tests

yashksaini-coder:
The formal go-libp2p interop tests for #1437.

A small pinned go-libp2p v0.49 harness (`tests/interop/go_libp2p/webrtc_direct/`, `listen` / `dial -version N`) is built on demand under a file lock and skipped when the Go toolchain is missing, then driven against our WebRTC-Direct transport in both directions and both versions.

- **py → go**: passes live for v1 and v2 (dials a go listener, completes Noise, checks the peer ID).
- **go → py**: `xfail` (strict) — the inbound path tears the peer connection down right after ICE completes, tracked in #1470. The strict xfail means the suite will flag it the moment that's fixed.

Best on top of #1459 (needs the v2 dialer + framing fixes), but the harness/tests stand alone. go → py is a real listener-side bug, not something this PR introduces.

Refs #1437

🤖 Generated with [Claude Code](https://claude.com/claude-code)


acul71:
> 🤖 Generated with [Claude Code](https://claude.com/claude-code)

Hi @yashksaini-coder , you used claude code with which LLM ?

acul71 (changes requested):
## Review readiness

Thanks for the clean interop harness — scope matches the description (pinned go-libp2p v0.49 listen/dial, py→go live, go→py strict `xfail` for #1470), and the #1459 dependency is already on `main`.

Not ready to dig into a full review pass yet; a couple of blockers first:

### Blockers

1. **Lint is red.** Recent CI fails typecheck on `test_webrtc_direct_interop.py`:
   ```
   Object of class `NoneType` has no attribute `receive_some`
   ```
   (`proc.stdout` is optional after `open_process`). Please assert/narrow `stdout` (or otherwise satisfy the checker) and get lint green.

2. **Maintainer question still open** — which Claude model did you use with Claude Code? (comment above)

### Follow-ups (can land with or right after the lint fix)

- **Go is not installed in CI.** `.github/workflows/tox.yml` sets up Nim/Node for interop, but not Go. Without `actions/setup-go` (matching `go.mod` / go-libp2p v0.49), these tests will skip on Actions and CI won’t actually exercise the harness.
- Newsfragment is `1470.internal.rst` (issue id); consider `1471` if that’s the project convention for the PR that adds the tests.
- Commit history is noisy (already-merged framing/v2 work + merges); net diff vs `main` is fine — squash-on-merge is enough.

Happy to do a deeper review once lint is green (and ideally once Go is wired into the interop job so CI runs the new tests for real).

yashksaini-coder:
Thanks! Addressed the blockers and follow-ups:

- **Lint green** — narrowed `proc.stdout` (it's `Optional` after `open_process`); pyrefly is now clean on the interop module.
- **Go in CI** — added an `actions/setup-go@v5` step (go 1.25, matching `go.mod`) to the interop tox job, right after the Node.js setup, so the harness actually builds and the tests run on Actions instead of skipping.
- **Newsfragment** renamed `1470.internal.rst` → `1471.internal.rst` (PR id).
- **History** — rebased onto current main (v2/framing already landed), so the branch is now a single clean interop commit; net diff is just `tests/interop/go_libp2p/` + the CI/newsfragment bits.

Local: 2 passed (py→go v1+v2) / 2 xfailed (go→py v1+v2 → #1470). Ready for a deeper pass whenever you are.

yashksaini-coder:
> > 🤖 Generated with [Claude Code](https://claude.com/claude-code)
> 
> Hi @yashksaini-coder , you used claude code with which LLM ?

I recently got hands on claude code, mainly for implementing and testing the main feature changes affect on other module features. It also helps me explain for other cases and scenraios, which helped me not to raise any more follow up PRs, after a big feature or PR lands, as I used to do before

acul71 (approved):
## Re-review

Previous blockers are addressed:

- Lint: `proc.stdout` narrowed with `assert stdout is not None` — lint green
- Go in CI: `actions/setup-go@v5` (`go-version: '1.25'`) on the interop tox job; harness builds and tests run (py→go PASS v1/v2, go→py strict XFAIL → #1470)
- Newsfragment renamed to `1471.internal.rst`
- CI currently all green; merges cleanly with current `main` (1 commit behind, no conflicts)

Harness design looks good: pinned go-libp2p v0.49, session fixture + file lock for xdist, skip when Go missing, both directions/versions covered. Strict `xfail` on go→py is the right call until #1470 lands.

### Non-blocking follow-ups

1. **`_read_lines` drops leftover LISTEN lines.** In `test_py_dials_go`, the first `_read_lines(..., count=1)` can consume a multi-line burst and discard extra `LISTEN` lines still in the local buffer; the second call (`count=3`) then sits on the 45s `move_on_after` timeout. That matches the ~45s py→go call times in CI and can defeat `_lan_addr`’s non-loopback preference. Prefer one read that collects all `LISTEN` lines (short settle after the first), or preserve leftover buffer across calls.
2. Branch is 1 commit behind `main` — optional merge/rebase before squash; not required.
3. Claude model question still unanswered by name — optional.

**LGTM** — OK to squash-merge.

acul71:
Follow-up to the review note on `_read_lines` / LISTEN collection:

Pushed a fix that waits for the first `LISTEN` line, then briefly settles (~250ms) to drain sibling addresses from the same burst in one buffer — instead of a second `count=3` wait that discarded already-read lines and burned ~45s per py→go case.

Locally: `test_py_dials_go` v1/v2 both PASS in ~0.3–0.4s each (was ~45s call time). Also merged latest `main`.

PR #1486, opened 2026-09-04
fix(swarm): close connections on shutdown so their sockets are released

yashksaini-coder:
Refs #1485 (the trio-era form of #92). This PR does the first two items of #1485; the `filterwarnings` guard that fully closes it is deferred to #1487, once the residual leaks are at zero.

### The bug
`Swarm.run()`'s service-manager shutdown closed the listeners but never the live connections. So whenever a swarm was stopped through the trio service manager (`background_trio_service` / `manager.stop()`) — which is how the whole test suite and any service-managed process shuts down — every dialed or accepted socket leaked. Only an explicit `swarm.close()` released them, and that path is rarely taken.

### The fix
- Close each tracked connection in `Swarm.run()`'s shutdown (aligned with `Swarm.close()`), then clear `self.connections`. Close failures log at `warning`, matching `Swarm.close()`.
- Tear swarms down deterministically in the test factory (`SwarmFactory.create_and_listen` → `swarm.close()`), which the manager-cancel path can't fully do once its tasks are already cancelled.
- `raw_conn_factory` now closes both ends of the pair (the accepted server side wasn't being released).

### Result
`ResourceWarning: unclosed socket` counts, before → after (`pytest <dir> -W all::ResourceWarning`):

| dir | before | after |
|---|---|---|
| tests/core/host | 16 | **0** |
| tests/core/network | 25 | **1** |
| tests/core/identity/identify | 9 | **1** |

All affected suites pass (network + host: 364; security: 164; identity: 11). ruff + mypy clean.

### Not in this PR
The muxer-pair fixtures (`stream_muxer`) and `pubsub` still show residual leaks from a different teardown path (directly-yielded muxer conn pairs) — deferred to #1487, along with the `filterwarnings` guard once they're at zero.

PR #1497, opened 2026-09-07
fix(mplex): close secured connection on shutdown so its socket is released

### LINKED ISSUE #1487: Residual unclosed-socket leaks in the muxer conn/stream-pair test fixtures (follow-up to #1485)
Follow-up to #1485 / #1486. #1486 fixed the swarm-shutdown socket leak — `host`, `network`, `identity`, and `pubsub` dropped to ~0 unclosed sockets. A residual remains, concentrated in the stream-muxer conn/stream-pair tests.

### Observed (post-#1486, `pytest <target> -W all::ResourceWarning -p no:cacheprovider -o addopts="" -q`)
- `tests/core/stream_muxer/test_mplex_stream.py` → **33** `ResourceWarning: unclosed <socket.socket>` (accounts for essentially the whole stream_muxer total); the yamux stream-pair equivalent is expected to be similar.
- `tests/core/pubsub/test_subscription.py`, `host`, `network`, `identity` → ~0 (fixed by #1486).

### Why this is a separate follow-up, not part of #1486
- **Different teardown path.** #1486 fixed the swarm *shutdown* path (`Swarm.run()` closing `self.connections` + `SwarmFactory.create_and_listen` → `swarm.close()`). These tests use `mplex_conn_pair_factory` / `mplex_stream_pair_factory` → `swarm_conn_pair_factory` → `swarm_pair_factory`, which yield the **muxed connection / streams extracted directly** from `swarm.connections[...]`. The swarm *is* closed on teardown, yet these sockets persist — so the leak is in how the muxer-pair connections are released, not the swarm-shutdown bug #1486 fixed.
- **Likely cause to confirm:** `SwarmConn.close()` skips `await self.muxed_conn.close()` when a connection is flagged `_shared_muxed_conn` (the Bug-3 dedup guard). If a pair-factory connection ends up flagged shared, its muxed connection — and the underlying socket — is never closed. Needs verification.
- **Scope hygiene:** keeping it out of #1486 keeps that PR one small, clearly-rooted fix with measured results, instead of mixing in a second fix with a different root cause.

### Approach
1. Confirm the source — instrument which `close()` path (or absence of one) leaves the muxer-pair sockets open; check the `_shared_muxed_conn` guard in `SwarmConn.close()` / `_cleanup()`.
2. Fix at the fixture and/or library — give `swarm_conn_pair_factory` (and the mplex/yamux conn + stream pair factories) explicit teardown that closes the yielded muxed connections; if the `_shared_muxed_conn` guard is wrongly suppressing socket close for non-shared connections, fix the guard.
3. Re-measure to ~0 across `stream_muxer`.
4. Once the whole suite is at zero, add `filterwarnings = ["error::ResourceWarning"]` to `pyproject.toml` so socket leaks fail CI and can't silently regress — deferred from #1486 for exactly this reason.

Refs #92, #1485, #1486

yashksaini-coder:
Refs #1487, #92. Clears the residual stream-muxer socket leaks left after #1486.

## The bug
`Mplex.close()` returned early once `event_shutting_down` was set:

```python
    if self.event_shutting_down.is_set():
        return          # <- skips secured_conn.close()
    self.event_shutting_down.set()
    await self.secured_conn.close()
    await self.event_closed.wait()
```

The read loop's `_cleanup()` sets `event_shutting_down` on a peer-initiated EOF **without** closing the underlying secured connection. So whenever the remote closed first, the local `close()` hit the early return and the socket was never released — a `ResourceWarning: unclosed <socket.socket>` per connection.

Confirmed by instrumenting the close chain on a muxer pair: both `Mplex.close()` calls ran, but only one reached `secured_conn.close()` / `RawConnection.close()`; the initiator side (whose read loop had already flagged shutdown) leaked its socket.

## The fix
Close the secured connection unconditionally (it is idempotent), mirroring `Yamux.close()`, which already does this.

## Result
`ResourceWarning: unclosed socket`, before -> after (`pytest <dir> -W all::ResourceWarning -p no:cacheprovider -o addopts="" -q`):

| dir | before | after |
|---|---|---|
| tests/core/stream_muxer | 33 | **0** |

Correctness: `stream_muxer` 128 passed; `network` + `relay` 395 passed (no regression). ruff + mypy clean.

## Not in this PR
Residual `ResourceWarning` leaks remain in other subsystems (transport, bitswap, identity, io, pubsub, discovery, network) from different teardown paths. The `filterwarnings = ["error::ResourceWarning"]` guard from #1487 is deferred until those reach zero, so this PR does not close #1487.


acul71 (approved):
LGTM — clear root cause and a minimal correct fix.

`Mplex.close()` no longer early-returns past `secured_conn.close()` when `_cleanup()` already set `event_shutting_down` on peer EOF, matching `Yamux.close()`. Stream-muxer ResourceWarnings cleared (local: 128 passed, 0 warnings); CI green. Optional follow-ups only: close in `_cleanup()` for defense-in-depth, and prefer closing #1487 with residuals on #1498.

Approved.

PR #1536, opened 2026-09-12
fix(rendezvous): strip /p2p suffix from REGISTER addrs

### LINKED ISSUE #1535: Rendezvous REGISTER advertises /p2p-suffixed addrs; strict servers reject registration
## Description

`RendezvousClient.register()` (and the auto-refresh loop) pass `host.get_addrs()` verbatim into `create_register_message()`. Those addresses carry a `/p2p/<peer-id>` suffix, e.g. `//ip4/10.0.0.5/tcp/34567/p2p/QmFoo..`.

Strict rendezvous servers dial back advertised addresses to verify reachability before storing the record. The suffixed addresses fail this dial-back ("no good addresses"), so registration is rejected.

## Reproduction

Register a py-libp2p client against a strict server (e.g. go-libp2p rendezvous):

1. Start a go-libp2p rendezvous server.
2. `client = RendezvousClient(host, server_peer_id); await client.register(ns, 300)`
3. Server rejects the REGISTER; wire capture shows advertised addrs with `/p2p` suffix.

Observed while building the libp2p unified-testing rendezvous interop suite (go server x py registrant); a client-side address-stripping workaround was needed in the test harness to get 8/8 green.

## Expected behavior

REGISTER should advertise bare transport addresses. The peer ID is already conveyed in the message's peer record, so the `/p2p` suffix is redundant and actively harmful.

## Proposed fix

Strip the `/p2p` suffix in `create_register_message()` (single choke point covering `register()` and refresh). Narrowly scoped: `host.get_addrs()` itself is left untouched.

## Environment

- py-libp2p: main (0.7.0 affected)
- Counterpart: go-libp2p rendezvous server (berty fork)

sumanjeet0012:
Fixes #1535.

## Problem

`RendezvousClient.register()` (and the auto-refresh loop) pass `host.get_addrs()` verbatim into the REGISTER message. Those addresses carry a `/p2p/<peer-id>` suffix. Strict servers dial back advertised addresses before storing the record, and the suffixed addresses fail dial-back ("no good addresses"), so registration against non-Python servers is rejected.

Found while building the unified-testing rendezvous interop suite (go server x py registrant) — a harness-side workaround was needed for 8/8 green.

## Fix

Strip the `/p2p` suffix in `create_register_message()` — the single choke point covering `register()` and refresh. `host.get_addrs()` is untouched, so no blast radius on identify/DHT/autonat. The peer ID is already conveyed in the message's peer record.

## Verification

- New unit test `test_create_register_message_strips_p2p_suffix`; full rendezvous suite 119/119 pass.
- Interop proof: with this fix, the unified-testing rendezvous suite (go x py, 8 combos) passes with the harness workaround removed.
- `ruff check` / `ruff format` clean; `mypy-local` and `pyrefly-local` report no errors in the touched file; newsfragment validated with `validate_files.py`.

## Newsfragment

`newsfragments/1535.bugfix.rst`

### DIFF
--- a/libp2p/discovery/rendezvous/messages.py
+++ b/libp2p/discovery/rendezvous/messages.py
@@ -3,6 +3,7 @@ Message construction helpers for rendezvous protocol.
 """
 
 from multiaddr import Multiaddr
+from multiaddr.exceptions import ProtocolLookupError
 
 from libp2p.peer.id import ID as PeerID
 
@@ -20,6 +21,15 @@ def create_register_message(
     peer_info = msg.register.peer
     peer_info.id = peer_id.to_bytes()
     for addr in addrs:
+        # Advertise bare transport addresses: strict servers reject (or fail
+        # to dial back) addresses carrying a `/p2p/<peer-id>` suffix, and the
+        # peer ID is already conveyed in `peer_info.id`.
+        try:
+            p2p_val = addr.value_for_protocol("p2p")
+        except ProtocolLookupError:
+            p2p_val = None
+        if p2p_val:
+            addr = addr.decapsulate(Multiaddr(f"/p2p/{p2p_val}"))
         peer_info.addrs.append(addr.to_bytes())
 
     msg.register.ns = namespace
--- a/newsfragments/1535.bugfix.rst
+++ b/newsfragments/1535.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the Rendezvous client advertising ``/p2p``-suffixed addresses in REGISTER messages. Strict servers dial back advertised addresses to verify reachability and rejected these registrations, so registering against non-Python servers failed. The client now advertises bare transport addresses.
--- a/tests/core/discovery/rendezvous/test_messages.py
+++ b/tests/core/discovery/rendezvous/test_messages.py
@@ -57,6 +57,19 @@ class TestMessageCreation:
         assert list(message.register.peer.addrs) == expected_addrs
         assert message.register.ttl == ttl
 
+    def test_create_register_message_strips_p2p_suffix(
+        self, sample_peer_id, sample_addrs
+    ):
+        """REGISTER must advertise bare transport addrs (strict servers dial back)."""
+        suffixed = [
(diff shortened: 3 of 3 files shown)


sumanjeet0012:
@seetadev This PR introduces an improvement to the Rendezvous implementation to enhance interoperability with other libp2p implementations.

The changes have been thoroughly tested, and Rendezvous interoperability tests are now passing successfully without any workarounds or patches.

The PR is ready for review and merge.

sumanjeet0012:
@acul71 Ready for review.

seetadev:
@sumanjeet0012, this is a great improvement to the Rendezvous implementation, and thank you for identifying and addressing this interoperability issue through the unified-testing work. The root cause is clearly documented, and the decision to strip the `/p2p` suffix at the `create_register_message()` choke point is a focused solution that preserves the behavior of `host.get_addrs()` for other protocols such as Identify, DHT, and AutoNAT.

I reviewed the approach and the verification details. The fact that the fix covers both the initial `register()` call and the auto-refresh loop is particularly important, as it ensures consistent address handling throughout the registration lifecycle. Keeping the peer ID in the peer record while advertising bare transport addresses also aligns with the requirements of strict Rendezvous servers that perform dial-back validation.

The interoperability validation is a strong addition here. Having the unified-testing Rendezvous suite pass all 8 Go × Python combinations without the previous harness workaround demonstrates that this fix addresses a practical cross-implementation compatibility issue. The dedicated unit test, full Rendezvous suite passing at 119/119, and successful formatting, linting, and type checks provide good coverage for the change.

I also appreciate the follow-up refactoring to use `decapsulate_code(P_P2P)`. This makes the implementation more concise and protocol-aware, while supporting a no-op when the `/p2p` component is absent. The approach avoids unnecessary assumptions about multiaddress ordering and keeps the change localized.

This looks good to me. Approving the PR. Great work connecting the issue found in the unified-testing interop suite back to a clean upstream fix in py-libp2p. This is exactly the kind of focused interoperability improvement that helps strengthen consistency across libp2p implementations. Thank you for the thorough testing and for removing the workaround from the interop harness. LGTM!

Ccing @acul71

Issue #1223, opened 2026-02-15
Improve Multiaddr Protocol Stack Handling using decapsulate_code()


anuragShingare30:
## Problem

Currently the code manually removes protocols using string-constructed Multiaddr objects
**Location**: `libp2p/transport/websocket/multiaddr_utils.py`
```py
rest_multiaddr = rest_multiaddr.decapsulate(Multiaddr("/ws"))
rest_multiaddr = rest_multiaddr.decapsulate(Multiaddr(f"/sni/{sni}"))
rest_multiaddr = rest_multiaddr.decapsulate(Multiaddr("/tls"))
```

And circuit relay parsing performs complex protocol extraction logic instead of using protocol stack operations.
```py
# Lines 480-521: Complex manual parsing of circuit addresses
def _parse_circuit_multiaddr(self, ma: multiaddr.Multiaddr) -> tuple[multiaddr.Multiaddr, ID]:
    # Manual protocol extraction and parsing
```

**Drawbacks:**
- String parsing overhead
- Error-prone when protocol ordering varies
- Harder to read intent
- Duplicates functionality already provided by py-multiaddr
- Fragile for edge cases (optional protocols, reordered stacks)

## Proposed Solution or improvement

Use `decapsulate_code(protocol_code)` provided by `py-multiaddr` and replace some manual instance from codebase

**Location:** `libp2p/transport/websocket/multiaddr_utils.py`
```py
# current code
rest_multiaddr = rest_multiaddr.decapsulate(Multiaddr("/ws"))
rest_multiaddr = rest_multiaddr.decapsulate(Multiaddr(f"/sni/{sni}"))
rest_multiaddr = rest_multiaddr.decapsulate(Multiaddr("/tls"))

# Replace
from multiaddr.protocols import P_WS, P_SNI, P_TLS
rest_multiaddr = maddr.decapsulate_code(P_WS)
if sni:
    rest_multiaddr = rest_multiaddr.decapsulate_code(P_SNI)
rest_multiaddr = rest_multiaddr.decapsulate_code(P_TLS)
```

**Location:** `libp2p/relay/circuit_v2/transport.py`

Simplify manual protocol extraction logic in `_parse_circuit_multiaddr` by using protocol codes
```py
# Current code
# Lines 480-521: Complex manual parsing of circuit addresses
def _parse_circuit_multiaddr(self, ma: multiaddr.Multiaddr) -> tuple[multiaddr.Multiaddr, ID]:
[...]

PR #1539, opened 2026-09-13
refactor(identify): strip /p2p via decapsulate_code(P_P2P) (refs #1223)

yashksaini-coder:
Refs #1223 (Improve Multiaddr Protocol Stack Handling using `decapsulate_code()`).

## Context
#1223 asks to replace manual string-constructed `decapsulate(Multiaddr("..."))` calls with py-multiaddr's `decapsulate_code(protocol_code)` (value- and order-independent, no CID/string reconstruction, clearer intent).

The two sites the issue cited are **already migrated** on `main`: `transport/websocket/multiaddr_utils.py` uses `decapsulate_code(P_WS/P_SNI/P_TLS/P_WSS)`, and circuit-relay parsing (`parse_circuit_ma`) uses `decapsulate_code`. A repo-wide grep for `\.decapsulate\(\s*Multiaddr\(` found **exactly one** remaining library instance:

`libp2p/identity/identify/identify.py` — `_strip_p2p_suffix()`:
```python
p2p_value = maddr.value_for_protocol("p2p")
return maddr.decapsulate(Multiaddr(f"/p2p/{p2p_value}"))
```

## Change
```python
return maddr.decapsulate_code(P_P2P)
```
`decapsulate_code(P_P2P)` removes the `/p2p` layer by protocol code and is a no-op when `/p2p` is absent, so the `try/except` is dropped too. This completes the migration #1223 tracked (11 `decapsulate_code` calls, now 0 string-built ones in the library).

## Behavior (verified empirically with valid peer IDs)
`_strip_p2p_suffix` feeds Identify's `listen_addrs` (the spec requires listenAddrs without a `/p2p` peer-id suffix).

| input | old | new |
|---|---|---|
| `/ip4/…/tcp/…` (no /p2p) | unchanged | unchanged |
| `/ip4/…/tcp/…/p2p/<peer>` | stripped | stripped (same) |
| `/ip6/::1/tcp/…/p2p/<peer>` | stripped | stripped (same) |
| `/ip4/…/p2p/<relay>/p2p-circuit/p2p/<peer>` | truncates at **first** /p2p → `/ip4/…/tcp/…` (drops relay path) | strips **trailing** /p2p → `/ip4/…/p2p/<relay>/p2p-circuit` |

The common single-/p2p and no-/p2p cases are identical; the relay/circuit case is a refinement — the new result is the spec-correct listenAddr (only the trailing peer id removed), where the old code truncated the whole relay path.

## Tests
Added `test_strip_p2p_suffix` (all four cases above). `tests/core/identity` — 41 passed; ruff/mypy clean.


acul71 (approved):
## Review

Nice, focused completion of #1223 — last library `decapsulate(Multiaddr(...))` site moved to `decapsulate_code(P_P2P)`, with a clear unit test and a sensible circuit+trailing-peer refinement.

**Approved.**

One docstring nit (optional, but appreciated): `_strip_p2p_suffix` says relay/circuit paths always keep `/p2p/<relay>/p2p-circuit`, but that only holds when a trailing dest `/p2p` is present. For `/…/p2p/<relay>/p2p-circuit` alone, `decapsulate_code(P_P2P)` still truncates (same as the old code — not a regression). Tightening the docstring would help; a follow-up for true suffix-only stripping is fine separately.

Holding the squash-merge until CI/CD is green.

